PR #27, opened 2015-03-02
Use virtual-dom for rendering the interface

mtyaka:
This patch rewrites the JS code to use [virtual-dom](https://github.com/Matt-Esch/virtual-dom) rather than ad-hoc jQuery element manipulation for rendering the interface.

It also includes a change where the items no longer occupy space in the list on the left after they are dropped to the correct zone.


e-kolpakov:
@mtyaka :+1: conditional review notes are addressed.

.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: python
+python:
+    - "2.7"
+before_install:
+    - "export DISPLAY=:99"
+    - "sh -e /etc/init.d/xvfb start"
+install:
+    - "sh install_test_deps.sh"
+    - "python setup.py develop"
```
antoviaque: This would not catch things like assets that aren't included in the package, right? To catch this type of issues with Travis, can you install the package instead of linking directly to the repository, like in https://github.com/open-craft/xblock-mentoring/blob/master/.travis.yml#L12 ?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -204,7 +183,7 @@ def do_attempt(self, attempt, suffix=''):
                 # If this item has no input value set, we are done with it.
                 is_correct = True
                 feedback = item['feedback']['correct']
-            state = {'top': attempt['top'], 'left': attempt['left']}
+            state = {'top': attempt['top'], 'left': attempt['left'], 'position': 'absolute'}
```
antoviaque: Would it be possible to avoid adding more CSS in the JS code in this case? I understand the need to specify the coordinates in the the code, but can `position: absolute` be set in the CSS files? Or via a class if this needs to change dynamically?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1,388 +1,325 @@
...
-    publish_event({event_type:"xblock.drag-and-drop-v2.loaded"});
+                var is_input_visible = item_state && !item_state.submitting_location;
+                input = (
+                    h('div.numerical-input', {className: className,
+                                              style: {display: is_input_visible ? 'block' : 'none'}},
+                      [
+                          h('input.input', {type: 'text', disabled: has_input,
+                                            focusHook: has_input ? undefined : FocusHook(),
+                                            value: has_input ? item_state.input : ''}),
+                          h('button.submit-input', {disabled: has_input}, 'ok')
```
antoviaque: I like the virtual-dom refactor, but one drawback is that you are now including HTML in the JS code. I imagine it isn't really possible to use templating with this approach? If not, can we at least keep remain as close to the way we would approach this via templating - ie extracting the definition to separate functions outside of the current scope, in a different JS file, and passing variables explicitly via a context object like you did below with `data`? To try to minimize the temptation to intertwine code and presentation, and keep these separate as much as possible.

install_test_deps.sh
```diff
@@ -0,0 +1,6 @@
+# Installs xblock-sdk and dependencies needed to run the tests suite.
+# Run this script inside a fresh virtual environment.
+pip install -e git://github.com/edx/xblock-sdk.git#egg=xblock-sdk
+pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements.txt
+pip install -r $VIRTUAL_ENV/src/xblock-sdk/test-requirements.txt
+python setup.py develop
```
antoviaque: Why not using a requirements.txt for declaring the xblock-sdk requirement? That take care of the first two lines, no?

tests/integration/test_base.py
```diff
@@ -48,8 +48,8 @@ def _add_scenario(self, identifier, title, xml):
         self.addCleanup(scenarios.remove_scenario, identifier)
 
     def _get_items(self):
-        items_container = self._page.find_element_by_css_selector('ul.items')
-        return items_container.find_elements_by_css_selector('li.option')
+        items_container = self._page.find_element_by_css_selector('div.items')
+        return items_container.find_elements_by_css_selector('div.option')
```
antoviaque: Maybe a good occasion to not select based on the DOM tag name, but simply on the class?

tests/integration/test_interaction.py
```diff
@@ -50,10 +50,13 @@ def setUp(self):
...
         self.browser.get(self.live_server_url)
         self._page = self.go_to_page(self.PAGE_TITLE)
+        # Resize window so that the entire drag container is visible.
+        # Selenium has issues when dragging to an area that is off screen.
+        self.browser.set_window_size(1024, 800)
 
     def _get_item_by_value(self, item_value):
-        items_container = self._page.find_element_by_css_selector('ul.items')
-        return items_container.find_elements_by_xpath("//li[@data-value='{item_id}']".format(item_id=item_value))[0]
+        items_container = self._page.find_element_by_css_selector('div.items')
```
antoviaque: Same remark about selecting by DOM tag name

antoviaque:
Also, when adding a drag and drop in Studio, the preview doesn't display it, you have to reload the Studio page to be able to see it:

screenshot-localhost 8001 2015-03-02 12-43-56

I haven't checked the tests, too - this should show up here via Travis on this once it is activated.

drag_and_drop_v2/drag_and_drop_v2.py
mtyaka: The `'position': 'absolute'` is just a flag for backwards compatibility, so that we know the stored `top` and `left` values are absolute. It might as well be called `'format': '2.0' or something like that.

Before this patch the values were relative. If we didn't have a flag and treated all stored values as absolute, the positions of values stored before this patch would be slightly off.

I hate the fact that we will have to keep a flag for backwards compatibility forever, but I can't think of a better way to do that.

install_test_deps.sh
mtyaka: It wouldn't - `pip install` does not recursively install `requrement.txt` files and while installing xblock-sdk with pip does install the dependencies specified in xblock-sdk's `setup.py` file, that file does not list the git dependencies from `requirements.txt`.

drag_and_drop_v2/drag_and_drop_v2.py
antoviaque: @mtyaka I see - it might be worth making it obvious that it's a flag, or at least using a different name/value than what is used in CSS, as it looks like it's going to be directly translated to a CSS property currently.

mtyaka: @antoviaque Makes sense, I"ll replace `'position': 'absolute'` with `'absolute': True` and add a comment explaining why the flag is needed.

.travis.yml
mtyaka: It took me a while to realize what you mean, but I think I get it now. This just ensures that the build fails if you forget to include some directories in the `package_data` in `setup.py`, but assuming your `package_data` is correct, there should be no difference, right?

antoviaque: @mtyaka Yup, exactly.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -204,7 +172,11 @@ def do_attempt(self, attempt, suffix=''):
                 # If this item has no input value set, we are done with it.
                 is_correct = True
                 feedback = item['feedback']['correct']
-            state = {'top': attempt['top'], 'left': attempt['left']}
+            state = {
+                'top': attempt['top'],
+                'left': attempt['left'],
+                'absolute': True  # flag for backwards compatibility (values used to be relative)
```
antoviaque: Thanks for adding a comment too, here

antoviaque:
@mtyaka Changes looks good! :+1: 

I still have the empty preview issue I described above though, upon adding a drag and drop the first time. It works well on reload and in apros, so I'm going to merge now to send to QA, but could you have a look?

mtyaka:
@antoviaque Thanks! The empty preview was supposed to be fixed with the last update. I just tried again and the preview loads fine after adding a new drag and drop block to the page for me, no need to reload.

antoviaque:
@mtyaka Might just have been me then - I might not have correctly reloaded the latest version - thanks for checking!

e-kolpakov:
@antoviaque @mtyaka adding drag-n-drop for the first time shows normal preview for me.

antoviaque:
@e-kolpakov Cool, thanks for the confirmation!

PR #28, opened 2015-03-20
Wrap bg image src in url(...).

mtyaka:
Background image style property needs to be given as
url(...src...) to be valid CSS.


mtyaka:
@antoviaque Please review.

antoviaque:
@mtyaka As discussed, merging to get this to QA & prod as a hotfix, but will still need tests in a follow-up PR. :+1:

PR #31, opened 2015-06-10
Use numbers instead of letters to identify sample zones

antoviaque:
Prior to this patch, the sample zones in the default dataset would use letters to identify zones ("Zone A"), but zones added to the default set would use numbers ("Zone 3"). This changes the sample zones to use numbers too.

For consistency and maintainability, the change was also applied to all tests, even when the test would still have worked with the letter format.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,5 @@ install:
 script: pep8 drag_and_drop_v2 --max-line-length=120 && python run_tests.py
 notifications:
   email: false
+addons:
+  firefox: "36.0"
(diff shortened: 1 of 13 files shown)


smarnach:
I tested this thoroughly on Apros, and it looks good to me.  The code changes also look good – the zone names had to be changed in surprisingly many places, but that's not your fault. :+1:

Kelketek:
@antoviaque Seems to work fine, and the changes LGTM. :+1:

PR #33, opened 2015-12-02
Allow authors to set background and text color of draggable items

itsjeyd:
This PR adds functionality for specifying background and text color of draggable items via the interactive editor.

If a course author does not specify custom colors for background and/or text, colors will be determined by the current theme, or default to values set via CSS.


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -50,6 +50,26 @@ class DragAndDropBlock(XBlock):
...
     )
 
+    item_background_color = String(
+        display_name="Item background color",
+        help=(
+            "Background color to use for draggable items. "
+            "Defaults to color specified by current theme, or blue if no theme is set."
+        ),
+        scope=Scope.settings,
+        default=""
```
bradenmacdonald: Nit: Always include a trailing comma in objects and lists that have one key-value per line, to make future diffs nicer.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -50,6 +50,26 @@ class DragAndDropBlock(XBlock):
...
+    )
+
+    item_text_color = String(
+        display_name="Item text color",
+        help=(
+            "Text color to use for draggable items. "
+            "Defaults to color specified by current theme, or blue if no theme is set."
```
bradenmacdonald: "or *white* if no theme is set"

tests/integration/test_render.py
```diff
@@ -8,22 +25,73 @@ class TestDragAndDropRender(BaseIntegrationTest):
...
-        style = item.get_attribute('style')
+    def _test_style(self, element, style_settings, element_type):
+        style = element.get_attribute('style')
         for style_prop, expected_value in style_settings.items():
+            if style_prop == 'color' or style_prop == 'background-color' and expected_value.startswith('#'):
+                expected_value = Colors.rgb(expected_value)
             expected = u"{0}: {1}".format(style_prop, expected_value)
             self.assertIn(expected, style)
+        if element_type == "item":
+            self._test_item_style(element, style_settings, style)
```
bradenmacdonald: I think you could do the test for CSS properties a lot more robustly with something like this:

```
bg_color =  self.browser.execute_script('getComputedStyle($("#tester").get(0)).backgroundColor'
```

The advantage of that is you don't have to worry about parsing the `style` string, and it checks the "computed" values, so you can check what color is actually used in the browser regardless of how complex the CSS styles/rules are.

tests/integration/test_render.py
```diff
@@ -8,22 +25,73 @@ class TestDragAndDropRender(BaseIntegrationTest):
...
+
+        self.assertEqual(items[1].get_attribute('data-value'), '1')
+        self.assertEqual(items[1].text, '2')
+        self.assertIn('ui-draggable', self.get_element_classes(items[1]))
+        self._test_style(items[1], {'width': '190px', 'height': 'auto'}, element_type='item')
+
+        self.assertEqual(items[2].get_attribute('data-value'), '2')
+        self.assertEqual(items[2].text, 'X')
+        self.assertIn('ui-draggable', self.get_element_classes(items[2]))
+        self._test_style(items[2], {'width': '100px', 'height': '100px'}, element_type='item')
```
bradenmacdonald: Can you shorten this method by rewriting it as a loop?

tests/integration/test_render.py
```diff
@@ -31,37 +99,53 @@ def test_items(self):
...
+            items[1], dict({'width': '190px', 'height': 'auto'}, **color_settings), element_type='item'
+        )
 
         self.assertEqual(items[2].get_attribute('data-value'), '2')
         self.assertEqual(items[2].text, 'X')
         self.assertIn('ui-draggable', self.get_element_classes(items[2]))
-        self._test_style(items[2], {'width': '100px', 'height': '100px'})
+        self._test_style(
+            items[2], dict({'width': '100px', 'height': '100px'}, **color_settings), element_type='item'
+        )
```
bradenmacdonald: Same here, can you replace these three chunks of code with a loop?

tests/integration/test_render.py
```diff
@@ -8,22 +25,73 @@ class TestDragAndDropRender(BaseIntegrationTest):
...
+        ('', Colors.GREY),
+    )
+    def test_items_custom_colors(self, item_background_color, item_text_color):
+        self.load_scenario(item_background_color, item_text_color)
+        color_settings = {}
+        if item_background_color:
+            color_settings['background-color'] = item_background_color
+        if item_text_color:
+            color_settings['color'] = item_text_color
+        print(color_settings)
```
bradenmacdonald: `print` left in

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -64,6 +64,18 @@
...
+                <section class="tab-content item-background-color-form">
+                    <label>
+                        Background color for items:
+                        <input class="item-background-color" type="text" value="{{ self.item_background_color}}">
+                    </label>
+                </section>
+                <section class="tab-content item-text-color-form">
+                    <label>
+                        Text color for items:
+                        <input class="item-text-color" type="text" value="{{ self.item_text_color}}">
```
bradenmacdonald: I don't think it looks good how these inputs are not aligned:
screen shot 2015-12-02 at 3 29 20 pm

Can you convert them to the same appearance/alignment as the fields on the first step?
screen shot 2015-12-02 at 3 29 14 pm

Also, it's not obvious to non-programmers what sort of values can be put into this field. Can you please add examples ("e.g. 'red' or '#ff0000'") and/or help text, either as a `placeholder` or as small text underneath the field?

bradenmacdonald: Also please put `{% trans "" %}` around any new strings that you add.

drag_and_drop_v2/drag_and_drop_v2.py
itsjeyd: @bradenmacdonald OK, I added trailing commas to the fields that this PR introduces, and also took the opportunity to add trailing commas to all of the remaining fields.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
itsjeyd: @bradenmacdonald OK, the input fields look like this now:

input-fields-updated

I removed the single-quotes around `red` and `#ff0000` to avoid potential confusion about whether or not color names should be quoted.

tests/integration/test_render.py
itsjeyd: @bradenmacdonald I added logic for checking the color that's actually used in the browser. Leaving the logic for checking the `style` string in place though, since I think it makes sense to check background and text colors from two angles (i.e., to check both what users see in the browser and whether the mechanism for injecting custom styles is working).

itsjeyd:
@bradenmacdonald Once you've re-reviewed, who should I ping for an upstream review for this?

bradenmacdonald:
As discussed, upstream review will come later.

PR #32, opened 2015-09-16
[MCKIN-3486] Enable course designer to hide header title

itsjeyd:
This PR adds an option to the interactive editor that allows course designers to specify whether or not the title of a drag-and-drop block should be shown:

interactive-editor-with-show-title-option

The default behavior is to show the title; if a course designer wants to change that they can simply uncheck the "Show title" check box. No `.problem-header` element will be added to the HTML for the block in this case.

#### Screenshots: LMS

"Show title" option checked:

drag-and-drop-block-with-title-shown

"Show title" option unchecked:

drag-and-drop-block-with-title-hidden

#### Screenshots: Apros

"Show title" option checked:

apros-drag-and-drop-block-with-title-shown

"Show title" option unchecked:

apros-drag-and-drop-block-with-title-hidden

PR #34, opened 2015-12-03
Drag and Drop v2 - Review for release on edx.org

bradenmacdonald:
This is the main PR for the [SOL-1403 DnDv2 on edx.org epic](https://openedx.atlassian.net/browse/SOL-1403).

Recent reviews have taken place on the following constituent PRs:
* #33 Allow authors to set background and text color of draggable items: [SOL-1406](https://openedx.atlassian.net/browse/SOL-1406), [SOL-1409](https://openedx.atlassian.net/browse/SOL-1409)
* #35 Add License (AGPLv3): [SOL-1420](https://openedx.atlassian.net/browse/SOL-1420)
* #38 Move draggable items above drop image area: [SOL-1404](https://openedx.atlassian.net/browse/SOL-1404)
* https://github.com/open-craft/xblock-drag-and-drop-v2/pull/4 Make the image resize to fit the container: [SOL-1450](https://openedx.atlassian.net/browse/SOL-1450), Allow problem reset after the first interaction: [SOL-1408](https://openedx.atlassian.net/browse/SOL-1408)
* #39 Support for multiple DnD blocks on the same page: [SOL-1448](https://openedx.atlassian.net/browse/SOL-1448)
* #40 Styling changes from UX review: [SOL-1410](https://openedx.atlassian.net/browse/SOL-1410)
* https://github.com/open-craft/xblock-drag-and-drop-v2/pull/5 Improve default settings and explain settings more clearly: [SOL-1451](https://openedx.atlassian.net/browse/SOL-1451)
* #41 Make DnDv2 accessible: [SOL-1412](https://openedx.atlassian.net/browse/SOL-1412)
* #42 DnDv2 - Show an indicator/spinner while AJAX requests are in progress: [SOL-1452](https://openedx.atlassian.net/browse/SOL-1452)
* #44 DnDv2 - Make the block keyboard navigable: [SOL-1462](https://openedx.atlassian.net/browse/SOL-1462)
* #45 DnDv2 - Allow authors to show the borders on drop zones: [SOL-1405](https://openedx.atlassian.net/browse/SOL-1405)
* #46 Fix flaky tests: [SOL-1449](https://openedx.atlassian.net/browse/SOL-1449)
* #47 Update and refine analytics events and their documentation: [SOL-1473](https://openedx.atlassian.net/browse/SOL-1473)
* #48 Add theming: [SOL-1411](https://openedx.atlassian.net/browse/SOL-1411)
* #52 Re-create all of the documentation screenshots: [SOL-1453](https://openedx.atlassian.net/browse/SOL-1453)
* #49 Improve usability on mobile: [SOL-1474](https://openedx.atlassian.net/browse/SOL-1474)

TODO (during review):
* a11y: Check if the grey color of the focus outline on the drop targets is OK, or if there's a better way to make it visible against varied backgrounds
* a11y: Check if the grey dotted borders around drop zones when authors turn on "show drop zones borders" are OK

**Testing**

We have a sandbox available for testing DnDv2: [Studio](http://studio.master.dndv2.sandbox.opencraft.com/), [LMS](http://master.dndv2.sandbox.opencraft.com/).

**Reviewers**

- [ ] a11y: @cptvitamin
- [x] UX: @talbs
- [x] Strings: @pdesjardins
- [x] TNL: @dianakhuang


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,479 @@
...
+
+    weight = Float(
+        display_name=_("Weight"),
+        help=_("This is the maximum score that the user receives when he/she successfully completes the problem"),
+        scope=Scope.settings,
+        default=1,
+    )
+
+    item_background_color = String(
+        display_name="Item background color",
```
cahrens: i18n

How will these strings get scraped for translation since they are not in edx-platform?

itsjeyd: @cahrens Thanks for the feedback.

We don't have any immediate plans to move the code inside edx-platform. So w/r/t to scraping strings the situation for DnDv2 will be the same as for every other XBlock that's not within the edx-platform repo: As far as I know strings will likely not be scraped, but that's due to an existing limitation of the platform that's a known issue.

dianakhuang:
:+1: from TNL on a high-level code review.

itsjeyd:
@zvishal FYI, as part of our recent effort to make DnDv2 available on edx.org we updated and refined the analytics events for this XBlock (cf. #47 and [SOL-1473](https://openedx.atlassian.net/browse/SOL-1473)). Could you please have a look and let us know if you are OK with these changes? There's an event design page [here](https://openedx.atlassian.net/wiki/display/AN/DnDv2+XBlock+Event+Design) that summarizes what the events look like now. Thanks!

zvishal:
Thanks Tim for the heads up. 

These changes look good to me.

Thanks, 
-Vishal

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+# Classes ###########################################################
+
+@XBlock.wants('settings')
+class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
+    """
+    XBlock providing a Drag and Drop question
+    """
+    display_name = String(
+        display_name=_("Title"),
+        help=_("The title of the Drag and Drop that is displayed to the user"),
```
pdesjardins: "The title of the drag and drop problem. The title is displayed to learners."

Will this help text be displayed to the course teams who are including a drag and drop problem in a course? We refer to questions as "problems" and students/users as "learners."

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        scope=Scope.settings,
+        default=_("Drag and Drop"),
+    )
+
+    show_title = Boolean(
+        display_name=_("Show title"),
+        help=_("Display the title to the user?"),
```
pdesjardins: "Display the title to the learner?"

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        scope=Scope.settings,
+        default=True,
+    )
+
+    question_text = String(
+        display_name=_("Question text"),
```
pdesjardins: "Problem text"

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        help=_("The question text (and/or instructions) shown to the user"),
```
pdesjardins: "The description of the problem or instructions shown to the learner"

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        scope=Scope.settings,
+        default="",
+    )
+
+    show_question_header = Boolean(
+        display_name=_("Show \"Question\" heading"),
```
pdesjardins: "Show \"problem\" heading"

Is the "heading" the same as the display_name "title" in line 33? If it is the same, we should call them both "title." I'm not sure what the "heading" is here.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        help=_("Display the \"Question\" heading above the question/instructions?"),
```
pdesjardins: "Display the heading \"Problem\" above the problem text?"

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        scope=Scope.settings,
+        default=True,
+    )
+
+    weight = Float(
+        display_name=_("Weight"),
+        help=_("This is the maximum score that the user receives when he/she successfully completes the problem"),
```
pdesjardins: "The maximum score the learner can receive for the problem"

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        scope=Scope.settings,
+        default=1,
+    )
+
+    item_background_color = String(
+        display_name=_("Item background color"),
+        help=_(
+            "Background color to use for draggable items. "
```
pdesjardins: "The background color of draggable items in the problem. "

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+            "Defaults to color specified by current theme, or blue if no theme is set."
```
pdesjardins: I think we should omit this line. The course team users who will see this UI might not have any control over the current theme or even understand what a theme is, right? At least, that is the way it would be for partners at edx.org. In any case, I think it's not really necessary to explicitly document the default color.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        ),
+        scope=Scope.settings,
+        default="",
+    )
+
+    item_text_color = String(
+        display_name=_("Item text color"),
+        help=_(
+            "Text color to use for draggable items. "
+            "Defaults to color specified by current theme, or white if no theme is set."
```
pdesjardins: I think we should not refer to the theme since course teams might have no idea what that is.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        ),
+        scope=Scope.settings,
+        default="",
+    )
+
+    data = Dict(
+        display_name=_("Drag and Drop"),
```
pdesjardins: Where does this string appear? The help string about the JSON spec and the builder doesn't make sense to me. Where would a user be able to see that help string?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        help=_("JSON spec as generated by the builder"),
+        scope=Scope.content,
+        default=DEFAULT_DATA,
+    )
+
+    item_state = Dict(
+        help=_("How the student has interacted with the problem"),
```
pdesjardins: What is the context for this help string? It sounds like this is a description of the state of the problem after an interaction. But where would a user be accessing that description? Is this available in a UI? Or part of an API?

Maybe: "Information describing the state of a problem after a learner has attempted it"

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -0,0 +1,480 @@
...
+        scope=Scope.user_state,
+        default={},
+    )
+
+    completed = Boolean(
+        help=_("The student has completed the problem at least once"),
```
pdesjardins: "Indicates whether the learner has completed the problem at least once"

drag_and_drop_v2/drag_and_drop_v2.py
pdesjardins: This looks like a description of the state of a problem after a learner has attempted it. Is that right?

Maybe: "A description of the state of a problem after a learner has attempted it"

drag_and_drop_v2/drag_and_drop_v2.py
pdesjardins: "Indicates whether a learner has completed the problem at least once"

drag_and_drop_v2/public/js/view.js
```diff
@@ -0,0 +1,258 @@
...
+            h('section.keyboard-help', [
+                h('a.keyboard-help-button', { attributes: { tabindex: 0 } }, gettext('Keyboard Help')),
+                h('div.keyboard-help-dialog', [
+                    h('div.modal-window-overlay'),
+                    h('div.modal-window', { attributes: dialog_attributes, style: dialog_style }, [
+                        h('div.modal-header', [
+                            h('h2.modal-window-title', gettext('Keyboard Help'))
+                        ]),
+                        h('div.modal-content', [
+                            h('p', gettext('You can complete this exercise using only your keyboard.')),
```
pdesjardins: 'You can complete this problem using only your keyboard.'

pdesjardins: 'You can complete this problem using only your keyboard.'

drag_and_drop_v2/public/js/view.js
```diff
@@ -0,0 +1,258 @@
...
+                            h('ul', [
+                                h('li', gettext('Use "Tab" and "Shift-Tab" to navigate between items and zones.')),
+                                h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')),
+                                h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" to drop the item on the current zone.')),
+                                h('li', gettext('Press "Escape" if you want to cancel the drop operation (e.g. because you would like to select a different item).')),
```
pdesjardins: 'Press "esc" if you want to cancel the drop operation (for example, to select a different item).'

drag_and_drop_v2/public/js/view.js
```diff
@@ -0,0 +1,258 @@
...
+                        ])
+                    ])
+                ])
+            ])
+        );
+    };
+
+    var mainTemplate = function(ctx) {
+        var problemHeader = ctx.show_title ? h('h2.problem-header', {innerHTML: ctx.header_html}) : null;
+        var questionHeader = ctx.show_question_header ? h('h3.title1', gettext('Question')) : null;
```
pdesjardins: I'm not sure where this string will be shown. We use the term "problem" not "question."

drag_and_drop_v2/templates/html/drag_and_drop.html
```diff
@@ -0,0 +1,4 @@
+{% load i18n %}
+<section class="themed-xblock xblock--drag-and-drop">
+    <i class="fa fa-spin fa-spinner initial-load-spinner"></i>{% trans "Loading drag and drop exercise." %}
```
pdesjardins: "Loading drag and drop problem."

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -0,0 +1,154 @@
+{% load i18n %}
+
+<div class="xblock--drag-and-drop--editor editor-with-buttons">
+    {{ js_templates|safe }}
+
+    <section class="drag-builder">
+        <div class="tab feedback-tab">
+            <p class="tab-content">
+                {% trans "Note: don't edit the question if students already answered it! Delete it and create a new one." %}
```
pdesjardins: "Note: do not edit the problem if students have already completed it. Delete the problem and create a new one."

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -0,0 +1,154 @@
...
+            </p>
+
+            <section class="tab-content">
+                <form class="feedback-form">
+                    <h3>{% trans "Question title" %}</h3>
```
pdesjardins: "Problem title"

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -0,0 +1,154 @@
...
+                    <label title="{{ help_texts.show_title }}">
+                      <input class="show-title" type="checkbox"
+                        {% if self.show_title %}checked="checked"{% endif %}>
+                      {% trans "Show title" %}
+                    </label>
+
+                    <h3>{% trans "Maximum score" %}</h3>
+                    <input type="number" step="0.1" class="weight" value="1" value="{{ self.weight }}" />
+
+                    <h3>{% trans "Question text" %}</h3>
```
pdesjardins: "Problem text"

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -0,0 +1,154 @@
...
+                    <textarea class="question-text">{{ self.question_text }}</textarea>
+                    <label title="{{ help_texts.show_question_header }}">
+                      <input class="show-question-header" type="checkbox"
+                        {% if self.show_question_header %}checked="checked"{% endif %}>
+                      {% trans "Show \"Question\" heading" %}
```
pdesjardins: "Show \"Problem\" heading"

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -0,0 +1,154 @@
...
+            </header>
+            <section class="tab-content">
+                <form class="target-image-form">
+                    <h3 id="background-url-label">
+                        {% trans "Background URL" %}
+                    </h3>
+                    <input type="text"
+                           class="url-input"
+                           aria-labelledby="background-url-label"
+                           placeholder="{% trans 'e.g. http://example.com/background.png or /static/background.png' %}">
```
pdesjardins: 'For example, http://example.com/background.png or /static/background.png'

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -0,0 +1,154 @@
...
+                    <h3 id="background-description-label">
+                        {% trans "Background description" %}
+                    </h3>
+                    <textarea class="description-input"
+                              aria-labelledby="background-description-label"
+                              aria-describedby="background-description-description"></textarea>
+                    <div id="background-description-description" class="target-image-form-help">
+                        {% blocktrans %}
+                            Please provide a description of the image for non-visual users.
+                            The description should provide sufficient information that would allow anyone
```
pdesjardins: The description should provide sufficient information to allow anyone

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -0,0 +1,121 @@
...
+        <div class="row advanced-link">
+            <a href="#">{{i18n "Show advanced settings" }}</a>
+        </div>
+        <div class="row advanced">
+            <label for="item-{{id}}-width-percent">{{i18n "Preferred width as a percentage of the background image width (or blank for automatic width):"}}</label>
+            <input type="number" id="item-{{id}}-width-percent" class="item-width" value="{{ singleDecimalFloat widthPercent }}" step="0.1" min="1" max="99" />%
+        </div>
+        <div class="row advanced">
+          <label for="item-{{id}}-numerical-value">
+            {{i18n "Optional numerical value (if you set this, students will be prompted for this value after dropping this item)"}}
```
pdesjardins: Change "students" to "learners."

"Optional numerical value (if you set this, learners will be prompted for this value after dropping this item)"

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -0,0 +1,121 @@
...
+          </label>
+          <input type="number"
+                 step="0.1"
+                 id="item-{{id}}-numerical-value"
+                 class="item-numerical-value" value="{{ numericalValue }}" />
+        </div>
+        <div class="row advanced">
+          <label for="item-{{id}}-numerical-margin">
+            {{i18n "Margin ± (when a numerical value is required, values entered by students must not differ from the expected value by more than this margin; default is zero)"}}
```
pdesjardins: Change "students" to "learners."

"Margin ± (when a numerical value is required, values entered by learners must not differ from the expected value by more than this margin; default is zero)"

drag_and_drop_v2/templates/html/js_templates.html
pdesjardins: Change "students" to "learners."

"Optional numerical value (if you set this, learners will be prompted for this value after dropping this item)"

pdesjardins: Change "students" to "learners."

"Optional numerical value (if you set this, learners will be prompted for this value after dropping this item)"

README.md
```diff
@@ -0,0 +1,428 @@
+Drag and Drop XBlock v2
+=======================
+
+This XBlock implements a friendly drag-and-drop style question, where
```
pdesjardins: Replace "question" with "problem."

README.md
```diff
@@ -0,0 +1,428 @@
...
+the student has to drag items on zones on a target image.
```
pdesjardins: Replace "student" with "learner."

README.md
```diff
@@ -0,0 +1,428 @@
...
+            }
+        }
+```
+
+You can use the same approach to apply a custom theme:
+
+`"package"` can refer to any Python package in your virtualenv, which
+means you can develop and maintain your own theme in a separate
+package. There is no need to fork or modify this repository in any way
+to customize the look and feel of your Drag and Drop exercises.
```
pdesjardins: Replace "excercises" with "problems."

pdesjardins:
This looks great! Thanks, in particular, for including such excellent documentation in the README file.

I reviewed the strings and left my comments. Please let me know if you need any information.

:+1:

drag_and_drop_v2/drag_and_drop_v2.py
itsjeyd: @pdesjardins The help text currently isn't displayed in the interactive editor, but that might change in the future.

talbs:
From a UX point of view, here are some thoughts:


**Please Address**
* When a user incorrectly drops an item into the wrong target, we should style the message that's displayed differently than any success/basic confirmation messages (currently all use a black background with white text).
* The incorrect messages disappear very quickly (this happens more often when displaying an incorrect message multiple times). We should slow this down or make the removal of the incorrect messages manual (through the close icon currently in place).
* Visually, smaller drag and drop items have extra whitespace after their left-aligned text while longer items have equal amounts of padding ([see example](https://cloud.githubusercontent.com/assets/163763/12510545/a8b252ea-c0d8-11e5-9f1f-48af85b5f50b.png)). Without getting too technical, I assume this is your ``min-width`` property in action. I'd recommend ``text-align: center`` so that even in the case of smaller items (which need the padding for their drag and drop affordance), the visual appearance always has equal whitespace.

**Questions**
* The drag and drop items currently have a button-like appearance using the edX/Open edX dark blue. are these properties theme-able/customizable currently? 
* What happens to the UI when there are more drag and drop items than the area they are initially contained in can display in one line?
* Are there any disabled states of drag and drop items that we need to account for?

Let me know when you have any revisions to or feedback to the first group of items. I'll take another look then.

drag_and_drop_v2/drag_and_drop_v2.py
itsjeyd: @pdesjardins Course authors can not edit the `data` field directly. It stores information about a problem that is derived from the input that a course author provides via the interactive editor (in Studio). This string and the help string below are not displayed to users, so the information they contain is currently only useful for developers. I updated both of them to be more specific about the purpose of the field.

drag_and_drop_v2/drag_and_drop_v2.py
itsjeyd: @pdesjardins This field is for internal use only; it's not available in any UI, and it's also not part of any (public) API. So just like the help string for `data`, the help string for this field is only used to convey information about the field to developers.

> This looks like a description of the state of a problem after a learner has attempted it. Is that right?

Yes, if "has attempted it" refers to any type of interaction (when a learner drops an item on the appropriate zone, information about the position of the item is added to `item_state`; it doesn't matter if the learner completed the problem by performing this action).

I updated the string to be more specific about the type of information that `item_state` stores; it now reads: "Information about current positions of items that a learner has dropped on the target image." Taking into account the considerations above, please let me know if you have any suggestions for further improvement.

drag_and_drop_v2/public/js/view.js
itsjeyd: @pdesjardins This string will be shown above the problem description if a course author enables the corresponding option ('Show "Problem" heading') in Studio. I changed it to "Problem".

README.md
itsjeyd: @pdesjardins Thanks for this. Searching the code base I found additional occurrences of the word "exercise(s)"; changed all of them to "problem(s)" in https://github.com/edx-solutions/xblock-drag-and-drop-v2/commit/ed3e323cffae49523fbe62f5056c2207ae38a0df. In an effort to further standardize the terminology, I also replaced remaining occurrences of "question(s)" with "problem(s)" wherever possible.

itsjeyd:
@pdesjardins Thanks a lot for reviewing! Glad you're happy with the documentation in the README :) 

I addressed your comments. If you have any additional suggestions based on the changes I made, please let me know.

itsjeyd:
@talbs Thanks for the feedback!

> The incorrect messages disappear very quickly (this happens more often when displaying an incorrect message multiple times). We should slow this down or make the removal of the incorrect messages manual (through the close icon currently in place).

I'm not sure what you mean by *this happens more often when displaying an incorrect message multiple times*. Could you list the steps to reproduce this behavior?

In general, the way this is currently implemented is that popups are dismissed as soon as the user sends any kind of mouse, touch, or keyboard input. (It is also possible to dismiss a popup by clicking the close icon.) We like the idea of closing popups automatically, since it allows users to focus on completing a problem without having to take care of popups themselves. Maybe we could do the following to address your point: Keep the ability to close popups automatically, but limit it to input events that target the problem container (or even just individual items). This would cut out a large portion of the events that currently cause a popup to disappear, while still allowing popups to politely "get out of the way" when a user continues to interact directly with a problem after dropping an item.

> Are there any disabled states of drag and drop items that we need to account for?

I'm not sure what you mean by that, could you please specify?

> The drag and drop items currently have a button-like appearance using the edX/Open edX dark blue. are these properties theme-able/customizable currently?

The color of the drag and drop items can be customized via the Studio editor; there are separate options for setting the background color and the text color. We also added theming support recently (which allows full control over the appearance of items).

> What happens to the UI when there are more drag and drop items than the area they are initially contained in can display in one line?

The block uses Flexbox with wrapping enabled to lay out items in the area above the target image, so when a row/line gets too full the items flow to the next line.

cptvitamin:
Things are great from a keyboard users perspective.  I have some comments below. I was not able to complete a full review, or a review on Windows (no access to machine today.)  However @clytwynec will take a look on Monday and get back to you with some more feedback.

* each drop zone has an ID defined.  However, if there are multiple DnD problems on the page, they are not unique.  The drop zone id values are repeated zone-1, zone-2.  This is a validation error and causes accessibility issues when referencing the ids or using them to move focus.  all IDs have to be unique.
* i see you have added tabindex=0 to the draggables.  this is great.  However to a screen reader user, the draggables are just divs.  Its not clear that they are draggables or that they can interact with them.  Try adding a role = button to them.  Also, at least using Voiceover, I could not pick up or drop any of the draggables because space, enter, cmd+m etc are all intercepted by the screen reader. For this, we have to add a role=application to the container of the draggable.  application is a powerful role. You have to be very careful with its scope. Any interactive element inside of if loses its native roles states and properties, so its up to the developer to define each one.  You ahve already done that, but you need to pass the keyboard command through to the element.
* Is there a specific use case for making the display of the title optional (checkbox in studio)?  Without a title, there is not Heading to introduce the problem, or provide quick navigation for screen reader users.
* there is a ```text-transform: uppercase```  style on ```.xblock--drag-and-drop .drag-container .target .zone p``` that is making the screen reader spell the region labels out e.g. T O P and M I D 
* all of the form fields in the Studio view are lacking proper form labels.  I would convert the ```<h3>``` 's to ```<label>``` elements.   It would be easiest to imply a relationship by wrapping the input with the label element (otherwise you have to use for and id attributes) . I also see ```title``` attributes on all of the checkboxes.  title attributes are not accessible (they display on mouse hover only). If you feel it is necessary to add additional guidance, that text should be onscreen for everyone to see or displayed offscreen for screen reader users, included in the label or reference via aria-describedby on the checkbox.
* in the studio view, the background image needs to have an alt attribute.  it should be set to the value defined by the author, or have a default value until the author defines it.  The description should be a required field to ensure that the image has a description.
* we should improve the demo example to actively demonstrate quality descriptive text.  TOP and MID does not help a non-visual user understand the context of this question . "Top third of a pyramid, Middle third of a pyramid, Bottom third of a pyramid" should be added to the example to demonstrate what would be required for a non-visual user to complete the exercise.  Both name and description should be required fields.

talbs:
:+1:

itsjeyd:
@cptvitamin Thanks for the comments! I'm handing this over to @Kelketek, so for reference, here is where we're w/r/t addressing them:

> each drop zone has an ID defined. However, if there are multiple DnD problems on the page, they are not unique. [...] all IDs have to be unique.

This is still to do.

> [...] to a screen reader user, the draggables are just divs. Its not clear that they are draggables or that they can interact with them. [...] Also, at least using Voiceover, I could not pick up or drop any of the draggables because space, enter, cmd+m etc are all intercepted by the screen reader. [...].

Addressed in https://github.com/edx-solutions/xblock-drag-and-drop-v2/commit/937f11e77af5b2fa9da9f8ad37b6235acb9150e7.

> Is there a specific use case for making the display of the title optional (checkbox in studio)? [...]

We recently added this functionality for one of our clients, cf. https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/32, so we can't remove it completely (at least not without having to maintain a separate version of DnDv2 that includes it). Please let us know if you have any ideas on how to discourage course authors from making use of this feature more actively.

> there is a text-transform: uppercase style on .xblock--drag-and-drop .drag-container .target .zone p that is making the screen reader spell the region labels out e.g. T O P and M I D

Addressed in https://github.com/edx-solutions/xblock-drag-and-drop-v2/commit/3644354d97984eb5d7c2345acb21727fb451858e.

> all of the form fields in the Studio view are lacking proper form labels. [...]. I also see title attributes on all of the checkboxes. title attributes are not accessible (they display on mouse hover only). If you feel it is necessary to add additional guidance, that text should be [...] displayed offscreen for screen reader users, [...].

Addressed in https://github.com/edx-solutions/xblock-drag-and-drop-v2/commit/3b99ed1fae4d9daf006d66ae17486c2ec9f59ab1 and https://github.com/edx-solutions/xblock-drag-and-drop-v2/commit/5b2037b0cd15aa82cf80d72c4c5582d90e086f52.

> in the studio view, the background image needs to have an alt attribute. it should be set to the value defined by the author, or have a default value until the author defines it. The description should be a required field to ensure that the image has a description.

If a description for a background image is set, the Studio editor uses it as a value for the `alt` attribute. What remains to be done here is to:

* Adjust the code to insert a default value for instances of the block that were created *before* we added support for specifying a background description (Based on your remaining feedback, I think the problem you tested was [Triangle Demo](http://master.dndv2.sandbox.opencraft.com/courses/course-v1:OpenCraft+DnDv2+demos/courseware/6915ee3dd6ab403d8c05c0ea3180a0ee/a1773b9a51164a8b98ca862edfafa68a/)? If so, this is an instance of a problem that was created prior to adding a11y support. All newly created blocks have the description for the background image set to [this](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/default_data.py#L3) by default.) What would be a good default value to use for old instances of this block?
* Add validation to ensure that course authors can't submit edits to a DnDv2 block without specifying a description.

> we should improve the demo example to actively demonstrate quality descriptive text. TOP and MID does not help a non-visual user understand the context of this question. [...] Both name and description should be required fields.

The "TOP" and "MID" labels come from a configured block ([Triangle Demo](http://master.dndv2.sandbox.opencraft.com/courses/course-v1:OpenCraft+DnDv2+demos/courseware/6915ee3dd6ab403d8c05c0ea3180a0ee/a1773b9a51164a8b98ca862edfafa68a/)). See [here](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/default_data.py#L9-L15) for default values for zone labels and descriptions (all newly created DnDv2 blocks get these). The corresponding problem on the sandbox is [Unconfigured Test](http://master.dndv2.sandbox.opencraft.com/courses/course-v1:OpenCraft+DnDv2+demos/courseware/6915ee3dd6ab403d8c05c0ea3180a0ee/a67029dc4d184bc5bc2b3c6398af13f1/). What's left to do here is to make zone labels and descriptions required fields.

clytwynec:
@itsjeyd @Kelketek, Mark is out this week, so I'll be helping with the accessibility review on this.  I see you've addressed a bunch of Mark's comments already.  Can you let me know when this is ready for me to continue reviewing?

Kelketek:
@clytwynec Yes, I will ping you once these issues are corrected. Could you also look at the questions @itsjeyd raised yesterday and give feedback on them?

Kelketek:
@clytwynec The accessibility changes should be ready for you. @talbs A minor style addition has been made for the studio when a field is in error. It makes the offending field glow red. I don't know if you'd care to look at it or not, but I figured I'd give you a heads up.

clytwynec:
@Kelketek 

> RE: Please let us know if you have any ideas on how to discourage course authors from making use of this feature more actively.
 
I'm still not sure I understand what the use case for excluding the heading would be. What was the clients reason for not wanting it included? 

> RE: What would be a good default value to use for old instances of this block?

I haven't looked closely enough at this feature yet to know.  Would this default description end up being used in the student view too if the author hasn't updated it?

Kelketek:
@clytwynec For clarity-- the changes that have been made are those that now require that alt text be provided in several fields. No change has been made to automatically supply old blocks with a default. This shouldn't affect edx.org, since it's not been released there yet.

However anyone editing an old block will now be required to supply this information to save it.

Kelketek:
> I'm still not sure I understand what the use case for excluding the heading would be. What was the clients reason for not wanting it included?

I think they just thought it was ugly or redundant in some layouts. @antoviaque can you give some insight here?

> I haven't looked closely enough at this feature yet to know. Would this default description end up being used in the student view too if the author hasn't updated it?

Yes.

clytwynec:
@Kelketek 

 > For clarity-- the changes that have been made are those that now require that alt text be provided in several fields. No change has been made to automatically supply old blocks with a default. This shouldn't affect edx.org, since it's not been released there yet. However anyone editing an old block will now be required to supply this information to save it.

I personally think this is a good solution.

clytwynec:
@Kelketek 
This is generally looking pretty good, but here's a bit more feedback on accessibility.  I'm seeing some other issues too, but I need to do a bit more testing to confirm what I think is happening. I'll follow up tomorrow on those.

* `Reset problem` button 
   * The reset button should really be a button element, not a link element. Is there a reason (other than styling) that it is not already?
   * Because the reset problem button is inside of an aria-live region, it is announced
by the screen reader when it is made visible. This seems confusing since it happens after the user puts the first draggable in a drop zone. (e.g. I think I have just put the draggable in a drop zone, and I hear "Reset problem".)

* `Keyboard help` button should also be a button element. (The rule of thumb is that if you have a link without an href value or href=“#”, it should really be a button element.)

* When the content of a draggable is an image, is there a way for the author to add an alt attribute to it?

Kelketek:
@clytwynec I've pushed changes to make the links into buttons.

Yes, you can specify in the studio alt text for draggable images.

clytwynec:
@Kelketek   Thanks for making those updates.  I had a chance to review this a bit more this afternoon and found a couple other issues. 

* color contrast is too low on items that have been placed correctly in the drop zones.

* keyboard help/reset problem buttons have no focus indicator in Firefox.

* Multiple drag-and-drop problems on a page results in duplicate IDs (‘#-zone-1’, etc.). There shouldn’t be more than one instance of an ID on a page.

* Mark mentioned before that `role=“application”` needed to go around the “draggable” region.  It looks like it also needs to the container with the drop zones also needs to have `role=“application”`, or instead put it on the `section.drag-container`, since that contains both the draggable container and drop zone container.  I was able to pick up an item, but the keystrokes for dropping an item in the drop zones were being intercepted by screen readers.   (Once the role="application" was added, this work really well with the windows screenreaders.)

* Some issues that only seem to be happening in voiceover:
	* The drop zones aren’t being recognized as focusable by voiceover.  I’m not sure why. For now, I don’t know how you can fix this, but I wanted to mention that the issue exists.
	* When dropping an item into a drop zone, I am unable to access the feedback (the black/red boxes in the triangle example) using voiceover on mac. `aria-live` regions that are loaded after the initial DOM load don’t always work in voiceover/safari. Since this seems to be an issue with the way a specific screenreader interprets valid markup, I’m not sure it is necessary to put in a workaround. However, one possible solution is to send focus to the feedback element instead of having it work as an aria-live region.

Kelketek:
@clytwynec I think I've hit everything you've listed there now. Please take another look.

clytwynec:
@Kelketek  Those changes look good to me.

* With multiple problems on a page, there are still some duplicate IDs. The one I see now is `#item-n-description` in the case where both problems have their nth item correctly dropped in a target. Can you take a look at any IDs that are being used in this block and make sure there isn’t a case where there could be multiple of the same ID on the page?

* I think the individual drop zones also need to have `role="button"` (much like the draggables). Without it the screenreader isn’t providing useful context to the user.

cahrens:
To find simple issues like multiple IDs or color contrast, I HIGHLY recommend that you add an accessibility bok choy test. It's super easy and useful-- all you have to do is navigate to a courseware page with 2 drag and drop xblocks on it and call a single method!

Kelketek:
@cahrens In this case it'd require a few extra steps, since those IDs don't get created until some interaction happens. As we're short on time I'd like to avoid introducing extra variables.

However I'm curious about this technique. Is there documentation for it? I'd like to use it in the future if I can.

@clytwynec Please take a look at the latest updates and see if they perform as you expect.

cahrens:
@Kelketek See https://openedx.atlassian.net/wiki/display/TE/Automated+Accessibility+Tests#AutomatedAccessibilityTests-WritinganAccessibilityTest and/or search in the edx-platform acceptance test directory for "check_for_accessibility_errors".

@clytwynec is the one who introduced the functionality, and the TNL team has found it very useful and simple to write.

clytwynec:
@Kelketek 

> When dropping an item into a drop zone, I am unable to access the feedback (the black/red boxes in the triangle example) using voiceover on mac. aria-live regions that are loaded after the initial DOM load don’t always work in voiceover/safari. Since this seems to be an issue with the way a specific screenreader interprets valid markup, I’m not sure it is necessary to put in a workaround. However, one possible solution is to send focus to the feedback element instead of having it work as an aria-live region.

I think you made the change I suggested here, but it seems to have made the feedback regions no longer work in windows screenreaders and not to have fixed the issue in safari/voicover. At this point, it seems better to leave it as it was (just using the aria-live region, not sending focus).  It will leave it without the feedback read aloud in voiceover without navigating to it, but allow it to work properly elsewhere. 

I think there is still one outstanding question that Mark had.

>  > Is there a specific use case for making the display of the title optional (checkbox in studio)? [...]

> We recently added this functionality for one of our clients, cf. #32, so we can't remove it completely (at least not without having to maintain a separate version of DnDv2 that includes it). Please let us know if you have any ideas on how to discourage course authors from making use of this feature more actively.

> > I'm still not sure I understand what the use case for excluding the heading would be. What was the clients reason for not wanting it included?

> I think they just thought it was ugly or redundant in some layouts. @antoviaque can you give some insight here?

Otherwise this is working really well.

antoviaque:
> > I'm still not sure I understand what the use case for excluding the heading would be. What was the clients reason for not wanting it included?
> I think they just thought it was ugly or redundant in some layouts. @antoviaque can you give some insight here?

@clytwynec Sorry that I had missed the question before - the reason they wanted the option to hide the title is that they sometimes want to be able to make titles themselves, with a different heading/formatting than the default one from the XBlock.

Kelketek:
@clytwynec I've reverted the changes to the feedback popup. Could you take another look?

clytwynec:
@Kelketek  This looks good to me except that the popup feedback still isn't being recognized by any screenreaders.   I'm out of ideas for how to get it working, since it appears to have the right markup and, looking at the code, looks like it is being rendered at the same time as other aria-live regions that are working.  Since @cptvitamin is out until Monday,  @clrux do you have any ideas what might be going on here?

clrux:
@Kelketek Remove `aria-live` from the injected popup and instead place it on the parent (`target`) like so:

```
<div class="target" aria-live="polite" aria-atomic="true" aria-relevant="additions">
```

Kelketek:
@clrux @clytwynec Fixed.

clytwynec:
@Kelketek  That works great!  Thanks for making these changes.

Kelketek:
@clytwynec Could I get your explicit +1 on this?

clytwynec:
:+1:

PR #36, opened 2015-12-03
Mark user-visible strings for translation

itsjeyd:


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -23,48 +23,48 @@ class DragAndDropBlock(XBlock):
     XBlock providing a Drag and Drop question
     """
     display_name = String(
-        display_name="Title",
-        help="The title of the Drag and Drop that is displayed to the user",
+        display_name=_("Title"),
+        help=_("The title of the Drag and Drop that is displayed to the user"),
         scope=Scope.settings,
         default="Drag and Drop"
```
bradenmacdonald: The `default` should be marked for translation too.

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -10,20 +10,20 @@
...
        <div class="layout">
-           <label>width</label>
+           <label>{{i18n "width"}}</label>
            <input type="text" class="size width" value="{{ width }}" />
-           <label>height</label>
+           <label>{{i18n "height"}}</label>
            <input type="text" class="size height" value="{{ height }}" />
            <br />
-           <label>x</label>
+           <label>{{i18n "x"}}</label>
```
bradenmacdonald: I think using `x` and `y` for cartesian axes is actually a standard mathematical notation [used](https://es.wikipedia.org/wiki/Coordenadas_cartesianas) [in](https://de.wikipedia.org/wiki/Kartesisches_Koordinatensystem) [every](https://ja.wikipedia.org/wiki/%E7%9B%B4%E4%BA%A4%E5%BA%A7%E6%A8%99%E7%B3%BB) [human](https://zh.wikipedia.org/wiki/%E7%AC%9B%E5%8D%A1%E5%84%BF%E5%9D%90%E6%A0%87%E7%B3%BB) [language](https://pl.wikipedia.org/wiki/Uk%C5%82ad_wsp%C3%B3%C5%82rz%C4%99dnych_kartezja%C5%84skich), so these do not need to be translated.

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -36,36 +36,36 @@
...
             <textarea class="success-feedback">{{ feedback.correct }}</textarea>
         </div>
         <div class="row">
-            <label>Error Feedback</label>
+            <label>{{i18n "Error Feedback"}}</label>
             <textarea class="error-feedback">{{ feedback.incorrect }}</textarea>
         </div>
         <div class="row">
-            <label>Width (px - 0 for auto)</label>
+            <label>{{i18n "Width (px - 0 for auto)"}}</label>
```
bradenmacdonald: The formatting of this string may be confusing for translators - I suggest changing it to `"Width in pixels (0 for auto)"`

bradenmacdonald:
Thanks @itsjeyd - this is :+1: if you address the minor notes I left.

itsjeyd:
@bradenmacdonald Thanks! Re: the issue with the `help` texts for `item_background_color` and `item_text_color`, I went with the approach of translating the strings in `studio_view` and passing them to the template context. Having `help` in field definitions doubles as documentation, and having `def _(self): ...` in place will make it easy to translate strings as necessary in future work.

PR #37, opened 2015-12-08
Show/hide question header

bradenmacdonald:
This change adds a new checkbox "Show 'Question' heading" to the XBlock editor wizard. The new field defaults to True (backwards compatible). When set to false, the "Question" heading (which is otherwise not editable) will be hidden.

**Before / Default**:
screen shot 2015-12-07 at 7 08 25 pm

**New setting**:
screen shot 2015-12-07 at 7 07 23 pm

**After / unchecked**:
screen shot 2015-12-07 at 7 08 10 pm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,16 +44,16 @@ coverage.xml
 *.mo
 *.pot
 
-# Django stuff:
-*.log
+# Integration test output:
+/*.log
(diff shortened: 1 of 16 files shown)


Kelketek:
:+1:

PR #38, opened 2015-12-08
Move draggable items above drop image area

itsjeyd:
**Screenshot**

drag-above-drop


drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -128,7 +133,28 @@ function DragAndDropBlock(runtime, element) {
...
+                newTop = parseFloat(item.top) - heightDelta,
+                itemElement = $('.option[data-value="' + item_id + '"]', element);
+
+            // Update top
+            top = newTop + 'px';
+            // Update state
+            item.top = top;
+            state.state.items[item_id] = item;
+            // Update position
+            itemElement.css('top', top);
```
mtyaka: You should never manually modify DOM elements. A more "correct" way to do this would be to call `setState(state)` and have virtualDom make sure the `top` value is set properly instead, but since we already call `setState` before `submitLocation`, I feel this is getting too messy. We should always try to avoid cascading updates like this, because they make it hard to understand what's getting updated where.

It would be great if we can find a different solution, one that doesn't require to measure DOM elements. I can think of two:
- Once the item is dropped, move it from `.items` to the `.target` element, so that the coordinates are always relative to the `.target` element.
- Prevent the `.items` element from shrinking as you drag items out of it. I imagine it should be possible to create invisible (with `visibility: hidden`) duplicates of each draggable item, and position the duplicates exactly underneath the visible items. The only function of the invislble duplicates would be to maintain the dimensions of the `.items` div as you drag the visible items out. 

I suspect the first one would be easier, but I'm not entirely sure.

bradenmacdonald: +1. I like the first idea above (move it to the .target element). In fact that will be almost necessary for my task, where I make the whole Xblock proportionally sized and resizable.

I think it's fine if the `.items` shrinks as you pull items out from it (that will actually improve usability if there are dozens of items, since the user can do less scrolling as the items stay close to the `.target` rather than having blank spaces appear there). However, the `items` should not shrink before or during the drag-and-drop. If it does shrink, it should shrink at the end, after the user has dropped an item. A simple way to achieve this is to set `$items.height($items.height())` in ondragstart and `$items.height('auto')` in ondragend. (Not sure if there's a more virtual-domy way to achieve that, but I know the jQuery code is already doing all kinds of direct DOM manipulation like that during the drag process anyways.)

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -44,12 +45,9 @@
...
 /** Draggable Items **/
 .xblock--drag-and-drop .drag-container .items {
-    width: 210px;
-    margin: 10px;
+    display: block;
     padding: 0 !important; /* LMS tries to override this */
     position: relative;
-    display: inline;
-    float: left;
     list-style-type: none;
```
bradenmacdonald: Is this rule doing anything? AFAIK, `.items` is a `DIV` so this would have no effect.

bradenmacdonald:
There is a bug with this:

Any existing dropped items will have the wrong position after this is merged:

Before:
screen shot 2015-12-09 at 2 07 28 pm

After:

screen shot 2015-12-09 at 2 07 34 pm

However, any fix for that in this PR will need to be re-implemented in my upcoming PR so I advise just leaving this as-is for now and I'll build my new code on top of this, then merge the combined PR.

PR #39, opened 2015-12-17
Support for multiple DnD blocks on the same page

e-kolpakov:
**Description:** this PR fixes a bug with having multiple DnD blocks on the same page - items from second and subsequent blocks, when dragged, jumped into the first block drop area.

**JIRA:** [SOL-1148](https://openedx.atlassian.net/browse/SOL-1448)

**Testing instructions:**
1. Add two or more DnD blocks in the unit.
2. Publish, open the unit in LMS.
3. Try dragging items in the first block - should work just fine (including correcr/incorrect zones and final feedback) - not changed.
4. Try dragging items in the second block. *Expected behavior:* works just fine. *Pre-fix behavior:* items jump to first block drop area; impossible to complete the second block.


tests/integration/test_interaction.py
```diff
@@ -144,21 +130,60 @@ def get_locations():
...
+        self.wait_until_html_in(feedback['intro'], self._get_feedback_message())
 
         locations_after_reset = get_locations()
         for item_key in items.keys():
             self.assertDictEqual(locations_after_reset[item_key], initial_locations[item_key])
 
 
+class InteractionTestFixture(InteractionTestBase):
+    """
+    Verifying Drag and Drop XBlock rendering against default data - if default data changes this would probably broke
```
itsjeyd: @e-kolpakov Can you change this to:

```
Verifying Drag and Drop XBlock rendering against default data - if default data changes this will probably break.
```

That would make the docstring consistent with the one in [test_render.py](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/tests/integration/test_render.py#L26-L27).

tests/integration/test_interaction.py
```diff
@@ -193,3 +218,78 @@ class CustomHtmlDataInteractionTest(InteractionTestFixture, BaseIntegrationTest)
...
+    }
+
+    all_zones = {
+        'block1': ['Zone 1', 'Zone 2'],
+        'block2': ['Zone 51', 'Zone 52']
+    }
+
+    feedback = {
+        'block1': {"intro": "Other Intro Feed", "final": "Other Final Feed"},
+        'block2': {"intro": "Some Intro Feed", "final": "Some Final Feed"},
```
itsjeyd: @e-kolpakov Could you switch the strings for initial and final feedback between the `.json` files? I found it slightly confusing that the data from `test_data.json` has "*Other* Intro Feed"/"*Other* Final Feed" while the data coming from `test_data_other.json` has "*Some* Intro Feed"/"*Some* Final Feed".

itsjeyd:
@e-kolpakov Looking good, just had a couple of small comments. :+1: once you address those.

PR #41, opened 2015-12-18
Make DnDv2 accessible

itsjeyd:
Cf. [SOL-1412](https://openedx.atlassian.net/browse/SOL-1412)

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -198,7 +198,6 @@ class DragAndDropBlock(XBlock):
         fragment.initialize_js('DragAndDropEditBlock', {
             'data': self.data,
             'target_img_expanded_url': self.target_img_expanded_url,
-            'target_img_description': self.target_img_description,
             'default_background_image_url': self.default_background_image_url,
         })
 
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -153,6 +153,7 @@ class DragAndDropBlock(XBlock):
             "question_text": self.question_text,
             "show_question_header": self.show_question_header,
             "target_img_expanded_url": self.target_img_expanded_url,
+            "target_img_description": self.target_img_description,
             "item_background_color": self.item_background_color or None,
             "item_text_color": self.item_text_color or None,
             "initial_feedback": self.data['feedback']['start'],
@@ -197,6 +198,7 @@ class DragAndDropBlock(XBlock):
         fragment.initialize_js('DragAndDropEditBlock', {
             'data': self.data,
             'target_img_expanded_url': self.target_img_expanded_url,
+            'target_img_description': self.target_img_description,
             'default_background_image_url': self.default_background_image_url,
         })
 
@@ -327,6 +329,11 @@ class DragAndDropBlock(XBlock):
             return self._expand_static_url(self.data["targetImg"])
         else:
             return self.default_background_image_url
+
+    @property
+    def target_img_description(self):
+        """ Get the description for the target image (the image items are dragged onto). """
+        return self.data.get("targetImgDescription", "")
 
     @property
     def default_background_image_url(self):
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -23,7 +23,7 @@
 /* Shared styles used in header and footer */
 
 .xblock--drag-and-drop .title1 {
-    color: rgb(85, 85, 85);
+    color: #555555;
     text-transform: uppercase;
     font-weight: bold;
     font-style: normal;
@@ -58,7 +58,7 @@
     align-items: center;
 
     position: relative;
-    border: 1px solid rgba(0,0,0, 0.1);
+    border: 1px solid rgba(0, 0, 0, 0.1);
     border-radius: 3px;
     padding: 5px;
 }
@@ -72,7 +72,7 @@
     border-radius: 3px;
     margin: 5px;
     padding: 10px;
-    background-color: #2e83cd;
+    background-color: #2872b2;
     font-size: 14px;
     color: #fff;
     opacity: 1;
@@ -112,8 +112,14 @@
     transform: translate(-50%, -50%); /* These blocks are to be centered on their absolute x,y position */
 }
 
+/* Focused option */
+.xblock--drag-and-drop .drag-container .item-bank .option:focus,
+.xblock--drag-and-drop .drag-container .item-bank .option:hover {
+    transform: scale(1.1);
+}
+
 .xblock--drag-and-drop .drag-container .ui-draggable-dragging {
-    box-shadow: 0 16px 32px 0 rgba(0,0,0,.3);
+    box-shadow: 0 16px 32px 0 rgba(0, 0, 0, 0.3);
     border: 1px solid #ccc;
     opacity: .65;
     z-index: 20 !important;
@@ -159,7 +165,7 @@
 
 .xblock--drag-and-drop .drag-container .option .numerical-input.correct .input {
     background: #ceffce;
-    color: #0dad0d;
+    color: #087108;
 }
 
 .xblock--drag-and-drop .drag-container .option .numerical-input.incorrect .input {
@@ -224,8 +230,12 @@
 
 }
 
+/* Focused zone */
+.xblock--drag-and-drop .zone:focus {
+    border: 2px solid #a5a5a5;
+}
+
 .xblock--drag-and-drop .zone p {
-    visibility: hidden;
     width: 100%;
     font-family: Arial;
     font-size: 16px;
@@ -250,7 +260,7 @@
 /*** FEEDBACK ***/
 .xblock--drag-and-drop .feedback {
     margin-top: 20px;
-    border-top: solid 1px rgb(189, 189, 189);
+    border-top: solid 1px #bdbdbd;
 }
 
 .xblock--drag-and-drop .popup {
@@ -258,8 +268,8 @@
     display: none;
     top: 5%;
     right: 5%;
-    border: 1px solid white;
-    background: none repeat scroll 0 0 rgba(0,0,0,.8);
+    border: 1px solid #fff;
+    background: none repeat scroll 0 0 rgba(0, 0, 0, 0.8);
     width: 500px;
     max-width: 90%;
     min-height: 50px;
@@ -269,7 +279,7 @@
 }
 
 .xblock--drag-and-drop .popup .popup-content {
-    color: #FFFFFF;
+    color: #ffffff;
     margin-left: 15px;
     margin-top: 35px;
     margin-bottom: 15px;
@@ -282,7 +292,7 @@
     margin-right: 8px;
     margin-top: 8px;
     margin-left: 20px;
-    color: #FFFFFF;
+    color: #ffffff;
     font-family: "fontawesome";
     font-size: 18pt;
 }
@@ -290,6 +300,20 @@
 .xblock--drag-and-drop .reset-button {
     cursor: pointer;
     float: right;
-    color: #3384CA;
+    color: #2d74b3;
     margin-top: 3px;
+}
+
+/* Make sure screen-reader content is hidden in the workbench: */
+.xblock--drag-and-drop .sr {
+    border: 0;
+    clip: rect(0 0 0 0);
+    height: 1px;
+    margin: -1px;
+    overflow: hidden;
+    padding: 0;
+    position: absolute;
+    width: 1px;
+    background: #ffffff;
+    color: #000000;
 }
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -72,10 +72,12 @@
     border-radius: 3px;
     margin: 5px;
     padding: 10px;
-    background-color: #2872b2;
+    background-color: #1d5280;
     font-size: 14px;
     color: #fff;
     opacity: 1;
+    outline-color: #fff;
+    outline-style: none;
     /* Some versions of the drag and drop library try to fiddle with this */
     z-index: 10 !important;
 }
@@ -115,7 +117,8 @@
 /* Focused option */
 .xblock--drag-and-drop .drag-container .item-bank .option:focus,
 .xblock--drag-and-drop .drag-container .item-bank .option:hover {
-    outline: 2px solid #fff;
+    outline-width: 2px;
+    outline-style: solid;
     outline-offset: -4px;
 }
 
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -115,7 +115,8 @@
 /* Focused option */
 .xblock--drag-and-drop .drag-container .item-bank .option:focus,
 .xblock--drag-and-drop .drag-container .item-bank .option:hover {
-    transform: scale(1.1);
+    outline: 2px solid #fff;
+    outline-offset: -4px;
 }
 
 .xblock--drag-and-drop .drag-container .ui-draggable-dragging {
--- a/drag_and_drop_v2/public/css/drag_and_drop_edit.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop_edit.css
@@ -141,9 +141,12 @@
     width: 18%;
 }
 
-.xblock--drag-and-drop--editor .zones-form .zone-row .title {
+.xblock--drag-and-drop--editor .zones-form .zone-row .title,
+.xblock--drag-and-drop--editor .zones-form .zone-row .description {
     width: 60%;
     margin: 0 0 5px;
+    line-height: 2.664rem;  /* .title gets line-height from a Studio rule that does not apply to .description;
+                               here we make sure that both input fields get the same value for line-height */
 }
 
 .xblock--drag-and-drop--editor .zones-form .zone-row .layout {
@@ -162,6 +165,7 @@
     height: 128px;
 }
 
+.xblock--drag-and-drop--editor .target-image-form .target-image-form-help,
 .xblock--drag-and-drop--editor .item-styles-form .item-styles-form-help {
     margin-top: 5px;
     font-size: small;
@@ -173,7 +177,7 @@
 }
 
 .xblock--drag-and-drop--editor .items-form .item {
-    background: #73bde7;
+    background: #8fcaec;
     padding: 10px 0 1px;
     margin: 15px 0;
 }
@@ -209,11 +213,12 @@
 
 /** Buttons **/
 .xblock--drag-and-drop--editor .btn {
-    background: #2e83cd;
+    background: #2872b2;
     color: #fff;
     border: 1px solid #156ab4;
     border-radius: 6px;
     padding: 5px 10px;
+    margin-top: 15px;
 }
 
 .xblock--drag-and-drop--editor .btn:hover {
@@ -228,7 +233,7 @@
 
 .xblock--drag-and-drop--editor .add-element {
     text-decoration: none;
-    color: #2e83cd;
+    color: #236299;
 }
 
 .xblock--drag-and-drop--editor .remove-zone {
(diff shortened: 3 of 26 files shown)


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -197,6 +198,7 @@ def studio_view(self, context):
         fragment.initialize_js('DragAndDropEditBlock', {
             'data': self.data,
             'target_img_expanded_url': self.target_img_expanded_url,
+            'target_img_description': self.target_img_description,
```
bradenmacdonald: If this is already in `data.targetImgDescription `, is there any reason to pass it separately here?

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -83,8 +83,14 @@
     transform: translate(-50%, -50%); /* These blocks are to be centered on their absolute x,y position */
 }
 
+/* Focused option */
+.xblock--drag-and-drop .drag-container .item-bank .option:focus,
+.xblock--drag-and-drop .drag-container .item-bank .option:hover {
+    transform: scale(1.1);
```
bradenmacdonald: This is kinda cool, but...

In a case like this, I don't think it looks good when adjacent items meld together:
screen shot 2015-12-18 at 11 37 46 am

And sometimes it doesn't actually stand out that much. e.g. in this example, it's not immediately obvious which block has focus:
screen shot 2015-12-18 at 11 37 52 am

I think that a better visual effect would be simply a colored border, which is more standard. Try something like `    outline: 2px solid white; outline-offset: -4px;` and see what you think.

I also think that including this effect with `hover` may be a bit overkill/distracting, but a more subtle hover effect would be nice, and helps users know what they can interact with and what they cannot.

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -141,9 +141,12 @@
     width: 18%;
 }
 
-.xblock--drag-and-drop--editor .zones-form .zone-row .title {
+.xblock--drag-and-drop--editor .zones-form .zone-row .title,
+.xblock--drag-and-drop--editor .zones-form .zone-row .description {
```
bradenmacdonald: Can you simplify this to `.xblock--drag-and-drop--editor .zones-form .zone-row input`, or would that have unwanted effects?

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -315,8 +320,9 @@
 
 .xblock--drag-and-drop--editor .remove-item .icon.remove {
     background: #fff;
+    color: #0072a7;
 }
 .xblock--drag-and-drop--editor .remove-item .icon.remove:before,
 .xblock--drag-and-drop--editor .remove-item .icon.remove:after {
-    background: #2e83cd;
+    background: #236299;
```
bradenmacdonald: What is the reason for all these subtle color tweaks? Also some of the new colors look the same to me but are using different hex values - should they not be made more consistent?

screen shot 2015-12-18 at 11 56 26 am

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -298,12 +299,16 @@ function DragAndDropBlock(runtime, element, configuration) {
                     input.class_name = item_user_state.correct_input ? 'correct' : 'incorrect';
                 }
             }
+            var content_html = item.displayName;
+            if (item.backgroundImage) {
+                content_html = '<img src="' + item.backgroundImage + '" alt="' + item.backgroundDescription + '" />';
```
bradenmacdonald: I know this was already coded this way, but it would be better to just pass `displayName`, `backgroundImage`, and `backgroundDescription` to the template and move this conditional logic and HTML generation there (`view.js`).

Also, `item.backgroundDescription` needs to be HTML escaped - you'll get that for free with the move to view.js.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -298,12 +299,16 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+            }
             var itemProperties = {
                 value: item.id,
                 drag_disabled: Boolean(item_user_state || state.finished),
                 class_name: item_user_state && ('input' in item_user_state || item_user_state.correct_input) ? 'fade': undefined,
                 input: input,
-                content_html: item.backgroundImage ? '<img src="' + item.backgroundImage + '"/>' : item.displayName
+                content_html: content_html
```
bradenmacdonald: Nit: We are only supporting ES5+ browsers, so you can add a trailing comma to objects/dicts like this if you'd like nicer diffs in the future, just like in Python :)

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -70,9 +70,11 @@ function DragAndDropEditBlock(runtime, element, params) {
                         }
 
                         // Set the target image and bind its event handler:
-                        $('.target-image-form input', element).val(_fn.data.targetImg);
+                        $('.target-image-form input.url-input', element).val(_fn.data.targetImg);
+                        $('.target-image-form input.description-input', element).val(_fn.data.targetImgDescription);
                         _fn.build.$el.targetImage.load(_fn.build.form.zone.imageLoaded);
                         _fn.build.$el.targetImage.attr('src', params.target_img_expanded_url);
+                        _fn.build.$el.targetImage.attr('alt', params.target_img_description);
```
bradenmacdonald: I asked this earlier obliquely, but is there any difference between `_fn.data.targetImgDescription` and `params.target_img_description`?

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -401,7 +414,8 @@ function DragAndDropEditBlock(runtime, element, params) {
                                         width: width,
                                         height: height
                                     },
-                                    backgroundImage: backgroundImage
+                                    backgroundImage: backgroundImage,
+                                    backgroundDescription: backgroundDescription
```
bradenmacdonald: Nit: can add a trailing comma here

drag_and_drop_v2/public/js/view.js
```diff
@@ -50,13 +51,21 @@
...
         return (
             h('div.option',
                 {
                     key: item.value,
                     className: item.class_name,
-                    attributes: {'data-value': item.value, 'data-drag-disabled': item.drag_disabled},
+                    attributes: {
+                        'tabindex': tabindex,
+                        'draggable': true,
+                        'aria-grabbed': false,
```
bradenmacdonald: This is good, but you also need to extend the 'startDragging' and 'endDragging' events to set this to `aria-grabbed=true` when the block is picked up and actively being dragged.

drag_and_drop_v2/public/js/view.js
```diff
@@ -50,13 +51,21 @@
         if (item.is_placed) {
             style.left = item.x_percent + "%";
             style.top = item.y_percent + "%";
+            tabindex = -1;  // If an item has been placed it can no longer be interacted with,
```
bradenmacdonald: In the future (https://openedx.atlassian.net/browse/SOL-1462 number 6) we will be adding `sr` descriptions to the dropped items as well, and it's important for screen reader users to be able to read the description of dropped items... but I'm not sure if they should have tabindex="0" or "-1" in that case, since they can't be interacted with.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -41,18 +41,39 @@
...
+                    </h3>
+                    <input type="text"
+                           class="description-input"
+                           aria-labelledby="background-description-label"
+                           aria-describedby="background-description-description">
+                    <div id="background-description-description" class="target-image-form-help">
+                        {% blocktrans %}
+                            Please provide a description of the image for non-visual users.
+                            The description should provide sufficient information that would allow anyone
+                            to solve the problem if the image did not load.
```
bradenmacdonald: "to solve the problem even without seeing the image" perhaps?

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -5,27 +5,33 @@
         width:{{ width_percent }}%;
         height:{{ height_percent }}%;">
```
bradenmacdonald: Since this is focusable, I'm wondering if we should give it aria-describedby="" that references the `p.description` element below.

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -5,27 +5,33 @@
...
-           <input type="text" class="size width" value="{{ width }}" />
-           <label>{{i18n "height"}}</label>
-           <input type="text" class="size height" value="{{ height }}" />
-           <br />
-           <label>x</label>
-           <input type="text" class="coord x" value="{{ x }}" />
-           <label>y</label>
-           <input type="text" class="coord y" value="{{ y }}" />
-       </div>
+        <label>{{i18n "Text"}}</label>
```
bradenmacdonald: The `label` elements in this section are not linked to the `input` elements (by wrapping them or with `for` attributes)

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -46,7 +52,11 @@
         </div>
         <div class="row">
             <label>{{i18n "Background image URL (alternative to the text)"}}</label>
-            <textarea class="background-image">{{ backgroundImage }}</textarea>
+            <textarea class="background-image-url">{{ backgroundImage }}</textarea>
```
bradenmacdonald: Can you please convert this to an `<input type="text"` while you're at this? I don't think it should be a `textarea`.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -41,18 +41,39 @@
...
-                    <input type="text">
+                <form class="target-image-form">
+                    <h3 id="background-url-label">
+                        {% trans "Background URL" %}
+                    </h3>
+                    <input type="text" class="url-input" aria-labelledby="background-url-label">
+                    <h3 id="background-description-label">
+                        {% trans "Background description" %}
+                    </h3>
+                    <input type="text"
```
bradenmacdonald: Should be a `textarea`.

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -46,7 +52,11 @@
         </div>
         <div class="row">
             <label>{{i18n "Background image URL (alternative to the text)"}}</label>
```
bradenmacdonald: I think we should actually just call this field "Image URL" since it is not used as a background - it replaces the text.

In fact, now that I think about it, we could consider using the text (`displayName`) as the description when an image is present, rather than needing a separate field. You could just say "If you put an image, it will be used instead of the text above. The text above should describe the image so that the exercise can be completed by users who are not able to see the image."

But I'm also fine with keeping the description separate - it's probably cleaner conceptually - and "Text" should be an input while "Description" a textarea so that's another reason to leave it separate.

tests/integration/data/test_data_a11y.json
```diff
@@ -0,0 +1,63 @@
+{
+    "zones": [
+        {
+            "index": 1,
+            "title": "Zone 1",
+            "description": "This describes zone 1",
```
bradenmacdonald: I'm worried about having too many fixture files, which arguably makes the tests harder to maintain when we change the data format. Could you just add a `description` to one of the existing fixtures? (unless that would break existing tests)

tests/integration/test_render.py
```diff
@@ -105,21 +104,36 @@ def test_items(self):
...
-
         color_settings = {}
         if item_background_color:
             color_settings['background-color'] = item_background_color
         if item_text_color:
             color_settings['color'] = item_text_color
 
+        self._test_items(color_settings=color_settings)
+
+    def _test_items(self, color_settings={}):  # pylint: disable=dangerous-default-value
```
bradenmacdonald: Please use `color_settings=None` as the default (together with `color_settings = color_settings or {}` in the function), rather than suppressing the warning.

tests/integration/test_render.py
```diff
@@ -128,19 +142,39 @@ def test_zones(self):
...
+            self.assertEqual(zone.get_attribute('aria-dropeffect'), 'move')
+            self.assertEqual(zone.get_attribute('data-zone'), 'Zone {}'.format(zone_number))
+            self.assertIn('ui-droppable', self.get_element_classes(zone))
+            zone_box_percentages = box_percentages[index]
+            self._assert_box_percentages(
+                '#zone-{}'.format(zone_number),
+                left=zone_box_percentages['left'],
+                top=zone_box_percentages['top'],
+                width=zone_box_percentages['width'],
+                height=zone_box_percentages['height']
```
bradenmacdonald: Just use `self._assert_box_percentages('#zone-{}'.format(zone_number), **zone_box_percentages)` ?

drag_and_drop_v2/public/css/drag_and_drop.css
itsjeyd: @bradenmacdonald I like the idea of modifying `outline` on `:focus`/`:hover`. The thing I'm worried about is this: What if a course author chooses a background color for items that is similar to the outline color? In that case it would be hard to tell which item currently has focus. Maybe it would make sense to combine the `outline` and `transform` effects on `:focus` (and only use `outline` on `:hover` to make the effect more subtle in that case)? This would also alleviate the problem of items melding together somewhat.

I pushed a version that only includes the changes you are suggesting, but let me know what you think re: combining the effects.

The code for the approach I am suggesting would look something like this:

```css
/* Focused option */
.xblock--drag-and-drop .drag-container .item-bank .option:focus,
.xblock--drag-and-drop .drag-container .item-bank .option:hover {
    outline: 2px solid #fff;
    outline-offset: -4px;
}

.xblock--drag-and-drop .drag-container .item-bank .option:focus {
    transform: scale(1.1);
    box-shadow: 0 16px 32px 0 rgba(0, 0, 0, 0.3);
}
```

drag_and_drop_v2/public/css/drag_and_drop_edit.css
itsjeyd: @bradenmacdonald Yes, I changed the colors to increase contrast. Using [this tool](http://webaim.org/resources/contrastchecker/) to check contrast values I found that many pairs of foreground and background colors did not pass the check for WCAG AA. To make the changes as subtle as possible I stopped lightening (or darkening) individual colors as soon as the tool reported that a given pair was passing for WCAG AA. For some pairs of colors adjusting the contrast involved a larger number of steps than for others. That's why some of the colors look similar but have slightly different values.

drag_and_drop_v2/templates/html/js_templates.html
itsjeyd: @bradenmacdonald I'm not 100% sure what you mean by `p.description`, but assuming that you're talking about referencing the

```html
<p class="sr">{{{ description }}}</p>
```

element, I don't think that's necessary: Since the elements that hold the `title` and `description` of a zone are children of the `div` that represents the zone, screen readers should automatically associate their text nodes with the `div`.

tests/integration/data/test_data_a11y.json
itsjeyd: @bradenmacdonald Using any one of the existing fixtures breaks the tests in `test_render.py` (before/without the changes that this PR introduces) since they are relying on a different configuration of items and zones than the tests that use existing fixtures.

drag_and_drop_v2/public/css/drag_and_drop.css
bradenmacdonald: @itsjeyd You could also just put:

```
.xblock--drag-and-drop .drag-container .item-bank .option {
    outline-color: white;
    outline-style: none;
}
.xblock--drag-and-drop .drag-container .item-bank .option:hover,
.xblock--drag-and-drop .drag-container .item-bank .option:focus {
    outline-width: 2px;
    outline-style: solid;
}
```

And then on the element itself in the HTML, you can use an inline `style="outline-color: {{textColor}}"` so that the outline will match the text color, which should always be a contrasting color.

drag_and_drop_v2/public/css/drag_and_drop_edit.css
mtyaka: What you did makes sense, but I think I would prefer to keep the number of different colors the same as before this patch - all colors that used to be `#2e83cd` should be changed to the same new color, for example - if that's possible. It might not be possible if you had to brighten it in some cases and darken in others.

drag_and_drop_v2/public/js/view.js
```diff
@@ -50,16 +51,28 @@
...
+                            // so remove the ability to move focus to it using the keyboard
+        }
+        var content_html = item.displayName;
+        if (item.imageURL) {
+            content_html = '<img src="' + item.imageURL + '" alt="' + item.imageDescription + '" />';
```
mtyaka: I can't remember why the `` element was originally constructed from a string, but I think now we can just use the `h` function instead:

    content_ytml = h('img', {src: item.imageURL, alt: item.imageDescription})

drag_and_drop_v2/public/js/view.js
mtyaka: @itsjeyd Instead of manually toggling the `aria-grabbed` attribute in drag start and stop event handlers, can we instead set a flag in the state and set the attribute here based on that flag? Something like `state.items[item_id].grabbed = true; applyState()` inside the `start` handler and then `attributes: {'aria-grabbed': item.grabbed, ....}` here in the view function.

It's true that during the drag the virtual dom representation doesn't exactly match the actual dom (the position of the dragged item is controlled by jquery dnd plugin rather than virtual dom while the drag is in progress), but it still makes sense to try and minimize manual dom manipulation as much as possible.

drag_and_drop_v2/public/css/drag_and_drop.css
itsjeyd: @bradenmacdonald That's a good idea. Implemented in [672aa8c](https://github.com/edx-solutions/xblock-drag-and-drop-v2/commit/https://github.com/open-craft/xblock-drag-and-drop-v2/commit/672aa8c0b72e80d7932e7160ce7666767b57b023).

drag_and_drop_v2/public/js/view.js
itsjeyd: @mtyaka Thanks for the update. We do support HTML inside `displayName`, see [test_html_data.json](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/a11y/tests/integration/data/test_html_data.json#L24).

drag_and_drop_v2/public/js/view.js
itsjeyd: @mtyaka What you are suggesting won't work with `state.items[item_id]` and `applyState` because

* the item being dragged only gets added to `state.items` *after* being dropped (which makes sense considering that `state.items` keeps track of items that have already been placed)
* `applyState` will destroy the draggable mid-drag (by calling `destroyDraggable`)

I pushed a version that uses `configuration.items` and `updateDOM` instead.

mtyaka:
Thanks for the updates, it looks great. :+1: once you get the tests to pass.

drag_and_drop_v2/public/js/view.js
bradenmacdonald: @cptvitamin Quick question for you about this drag and drop block: Once a draggable is dropped onto a drop zone, the draggable cannot be interacted with but will still display useful information (e.g. its `<span class="sr">` text will say "The 'circle' item, dropped onto the 'round shapes' zone".

Since it cannot be interacted with at that point, should it have `tabindex=-1`? Or should it have a tabindex since it represents a UI element that contains useful information and was previously interactive?

cptvitamin: @bradenmacdonald just to be clear, a tabindex of -1 only makes an item focusable by javascript, so unless you had some other control that would move focus to that item, -1 would have no affect for the user.  tabindex of 0 will add a non-focusable element (like a div) into the default tab order.  If it cannot be interacted with, I say that it shouldn't have a tabindex at all.  Even if it has useful information, the user can still access the info using standard navigation (just not TAB navigation.)  Sound good?

bradenmacdonald: @cptvitamin Thanks - that answers my question perfectly.

@itsjeyd When an item is correctly placed, can you remove the tabindex completely instead of setting it to `-1`? Either approach should work for us here but as Mark mentions, that is preferable. If it's too much work, the current approach should also be fine and this is good to go :)

PR #42, opened 2015-12-21
Ajax loader spinner

e-kolpakov:
**Description:** this PR adds loader spinners to elements that trigger ajax request.

**Testing instructions:**
1. Loader spinner is added to initial DnD view (the one with "Loading drag and drop exercise")
2. Loader spinner is added to an item when it is dragged into a zone (both correct and incorrect), and removed from it when ajax request finishes

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -76,6 +76,16 @@
     z-index: 10 !important;
 }
 
+.xblock--drag-and-drop .drag-container .option .spinner-wrapper {
+    margin-right: 3px;
+    float: left;
+    display: none;
+}
+
+.xblock--drag-and-drop .drag-container .option .item-content {
+    display: inline-block;
+}
+
 /* Placed option */
 .xblock--drag-and-drop .drag-container .target .option {
     position: absolute;
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -112,6 +112,19 @@
     top: calc(50% - 16px);
 }
 
+.xblock--drag-and-drop .drag-container .option .numerical-input .spinner-wrapper {
+    position: absolute;
+    right: 0;
+    top: 0;
+    margin-right: 5px;
+    padding-top: 6px;
+    display: none;
+}
+
+.xblock--drag-and-drop .drag-container .option .numerical-input .spinner-wrapper {
+    color: #333;
+}
+
 .xblock--drag-and-drop .drag-container .option .numerical-input .input {
     display: inline-block;
     width: 144px;
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -5,6 +5,10 @@
     padding: 0;
 }
 
+.xblock--drag-and-drop .initial-load-spinner {
+    margin-right: 3px;
+}
+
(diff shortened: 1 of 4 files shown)


drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -102,6 +112,19 @@
     top: calc(50% - 16px);
 }
 
+.xblock--drag-and-drop .drag-container .option .numerical-input .spinner-wrapper {
+    position: absolute;
+    right: 0;
+    top: 0;
+    margin-right: 5px;
+    padding-top: 6px;
```
itsjeyd: @e-kolpakov Any specific advantage(s) of using `padding-top` instead of `margin-top` here? (Just asking because I'm curious :))

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -102,6 +112,19 @@
...
+    display: none;
+}
+
+.xblock--drag-and-drop .drag-container .option .numerical-input .spinner-wrapper {
```
itsjeyd: @e-kolpakov I think you can merge this rule with the previous one; the selectors are identical.

drag_and_drop_v2/templates/html/drag_and_drop.html
```diff
@@ -1,4 +1,4 @@
 {% load i18n %}
 <section class="xblock--drag-and-drop">
-    {% trans "Loading drag and drop exercise." %}
+    <i class="fa fa-spin fa-spinner"></i>{% trans "Loading drag and drop exercise." %}
```
itsjeyd: @e-kolpakov There is no/very little horizontal space between the spinner and the message. Can you make the right-hand margin of the spinner shown in this context consistent with the right-hand margin of the spinner that is shown when dropping an item?

itsjeyd:
@e-kolpakov The spinners work well, and aside from the things I pointed out above the code LGTM as well. I like the idea of showing spinners right where users would be focusing their attention when dropping an item or submitting an additional answer! The only aspect I'm not entirely convinced about is how they behave for items that contain images instead of text -- it looks a bit weird to have the spinner displayed above the image. Do you think it would be possible to display the spinner on top of the image instead (sort of like an overlay)?

mtyaka:
@e-kolpakov Instead of calling `show`/`hide` on the spinner div imperatively, the more correct virtual dom inspired approach would be to set a flag in the `state` (`state.items[item_id].xhr_active = true`), and then show/hide the spinner div inside `view.js` based on the `xhr_active` flag; something like:

    h('div.spinner-wrapper', {style: {display: item.xhr_active ? 'block' : 'none'}}, [...])

In fact, we already have `item.submitting_location` and `item.submitting_input` that are supposed to be `true` when the xhr request that submits the location or input data is active, so at least in theory `item.submitting_location || item.submitting_input` should return `true` when any xhr requests related to the item are in progress and `false` otherwise.

drag_and_drop_v2/public/css/drag_and_drop.css
e-kolpakov: @itsjeyd the spinner is shown as if it is inside the text input, so it's just for the symmetry: both spinner wrapper and text input have the same padding.

itsjeyd:
@e-kolpakov Looking great, thanks for the additional changes! :+1: from me once you remove that leftover comment. But since I don't know virtualdom that well, I'll let @mtyaka give the final go-ahead for merging.

mtyaka:
@e-kolpakov Thanks for the changes, they look good! :1 once you get the test passing.

PR #44, opened 2016-01-07
Make DnDv2 keyboard-accessible

itsjeyd:
Cf. [SOL-1462](https://openedx.atlassian.net/browse/SOL-1462)

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -13,7 +13,6 @@ function DragAndDropBlock(runtime, element, configuration) {
     var __vdom = virtualDom.h();  // blank virtual DOM
 
     // Keyboard accessibility
-    var CTRL = 17;
     var ESC = 27;
     var RET = 13;
     var SPC = 32;
@@ -21,7 +20,6 @@ function DragAndDropBlock(runtime, element, configuration) {
     var M = 77;
     var QUESTION_MARK = 63;
 
-    var ctrlDown = false;
     var placementMode = false;
     var $selectedItem;
     var $focusedElement;
@@ -213,16 +211,17 @@ function DragAndDropBlock(runtime, element, configuration) {
         });
     };
 
-    var isCycleKey = function(key) {
-        return !ctrlDown && key === TAB;
+    var isCycleKey = function(evt) {
+        return !evt.ctrlKey && !evt.metaKey && evt.which === TAB;
     };
 
-    var isCancelKey = function(key) {
-        return !ctrlDown && key === ESC;
+    var isCancelKey = function(evt) {
+        return !evt.ctrlKey && !evt.metaKey  && evt.which === ESC;
     };
 
-    var isActionKey = function(key) {
-        if (ctrlDown) {
+    var isActionKey = function(evt) {
+        var key = evt.which;
+        if (evt.ctrlKey || evt.metaKey) {
             return key === M;
         }
         return key === RET || key === SPC;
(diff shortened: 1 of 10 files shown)


tests/integration/test_interaction.py
```diff
@@ -104,13 +199,19 @@ def parameterized_item_positive_feedback_on_bad_input(self, items_map, scroll_do
 
         for definition in self._get_correct_item_for_zone(items_map).values():
             if definition.input:
-                self.drag_item_to_zone(definition.item_id, definition.zone_id)
+                if use_keyboard:
```
e-kolpakov: @itsjeyd nit: this if use_keyboard: ... else: ... is repeated multiple times without any changes - might make sense to extract into method. Also, you can use `if action_key is not None` to have three parameters instead of four

e-kolpakov:
@itsjeyd nit: I believe it would be helpful if selected item was highlighted somehow, to indicate which one is currently selected. It won't make any difference for screen reader users, but could be helpful for users with motion disabilities (i.e. can't control mouse).

e-kolpakov:
@itsjeyd :+1: conditional two minor nits are addressed and tests are passed. Overall looks good - well done! :)

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -12,6 +12,20 @@ function DragAndDropBlock(runtime, element, configuration) {
     var state = undefined;
     var __vdom = virtualDom.h();  // blank virtual DOM
 
+    // Keyboard accessibility
+    var CTRL = 17;
+    var ESC = 27;
+    var RET = 13;
+    var SPC = 32;
+    var TAB = 9;
+    var M = 77;
```
e-kolpakov: @itsjeyd I believe it is only use when "dragging" an item into a zone using keyboard, in "Ctrl+M" chord, as an alternative to just hitting space or return. And this feature is not mentioned in Keyboard Help. I woould suggest either adding "Ctrl+M" to keyboard help or remove this `M` altogether.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -12,6 +12,20 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+    var QUESTION_MARK = 63;
+
+    var ctrlDown = false;
```
e-kolpakov: @itsjeyd I don't think I understand what this `CtrlDown` does. If it is there to check for "Crtl+Something", there's [better way](http://stackoverflow.com/questions/2767126/how-to-detect-that-ctrlr-was-pressed) (basically, `e.ctrlKey`) to do that than global variable.

drag_and_drop_v2/public/js/drag_and_drop.js
itsjeyd: @e-kolpakov Went for the first option :)

itsjeyd:
@e-kolpakov Thanks again for the feedback!

> ... when I've placed the two items that have zones into their corresponding zones, I can no longer use neither drag and drop nor keyword to place the third item. Is it supposed to be so?

Yes, that's how it's supposed to work -- once an exercise has been completed, it should not be possible to interact with items anymore. But you're right, the fact that decoy item(s) that remain in the item bank still *look* like they can be interacted with is confusing. It's been a while but I'm pretty sure I've seen it work differently in the past. (After completing an exercise, any items that remained in the item bank were "greyed out".) I've brought that behavior back now (and extended the tests to make sure we don't break it again).

e-kolpakov:
@itsjeyd :+1:

PR #45, opened 2016-01-07
Add a checkbox to the zone tab to control visibility of zone borders.…

agaylard:
# Add a checkbox to the zone tab to control visibility of zone borders.  Render the borders accordingly.

## Issues addressed
This change addresses:
* https://openedx.atlassian.net/browse/SOL-1405
* https://tasks.opencraft.com/browse/OC-1161.

## User-visible changes

* This screenshot shows the newly-added "Zone borders" checkbox control in the Drag and Drop editor:
oc-1161-0

* This screenshot shows the newly-added borders of the drop-zones displayed to a student (when the checkbox is ticked):
oc-1161-1

## Test prodedure
1. Sign in to Studio as as the "staff@example.com" user.
2. Click on "edX Demonstration Course". This opens the course outline page.
3. Scroll to "Example Week 1: Getting Started".  Click and expand it if necessary.
4. Select a subsection, e.g. " Homework - Question Styles".
5. Click "New Unit".  Give the new unit a name: "Test drag and drop borders"
6. Click "Advanced" → "Drag and drop".  The drag and drop XBlock loads.
7. Click "Continue".  The zones pane of the wizard loads.
8. Scroll down.  **Confirm that the "Zone borders" checkbox is displayed**.  Tick it.
9. Click "Continue" → "Save".  The XBlock is displayed again.
10. **Confirm that the borders of the zones appear, as in the second screenshot above.**
11. Click "Edit" and repeat steps 1-10, this time with the box unticked.
12. **Confirm that the borders of the zones no longer appear.**
13. Click "Preview".  **Confirm that the XBlock displays correctly in LMS.**

## Legal
This PR is covered by OpenCraft's contributor agreement


drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -209,6 +209,7 @@
 
 .xblock--drag-and-drop .zone {
     position: absolute;
+    border: 1px hidden #f00;
```
itsjeyd: @agaylard I'm not sure about using `#f00` (red) for the border color. I would go for something a little more "neutral", i.e., a shade of `#666` (which is currently used for borders in the Studio editor). For accessibility reasons, we do want good contrast relative to the default image. Something like `#565656` or even `#373737` would work. When combined with the color of the darkest portion of the background image (`#c9db38`), the former passes WCAG AA for normal text, and the latter passes WCAG AAA for normal text. (I used [this tool](http://webaim.org/resources/contrastchecker/) to check contrast values.)

When you change the color here, please adjust the color used for borders in the Studio editor as well.

drag_and_drop_v2/public/js/view.js
```diff
@@ -116,6 +116,7 @@
                     style: {
                         top: zone.y_percent + '%', left: zone.x_percent + "%",
                         width: zone.width_percent + '%', height: zone.height_percent + "%",
+                        borderStyle: ctx.display_zone_borders ? 'dotted' : 'hidden'
```
itsjeyd: @agaylard Setting the default border style in `drag_and_drop.css` and changing it here accomplishes what we want, but there's another approach that would be completely compliant with the task description:

1. Introduce an additional CSS class such as `zone-with-borders`:

   ```css
   .xblock--drag-and-drop .zone {
       border: 1px hidden #565656;
   }
   ```

2. Instead of setting the `borderStyle` here, conditionally add `.zone-with-borders` to the "selector" argument that gets passed to the [`h` function](https://github.com/Matt-Esch/virtual-dom/blob/master/virtual-hyperscript/README.md#hselector-properties-children) (`div.zone` in L107 above):

   ```js
   var selector = ctx.display_zone_borders ? 'div.zone.zone-with-borders' : 'div.zone';
   return (
       h(
           selector,
           {  // ...
   ```

itsjeyd:
@agaylard Nice work! This is looking pretty good already :) I left a couple of small comments for you to address. You'll also need to fix some of the existing tests to get a green build. Let me know if you have any questions about that.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -209,6 +209,7 @@
 
 .xblock--drag-and-drop .zone {
     position: absolute;
+    border: 1px hidden #565656;
```
itsjeyd: @agaylard Now that you have the `zone-with-borders` class, I think you can remove this line.

agaylard: Good catch! Fixed.

drag_and_drop_v2/public/js/view.js
```diff
@@ -115,7 +116,7 @@
                     },
                     style: {
                         top: zone.y_percent + '%', left: zone.x_percent + "%",
-                        width: zone.width_percent + '%', height: zone.height_percent + "%",
+                        width: zone.width_percent + '%', height: zone.height_percent + "%"
```
itsjeyd: @agaylard Nit: Can you bring the trailing comma back (to make future diffs nicer)? I know this isn't supported in some older browsers, but since we are only supporting ES5+ browsers, including it won't be a problem.

tests/integration/test_render.py
```diff
@@ -33,7 +33,11 @@ class TestDragAndDropRender(BaseIntegrationTest):
     PAGE_ID = 'drag_and_drop_v2'
     ITEM_PROPERTIES = [{'text': '1'}, {'text': '2'}, {'text': 'X'}, ]
 
-    def load_scenario(self, item_background_color="", item_text_color=""):
+    def load_scenario(self, item_background_color="", item_text_color="", borders=False):
+        if borders:
+            json = "integration/data/test_data_a11y_borders.json"
+        else:
+            json = "integration/data/test_data_a11y.json"
```
itsjeyd: @agaylard To avoid having to touch multiple files when fixture data changes in the future, we should try and keep the number of fixture files minimal. Instead of introducing `test_data_a11y_borders.json`, you can parameterize the existing file (`test_data_a11y.json`) as follows:

1. Instead of providing an explicit value for `"displayBorders"` (and `"displayLabels"`, if you're going to add tests for that as well), use a placeholder:

   ```js
     "displayBorders": {display_borders_value},
     "displayLabels": {display_labels_value}
   ```

2. In `load_scenario`, replace the placeholder(s) based on the value of `display_zone_borders` (and `display_zone_labels`):

   ```python
   def load_scenario(
       self, item_background_color="", item_text_color="", display_zone_borders=False, display_zone_labels=False
   ):
       data = load_resource("integration/data/test_data_a11y.json")
       data.replace('{display_borders_value}', 'true' if zone_borders else 'false')
       data.replace('{display_labels_value}', 'true' if zone_labels else 'false')
       scenario_xml = """
           <vertical_demo>
               <drag-and-drop-v2 item_background_color='{item_background_color}'
                                 item_text_color='{item_text_color}'
                                 data='{data}' />
           </vertical_demo>
       """.format(
           item_background_color=item_background_color,
           item_text_color=item_text_color,
           data=data
       )
   ```

tests/integration/test_render.py
```diff
@@ -57,6 +61,10 @@ def _get_style(self, selector, style, computed=True):
             query = 'return $("{selector}").get(0).style.{style}'
         return self.browser.execute_script(query.format(selector=selector, style=style))
 
+    def _get_border_style(self, elt, style):
+        query = 'return window.getComputedStyle(document.getElementById("{elt}"), null).getPropertyValue("{style}")'
+        return self.browser.execute_script(query.format(elt=elt, style=style))
+
```
itsjeyd: @agaylard I think you can do without this method -- `_get_style` above should do what you want, you just need to change the inputs that you are passing in `test_borders_inactive` and `test_borders_active` a little bit (I'll give more detail in another comment).

tests/integration/test_render.py
```diff
@@ -194,3 +202,21 @@ def test_background_image(self):
         image_path = '/resource/drag-and-drop-v2/public/img/triangle.png'
         self.assertTrue(bg_image.get_attribute("src").endswith(image_path))
         self.assertEqual(bg_image.get_attribute("alt"), 'This describes the target image')
+
+    def test_borders_inactive(self):
+        self.load_scenario()
+        zones = self._get_zones()
+        for index, zone in enumerate(zones):
+            elt = 'zone-{}'.format(index + 1)
```
itsjeyd: @agaylard To make this work as an input for `_get_style`, do:

```python
selector = `#zone-{}'.format(index+1)
```

This should enable `$("{selector}")` in `_get_style` to pick up the correct element.

tests/integration/test_render.py
```diff
@@ -194,3 +202,21 @@ def test_background_image(self):
...
+            # Firefox does not provide a single "border-style", so we check "border-top-style"
+            self.assertEqual(self._get_border_style(elt, 'border-top-style'), 'none')
+            self.assertEqual(self._get_border_style(elt, 'border-top-width'), '0px')
```
itsjeyd: @agaylard To make this work with `_get_style`, pass the names of the JS attributes that corresponds to the CSS properties you want to check:

```python
self.assertEqual(self._get_style(zone_selector, 'borderTopStyle'), 'none')
self.assertEqual(self._get_style(zone_selector, 'borderTopWidth'), '0px')
```

(As you probably already guessed, there is a general rule underlying this: To translate names of CSS properties to the corresponding JS attributes, remove the dashes and camelCase individual components.)

> Firefox does not provide a single "border-style", so we check "border-top-style"

AFAIK it's not just Firefox that does this -- `border-style` is a shorthand for setting the styles of each of the four borders of an element to the same value, so browsers translate `border-style: dotted` into four separate properties (`border-{top,right,bottom,left}-style`). To make the tests more restrictive, it might make sense to iterate over a list of these properties and assert that each one of them has the correct value. (The same applies to `border-width` and `border-color`.)

tests/integration/test_render.py
```diff
@@ -194,3 +202,21 @@ def test_background_image(self):
...
+
+    def test_borders_active(self):
+        self.load_scenario("", "", True)
```
itsjeyd: @agaylard Since all parameters of `load_scenario` have default values (i.e., they are keyword arguments), you don't have to pass values for `item_background_color` and `item_text_color` here. Just do:

```python
self.load_scenario(borders=True)
```

tests/integration/test_render.py
```diff
@@ -194,3 +202,21 @@ def test_background_image(self):
...
+        zones = self._get_zones()
+        for index, zone in enumerate(zones):
+            elt = 'zone-{}'.format(index + 1)
+            self.assertEqual(self._get_border_style(elt, 'border-top-style'), 'dotted')
+            self.assertEqual(self._get_border_style(elt, 'border-top-width'), '1px')
+            self.assertEqual(self._get_border_style(elt, 'border-top-color'), 'rgb(86, 86, 86)')  # == #565656
```
itsjeyd: @agaylard For readability, can you add another color (such as `DARK_GREY`) to the `Colors` class above and check against that here (instead of specifying the RGB and hex values here)?

tests/integration/test_render.py
```diff
@@ -194,3 +199,36 @@ def test_background_image(self):
         image_path = '/resource/drag-and-drop-v2/public/img/triangle.png'
         self.assertTrue(bg_image.get_attribute("src").endswith(image_path))
         self.assertEqual(bg_image.get_attribute("alt"), 'This describes the target image')
+
+    def test_borders_hidden(self):
+        self.load_scenario()
+        zones = self._get_zones()
+        for index in range(len(zones)):
```
itsjeyd: @agaylard Nit: I think the version using `enumerate(zones)` that you were using earlier is much more readable. If you name a variable `_` or `dummy`, Pylint will not produce an error about it being unused (cf. [this answer](http://stackoverflow.com/a/10107410/1199226) on SO), which means that this:

```python
for index, dummy in enumerate(zones, start=1):
```

should pass the quality checks.

The `start=1` part let's you use the simpler `.format(index)` instead of `.format(index+1)`.

tests/integration/test_render.py
```diff
@@ -32,18 +33,22 @@ class TestDragAndDropRender(BaseIntegrationTest):
     PAGE_TITLE = 'Drag and Drop v2'
     PAGE_ID = 'drag_and_drop_v2'
     ITEM_PROPERTIES = [{'text': '1'}, {'text': '2'}, {'text': 'X'}, ]
+    SIDES = ['Top', 'Bottom', 'Left', 'Right']
 
-    def load_scenario(self, item_background_color="", item_text_color=""):
+    def load_scenario(self, item_background_color="", item_text_color="", zone_labels=False, zone_borders=False):
+        json = load_resource("integration/data/test_data_a11y.json")
```
itsjeyd: @agaylard It would be more accurate to name this variable `data` (and to also use that name in the `scenario_xml` template below): DnDv2 blocks have a `data` field, and this line loads the value for this field. (Also, `load_resource` calls `pkg_resources.resource_string`, which I think is agnostic to the format of the data that it loads).

tests/integration/test_render.py
```diff
@@ -194,3 +199,36 @@ def test_background_image(self):
         image_path = '/resource/drag-and-drop-v2/public/img/triangle.png'
         self.assertTrue(bg_image.get_attribute("src").endswith(image_path))
         self.assertEqual(bg_image.get_attribute("alt"), 'This describes the target image')
+
+    def test_borders_hidden(self):
```
itsjeyd: @agaylard To make it immediately obvious what's being tested you could change the names of the new tests to start with `test_zone_`. This can also be an advantage when having to debug tests (you'd know right away that it's *zone* borders that are causing trouble).

tests/integration/test_render.py
```diff
@@ -194,3 +199,36 @@ def test_background_image(self):
...
+        self.load_scenario()
+        zones = self._get_zones()
+        for index in range(len(zones)):
+            z = '#zone-{}'.format(index + 1)
+            for s in self.SIDES:
```
itsjeyd: @agaylard Any specific reason why you're not using `selector` for `z` and `side` for `s` in these two lines?

tests/integration/test_render.py
```diff
@@ -194,3 +199,36 @@ def test_background_image(self):
...
+        for zone in zones:
+            zone_name = zone.find_element_by_css_selector('p.zone-name')
+            self.assertEqual(zone_name.get_attribute('class'), 'zone-name sr')
+
+    def test_labels_shown(self):
+        self.load_scenario(zone_labels=True)
+        zones = self._get_zones()
+        for zone in zones:
+            zone_name = zone.find_element_by_css_selector('p.zone-name')
+            self.assertEqual(zone_name.get_attribute('class'), 'zone-name')
```
itsjeyd: @agaylard This is great, thanks for adding this test and the one above! One small change that you could make here is the following: In this line

```python
zone_name = zone.find_element_by_css_selector('p.zone-name')
```

you are already asserting that `zone` contains a `<p>` element that has a class of `zone-name` (the call to `find_element_by_css_selector` would error out if that wasn't the case), so it's not necessary to check that again with `self.assertEqual(...)`. What you *do* want to assert, however, is that the element does *not* have a class of `sr`:

```python
self.assertNotIn('sr', zone_name.get_attribute('class'))
```

To be consistent could then also change the last line of the previous test to:

```python
self.assertIn('sr', zone_name.get_attribute('class'))
```

tests/integration/test_render.py
agaylard: No specific reason, only force of habit. I'll change s → side and z → zone.

tests/integration/test_render.py
agaylard: Right — that's a better way. Thanks, I'll go with it.

tests/integration/test_render.py
agaylard: @itsjeyd I did have this as "data", but pylint complains:

    $ pylint drag_and_drop_v2 tests
    ************* Module tests.integration.test_render
    W: 39,8:TestDragAndDropRender.load_scenario: Redefining name 'data' from outer scope (line 1)

Any suggestions?

itsjeyd:
@agaylard Looking great! :+1:

PR #47, opened 2016-01-13
Update and refine analytics events and their documentation.

Kelketek:
This updates the readme to include the current analytics events, including sample output from the logs and a description of when the events are fired. @stroilova would you be able to look at this, per https://openedx.atlassian.net/browse/SOL-1473 , or tell me who can? I'll be updating this PR with test coverage per the ticket, but I need to know if the existing events are fine to begin with or if more need to be added, and if so, under what circumstances.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,10 +157,7 @@ Fired when the Drag and Drop XBlock is finished loading.
     "host": "precise64",
     "referer": "http://example.com/courses/course-v1:DnD+DnD+DnD/courseware/ec546c58d2f447b7a9223c57b5de7344/756071f8de7f47c3b0ae726586ebbe16/1?activate_block_id=block-v1%3ADnD%2BDnD%2BDnD%2Btype%40vertical%2Bblock%40d2fc47476ca14c55816c4a1264a27280",
     "accept_language": "en;q=1.0, en;q=0.5",
-    "event": {
-        "component_id": "6b80ce1e8b78426898b47a834d72ffd3",
-        "user_id": 5
-    },
+    "event": {},
     "event_source": "server",
     "context": {
         "course_user_tags": {},
@@ -195,8 +192,6 @@ Fired when a student picks up a draggable item.
     "accept_language": "en;q=1.0, en;q=0.5",
     "event": {
         "item_id": 0,
-        "component_id": "6b80ce1e8b78426898b47a834d72ffd3",
-        "user_id": 5
     },
     "event_source": "server",
     "context": {
@@ -231,8 +226,6 @@ Fired when a student drops a draggable item.
     "referer": "http://example.com/courses/course-v1:DnD+DnD+DnD/courseware/ec546c58d2f447b7a9223c57b5de7344/756071f8de7f47c3b0ae726586ebbe16/1?activate_block_id=block-v1%3ADnD%2BDnD%2BDnD%2Btype%40vertical%2Bblock%40d2fc47476ca14c55816c4a1264a27280",
     "accept_language": "en;q=1.0, en;q=0.5",
     "event": {
-        "component_id": "6b80ce1e8b78426898b47a834d72ffd3",
-        "user_id": 5,
         "is_correct_location": true,
         "is_correct": true,
         "location": "The Top Zone",
@@ -273,9 +266,7 @@ Fired when the feedback pop-up is opened.
     "accept_language": "en;q=1.0, en;q=0.5",
     "event": {
         "content": "Correct! This one belongs to The Top Zone.",
-        "component_id": "6b80ce1e8b78426898b47a834d72ffd3",
         "truncated": false,
-        "user_id": 5
     },
     "event_source": "server",
     "context": {
@@ -311,8 +302,6 @@ Fired when the feedback popup is closed.
     "accept_language": "en;q=1.0, en;q=0.5",
     "event": {
         "content": "No, this item does not belong here. Try again."
-        "component_id": "13d1b859a2304c858e1810ccc23f29b2",
-        "user_id": 5,
         "manually": true
         "truncated": false,
     },
--- a/README.md
+++ b/README.md
@@ -137,6 +137,200 @@ any zone.
 
 You can define an arbitrary number of drag items.
 
+Analytics Events
+----------------
+
+The following analytics events are provided by this block.
+
+## `xblock.drag-and-drop-v2.loaded`
+
+Fired when the Drag and Drop XBlock is finished loading.
+
+### Example
+
+```
+{
+    "username": "staff",
+    "event_type": "xblock.drag-and-drop-v2.loaded",
+    "ip": "10.0.2.2",
+    "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:43.0) Gecko/20100101 Firefox/43.0",
+    "host": "precise64",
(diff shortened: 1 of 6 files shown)


Kelketek:
@itsjeyd This is ready for you.

@bradenmacdonald Note that this changes the names of some events, per @stroilova 's request. Do you know if this is likely to cause any issues?

bradenmacdonald:
@Kelketek Not sure - deferring to @lgerhat

bradenmacdonald:
@Kelketek You've got some pylint issues to fix here.

antoviaque:
@bradenmacdonald @Kelketek Note that this XBlock is actively used on the McKinsey instance, so if event names change it would be important to coordinate it with their team, to make sure they are aware too.

CC @zvishal

README.md
```diff
@@ -101,6 +101,200 @@ any zone.
 
 You can define an arbitrary number of drag items.
 
+Analytics Events
```
itsjeyd: @Kelketek As a general comment/question, would it make sense to exclude

> "common" fields that are not interesting in this context

from the examples below? (I.e., use the *first* example from the [event design page](https://openedx.atlassian.net/wiki/display/AN/DnDv2+XBlock+Event+Design) for each event)

README.md
```diff
@@ -101,6 +101,200 @@ any zone.
...
+## `xblock.drag-and-drop-v2.loaded`
+
+Fired when the Drag and Drop XBlock is finished loading.
+
+### Example
+
+```
+{
+    "username": "staff",
+    "event_type": "xblock.drag-and-drop-v2.loaded",
```
itsjeyd: @Kelketek I think you'll need to adjust these `"event_type"` entries to use underscores instead of dashes.

README.md
```diff
@@ -101,6 +101,200 @@ any zone.
...
+    "ip": "10.0.2.2",
+    "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:43.0) Gecko/20100101 Firefox/43.0",
+    "host": "precise64",
+    "referer": "http://localhost:8000/courses/course-v1:DnD+DnD+DnD/courseware/ec546c58d2f447b7a9223c57b5de7344/756071f8de7f47c3b0ae726586ebbe16/1?activate_block_id=block-v1%3ADnD%2BDnD%2BDnD%2Btype%40vertical%2Bblock%40d2fc47476ca14c55816c4a1264a27280",
```
itsjeyd: @Kelketek Is this an Apros URL? In the LMS I only get 

    http://localhost:8000/courses/course-v1:DnD+DnD+DnD/courseware/ec546c58d2f447b7a9223c57b5de7344/756071f8de7f47c3b0ae726586ebbe16/

for the `loaded` event. If we can't remove common fields from the example events, it might make sense to shorten these `referer` URLs.

README.md
```diff
@@ -101,6 +101,200 @@ any zone.
...
+        "path": "/courses/course-v1:DnD+DnD+DnD/xblock/block-v1:DnD+DnD+DnD+type@drag-and-drop-v2+block@6b80ce1e8b78426898b47a834d72ffd3/handler/do_attempt"
+    },
+    "time": "2016-01-13T01:58:45.202313+00:00",
+    "page": "x_module"
+}
+```
+
+## `xblock.drag-and-drop-v2.feedback.opened`
+
+Fired when the feedback pop-up is opened.
```
itsjeyd: @Kelketek Nit: `pop-up` ⟶ `popup` (just to make it consistent with the way it is spelled throughout the rest of the README).

README.md
```diff
@@ -101,6 +101,200 @@ any zone.
...
+    "ip": "10.0.2.2",
+    "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:43.0) Gecko/20100101 Firefox/43.0",
+    "host": "precise64",
+    "referer": "http://localhost:8000/courses/course-v1:DnD+DnD+DnD/courseware/ec546c58d2f447b7a9223c57b5de7344/756071f8de7f47c3b0ae726586ebbe16/1?activate_block_id=block-v1%3ADnD%2BDnD%2BDnD%2Btype%40vertical%2Bblock%40d2fc47476ca14c55816c4a1264a27280",
+    "accept_language": "en;q=1.0, en;q=0.5",
+    "event": {
+        "content": "Correct! This one belongs to The Top Zone.",
+        "component_id": "6b80ce1e8b78426898b47a834d72ffd3",
+        "user_id": 5
+    },
```
itsjeyd: @Kelketek Comparing this to the example from the [event design page](https://openedx.atlassian.net/wiki/display/AN/DnDv2+XBlock+Event+Design) I think this is missing the `truncated` property.

README.md
```diff
@@ -101,6 +101,200 @@ any zone.
...
+    "ip": "10.0.2.2",
+    "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:43.0) Gecko/20100101 Firefox/43.0",
+    "host": "precise64",
+    "referer": "http://localhost:8000/courses/course-v1:DnD+DnD+DnD/courseware/ec546c58d2f447b7a9223c57b5de7344/756071f8de7f47c3b0ae726586ebbe16/1?activate_block_id=block-v1%3ADnD%2BDnD%2BDnD%2Btype%40vertical%2Bblock%40d2fc47476ca14c55816c4a1264a27280",
+    "accept_language": "en;q=1.0, en;q=0.5",
+    "event": {
+        "component_id": "13d1b859a2304c858e1810ccc23f29b2",
+        "user_id": 5,
+        "manually": true
+    },
```
itsjeyd: @Kelketek This seems to be missing `truncated` and `content` properties.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -20,6 +20,9 @@ function DragAndDropBlock(runtime, element, configuration) {
     var M = 77;
     var QUESTION_MARK = 63;
 
+    // Event string size limit.
+    var MAX_LENGTH = 255;
+
```
itsjeyd: @Kelketek Can you move this block above the `// Keyboard accessibility` block please? The three vars below (`placementMode`, `$selectedItem`, `$focusedElement`) belong to the "Keyboard accessibility" section, and they are now separated from it.

tests/integration/test_interaction.py
```diff
@@ -336,6 +344,86 @@ def test_keyboard_help(self):
...
+                event_data, {
+                    'component_id': u'drag-and-drop-v2.drag-and-drop-v2.d0.u0',
+                    'user_id': 'student_1',
+                    'item_id': 0,
+                }
+        )
+
+    def test_dropped(self):
+        self.parameterized_item_positive_feedback_on_good_move(self.items_map)
+        # Skipping to 3, since 2 is grade event.
```
itsjeyd: @Kelketek Maybe add a docstring to the class that lists the order of events (`loaded` + sequence of events that happen when you do a "good move")? You wouldn't need this comment anymore, and similar information for the remaining tests in this class would be available in one place.

README.md
Kelketek: How old is your upstream devstack? The activate_block_id parameter was merged in a few months ago. It's not from Apros.

README.md
Kelketek: We could just verbatim copy the design page. I put the actual output because it would be reflective of what all the person handling analytics would be able to get, which seemed most important to me.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -19,6 +19,9 @@ function DragAndDropBlock(runtime, element, configuration) {
     var TAB = 9;
     var M = 77;
 
+    // Event string size limit.
+    var MAX_LENGTH = 255;
+
```
itsjeyd: @Kelketek This is still in the wrong place. Can you move it above the `// Keyboard accessibility` section, so that the latter is not broken up by this block?

itsjeyd:
@Kelketek Looking good! :+1: once you fix that last little issue I pointed out above.

PR #48, opened 2016-01-13
Make DnDv2 themable

itsjeyd:
This PR adds the necessary infrastructure for applying custom themes to DnDv2. It also makes legacy (Apros) styling of the block available as an opt-in theme.

**Screenshot**

Appearance of a DnDv2 exercise with legacy theme enabled:

apros-theme

**Testing**

1. Verify that DnDv2 uses default colors when legacy theme is not enabled.
2. Follow instructions in updated README to enable legacy theme, then verify that DnDv2 uses styles shown in screenshot above.

### DIFF
--- a/drag_and_drop_v2/public/themes/apros.css
+++ b/drag_and_drop_v2/public/themes/apros.css
@@ -1,5 +1,5 @@
 .themed-xblock.xblock--drag-and-drop {
-    background: #fff;
+    background-color: #fff;
 }
 
 /* Shared styles used in header and footer */
@@ -14,7 +14,7 @@
 /* drag-container holds the .item-bank and the .target */
 
 .themed-xblock.xblock--drag-and-drop .drag-container {
-    background: #ebf0f2;
+    background-color: #ebf0f2;
 }
 
 .themed-xblock.xblock--drag-and-drop .item-bank {
@@ -32,12 +32,12 @@
 }
 
 .themed-xblock.xblock--drag-and-drop .drag-container .option .numerical-input.correct .input {
-    background: #ceffce;
+    background-color: #ceffce;
     color: #087108;
 }
 
 .themed-xblock.xblock--drag-and-drop .drag-container .option .numerical-input.incorrect .input {
-    background: #ffcece;
+    background-color: #ffcece;
     color: #ad0d0d;
 }
 
@@ -48,7 +48,7 @@
 /*** DROP TARGET ***/
 
 .themed-xblock.xblock--drag-and-drop .target {
-    background: #fff;
+    background-color: #fff;
 }
 
 .themed-xblock.xblock--drag-and-drop .zone p {
@@ -66,7 +66,7 @@
 }
 
 .themed-xblock.xblock--drag-and-drop .popup {
-    background: none repeat scroll 0 0 #66a5b5;
+    background-color: #66a5b5;
 }
 
 .themed-xblock.xblock--drag-and-drop .popup .popup-content {
--- a/drag_and_drop_v2/public/themes/apros.css
+++ b/drag_and_drop_v2/public/themes/apros.css
@@ -1,0 +1,88 @@
+.themed-xblock.xblock--drag-and-drop {
+    background: #fff;
+}
+
+/* Shared styles used in header and footer */
+
+.themed-xblock.xblock--drag-and-drop .title1 {
+    color: #555555;
+    text-transform: uppercase;
+    font-weight: bold;
+    font-style: normal;
+}
+
+/* drag-container holds the .item-bank and the .target */
+
+.themed-xblock.xblock--drag-and-drop .drag-container {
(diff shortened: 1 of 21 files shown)


drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -123,7 +123,7 @@
         </div>
         <div class="row">
           <label for="item-{{id}}-numerical-margin">
-            {{i18n "Margin ± (when a numerical value is required, values entered by students must not differ from the expected value by more than this margin; default is zero)"}}
+            {{i18n "Margin +/- (when a numerical value is required, values entered by students must not differ from the expected value by more than this margin; default is zero)"}}
```
bradenmacdonald: Why this change? Personally I prefer the correct symbol, but don't have a strong opinion.

setup.py
```diff
@@ -31,7 +31,6 @@ def package_data(pkg, root_list):
         'xblock-utils',
         'ddt'
     ],
-    dependency_links = ['http://github.com/edx/xblock-utils/tarball/master#egg=xblock-utils'],
```
bradenmacdonald: Something has broken the Travis build. I think it might be this change?

drag_and_drop_v2/public/themes/apros.css
```diff
@@ -0,0 +1,88 @@
...
+.themed-xblock.xblock--drag-and-drop .popup .popup-content {
+    color: #ffffff;
+    font-size: 14px;
+}
+
+.themed-xblock.xblock--drag-and-drop .popup .close {
+    cursor: pointer;
+    color: #ffffff;
+    font-family: "fontawesome";
+    font-size: 18pt;
```
bradenmacdonald: These rules (and probably others in this file) are identical to the unthemed version - can you remove them from this file? (It's good that these things can be overridden by the theme, but we don't need to actually include new values for rules that aren't different here in the apros theme.)

drag_and_drop_v2/public/themes/apros.css
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/*** FEEDBACK ***/
+
+.themed-xblock.xblock--drag-and-drop .feedback {
+    border-top: solid 1px #bdbdbd;
+}
+
+.themed-xblock.xblock--drag-and-drop .popup {
+    background: none repeat scroll 0 0 #66a5b5;
```
bradenmacdonald: I think you can just use `background-color: #66a5b5;` here and it has the same effect.

drag_and_drop_v2/templates/html/js_templates.html
itsjeyd: @bradenmacdonald 

After replacing `load_resource` with `ResourceLoader.load_unicode` in the [line](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/add-theming/drag_and_drop_v2/drag_and_drop_v2.py#L180) that pulls in `js_templates.html` for the EDIT dialog, I got the following error:

    Error: 'ascii' codec can't decode byte 0xc2 in position 5557: ordinal not in range(128)

The difference between `load_resource` and `load_unicode` from xblock-utils is that the former wasn't calling `unicode` on the content pulled from a resource before returning it.

Changing the symbol seemed like a better option than keeping `load_resource` just for the sake of avoiding this error, but let me know -- maybe there's another way to approach this?

drag_and_drop_v2/public/themes/apros.css
itsjeyd: @bradenmacdonald The reason I'm making these styles explicit here is to make sure they stay the same even if the default styles change. Based on your comment I'm not completely sure if that's a valid concern? If you think it's unlikely that the defaults will change in the future, let me know and I'll remove the rules/declarations that are identical to the unthemed version.

bradenmacdonald: @itsjeyd Ok, fair enough.

drag_and_drop_v2/templates/html/js_templates.html
bradenmacdonald: @itsjeyd Hmm, you'd think a method called `load_unicode` would support loading unicode text :-p

Your workaround is fine for now, so we can get this merged on time. But the proper fix is to change xblock-utils to use `return unicode(resource_content, 'utf-8')` instead of just `return unicode(resource_content)`. Would you be able to open a second PR to fix that in xblock-utils?

itsjeyd: Already mentioned this on OC-1164 but for reference, https://github.com/edx/xblock-utils/pull/36 implements the fix for this issue.

bradenmacdonald:
:+1: Great work @itsjeyd !

Let's merge this now. I've made a note on the big PR, #34, that we can restore the "±" character if the xblock-utils fix makes it into Dogwood.

PR #49, opened 2016-01-15
Mobile support: Improve draggable width options for more flexible authoring

bradenmacdonald:
Previously, I removed support for authors setting a fixed pixel width/height on draggable items, since it conflicted with the new responsive background image that fits to the browser window. This also improved the experience on mobile, allowing draggables to shrink on smaller screens in a responsive way, controlled by the browser.

But since some courses [will] have drag and drop exercises that work better when authors can set a fixed width on the draggable items, I've restored that functionality, but converted it so that authors now specify an optional "preferred width" as a percentage, if they want to. If they leave that field blank (the default), a smart responsive sizing algorithm will be used instead.

I've made the new field hidden to discourage authors from setting a width when not necessary.

**Changes**:
* In Studio: New "Width (%)" field, to allow authors to specify a specific width on elements (as a percentage since the image is responsive and sometimes gets shrunk down)
* In Studio: Convert from "Width (px)" to "Width (%)" as needed
* In the LMS: If a "Width (%)" (or old "Width (px)") is set, then use that as the preferred/hinted width of the draggable block.
* Fixed a bug: a past commit has renamed `backgroundImage` to `imageURL` but it made previous values of that setting appear to be blank in Studio. (The images still showed up in the LMS though)
* Hid some uncommonly used settings behind a new "Show advanced settings" link.
* Added a workaround to resolve a seemingly intractable difference in how Webkit and Firefox handle ambiguous box models involving absolutely positioned images inside a div with a max-width. The responsive auto-sizing of draggables should now be consistent on all supported browsers.
* Marked the block as "multi-device capable" so that it can be used within the edX mobile app

**Expected results**:
* When authors specify a percentage width (within reason, i.e. 10% - 90%), it is respected
* Otherwise, smart sizing is used - draggables are sized based on their content and the screen size, using the following algorithm: "the draggable should be as wide as possible but never wider than ~30% of the background image width and never wider than the natural size of the text or image that is this draggable's content." (and also never narrower than `4em`, which overrides all previous conditions)


agaylard:
Hi @bradenmacdonald ,  this looks good.  The only comment I have is that these files have CR-LF endings:

```
tests/integration/data/200x200.svg
tests/integration/data/400x300.svg
tests/integration/data/dnd-bg-square.svg
tests/integration/data/dnd-bg-wide.svg
```

Since `tests/integration/data/60x60.svg` doesn't, I'm not sure if this is what you intended.

agaylard:
@bradenmacdonald the .apk problem remains: I have tried it on a 4.2.2 Samsung S4 phone and a 4.0 no-name tablet, and @itsjeyd has tried it on his phone, all without success.

agaylard:
@bradenmacdonald :+1:

PR #50, opened 2016-01-15
Fix keyboard help hotkey issues

itsjeyd:
This PR fixes the following bugs affecting user interaction with DnDv2:

* [x] If multiple DnDv2 blocks are on the same page, pressing hotkey (`?`) brings up multiple dialogs.
* [x] It's almost impossible to enter `?` into any of the text boxes in the block settings in Studio, since it brings up the keyboard help dialog instead.

It also extends existing tests for interacting with the "Keyboard Help" dialog:

* `interact_with_keyboard_help` now checks if dialog can be dismissed by pressing ESC.
* `MultipleBlocksDataInteraction` test now includes a `test_keyboard_help` method which verifies that keyboard help works correctly when multiple DnDv2 blocks are present.

**Test instructions**

1. In the LMS, verify that pressing "?" does *not* bring up the "Keyboard Help" dialog -- we decided to remove this functionality (cf. [this comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/50#commitcomment-15478277)).
2. In Studio, try to edit a DnDv2 exercise and enter `?` into a few different input fields. You should be able to do that without the "Keyboard Help" dialog popping up.
3. Lastly, using a unit that includes more than one DnDv2 exercise, check that clicking "Keyboard Help" (in the LMS) does not bring up multiple dialogs.

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -47,9 +47,12 @@ function DragAndDropBlock(runtime, element, configuration) {
             initDroppable();
 
             $(document).on('keydown mousedown touchstart', closePopup);
-            $(document).on('keypress', function(evt) {
-                runOnKey(evt, QUESTION_MARK, showKeyboardHelp);
-            });
+            if (isFirstExercise()) {  // Set up handler for "?" key only once,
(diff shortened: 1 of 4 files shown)


agaylard:
@itsjeyd :+1:

PR #51, opened 2016-01-18
Ensure correct alignment of zone labels; remove IE9 shims

itsjeyd:
Changes in this PR ensure that zone labels are centered vertically in the LMS. The PR also removes shims for IE9, which we don't need to support anymore.

**Screenshots**

Before:

zone-labels-before

After:

zone-labels-after


drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -249,19 +249,9 @@
...
-    display: table;
-}
-
-.lt-ie10 .xblock--drag-and-drop .zone p {
-    display: table-cell;
-    vertical-align: middle;
-    text-align: center;
+    margin-top: auto;  /* Shield zone labels from LMS rules that change margin-bottom for <p> elements */
+                       /* to make sure they are centered vertically */
+    margin-bottom: auto !important;
```
bradenmacdonald: Note: Unless the rule you're trying to shield this block from is also using `!important`, it's better to override it by increasing the specificity of the selector for this one rule, i.e. using `.xblock--drag-and-drop .drag-container div.target div.zone p`, rather than using `!important`. Otherwise you can end up with a random mix of `!important` rules scattered throughout the CSS and theme CSS overrides, which is harder to maintain. And I think upstream frowns upon `!important` whenever not necessary.

bradenmacdonald:
@itsjeyd You can also remove `public/js/vendor/jquery.html5-placeholder-shim.js` (remove the file and remove the reference from `drag_and_drop_v2.py`.

bradenmacdonald:
@itsjeyd Please make the corresponding selector change to `.themed-xblock.xblock--drag-and-drop .zone p {` in apros.py, otherwise the theme won't be overriding the `p` styles.

Then this is :+1:

PR #52, opened 2016-01-18
Update screenshots

itsjeyd:
... and make sure README covers all options for customizing DnDv2 exercises.

Cf. [SOL-1453](https://openedx.atlassian.net/browse/SOL-1453)


README.md
```diff
@@ -9,8 +9,10 @@ The editor is fully guided. Features include:
 * custom target image
 * free target zone positioning and sizing
 * custom zone labels
+* ability to show or hide zone borders
 * custom text and background colors for items
 * image items
+* items prompting for additional (numerical) input after being dropped
```
bradenmacdonald: I'm actually hoping we can remove this feature in the medium term... since I don't think it will be used much (ever?) and it adds a lot of complexity to the code. So I'd rather not mention it. In my current PR, I've actually hidden the settings for it behind a "Show advanced settings" link. What do you think though?

itsjeyd: @bradenmacdonald Good question, now that I think about it it seems that it would make sense to either remove the "input" feature or generalize it to accept not only numerical but also text input (to make it more flexible). The latter might open it up to another class of use cases. But that's just speculation at this point, I haven't seen any data about how this XBlock is used that would back this up.

As far as this change is concerned, I made it to make the feature list consistent with the description in the "Usage" section, which does include info about the "input" feature. IMHO it would make sense to keep the latter as long as the feature exists, but I'd be fine with removing the feature from the list above.

bradenmacdonald: Ok, fine to leave as is for now.

bradenmacdonald:
:+1: LGTM! Thanks Tim. Cheers :beer:

itsjeyd:
Cheers! :beer: :)

PR #54, opened 2016-02-08
Various Bug Fixes

bradenmacdonald:
This fixes the following bugs:

1. If zone labels are numbers, like "1", "2", etc., then the draggables will not match with that zone (you cannot drop the draggables onto that zone) - this was a consequence of the unfortunate way that the block used zone *titles* as the identifiers of each drag target, and the fact that `$('<x data-x="5" />').data('x') === 5` (jQuery converts data attributes to numbers if they look numeric).

     I solved this by making the block use the new `uid` field to uniquely identify each zone rather than the title. This is something I'd wanted to fix for ages, but hadn't seen a clean way to do so that was fully backwards compatible.
     I think this fix takes a good approach though: Each zone gets a `uid` (string) value and if it doesn't have a `uid`, the `title` is used as the `uid`, which makes it backward compatible.

    This also has the big advantage that when editing the drag and drop block, renaming a zone will no longer reset the associations between each item and its correct zone.

    To test: Confirm that [this example exercise](http://studio.master.dndv2.sandbox.opencraft.com/container/block-v1:OpenCraft+DnDv2+demos+type@vertical+block@f4759960580046f2a92dbefaf4fffc36) on the sandbox is now working and that none ofo teh other demos in that course were broken by this fix.

1. If the platform's locale is set to a language like "de" or "eo" that formats numbers like "3,14", then the "Maximum Score" field appears blank when editing a dndv2 exercise in Studio. Attempting to save the exercise with that field blank causes a 500.

    To test: Go to any course in Studio that has a dndv2 exercise, append `?preview-lang=de` to the URL, then edit the dndv2 block - without this fix, you won't be able to save your changes. (The reason for the bug was that the template produced `value="1,0"` in the HTML which is not a valid value for an HTML5 `number` input - it should always use `value="1.0"` regardless of the locale).

JIRA: [SOL-1413](https://openedx.atlassian.net/browse/SOL-1413)

Sandbox: http://studio.master.dndv2.sandbox.opencraft.com/ and http://master.dndv2.sandbox.opencraft.com/ (shared by a few different open PRs, but it does currently include the changes from this PR)

### DIFF
--- a/drag_and_drop_v2/default_data.py
+++ b/drag_and_drop_v2/default_data.py
@@ -6,6 +6,10 @@ TARGET_IMG_DESCRIPTION = _(
     "and the narrowest layer is located at the top."
 )
 
+TOP_ZONE_ID = "top"
+MIDDLE_ZONE_ID = "middle"
+BOTTOM_ZONE_ID = "bottom"
+
 TOP_ZONE_TITLE = _("The Top Zone")
 MIDDLE_ZONE_TITLE = _("The Middle Zone")
 BOTTOM_ZONE_TITLE = _("The Bottom Zone")
@@ -25,8 +29,7 @@ DEFAULT_DATA = {
   "targetImgDescription": TARGET_IMG_DESCRIPTION,
   "zones": [
     {
-      "index": 1,
-      "id": "zone-1",
+      "uid": TOP_ZONE_ID,
       "title": TOP_ZONE_TITLE,
       "description": TOP_ZONE_DESCRIPTION,
       "x": 160,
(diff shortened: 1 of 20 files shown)


Kelketek:
:+1:

bradenmacdonald:
@dianakhuang Can you please provide a second review for this? (next week is OK if you're not available this week). This (and a second PR I'm about to ping you on) are some follow-up fixes to bugs found after #34 merged.

dianakhuang:
:+1: from me as well.

bradenmacdonald:
Thanks @dianakhuang !

PR #55, opened 2016-02-09
Fix: Background Description was required but not always saved

bradenmacdonald:
This PR fixes the following bugs with the new drag and drop component:

1. "Background description" was marked as required, but if you filled it out and pressed "Continue", "Save", it would accept the data but not actually save. Then next time you went to edit, the description would be blank and you'd get an error saying the field is required. (Fix was to always save the description field - not to only save it when users click on the "Change Background" button)
1. When configuring the draggable items, the "Image Description" was always required, even if the "Image URL" was blank. With this fix, the image description is only required if the Image URL is not blank.
   screen shot 2016-02-09 at 3 50 09 pm

1. When clicking certain action links in the dndv2 editor (e.g. "Add a Zone"), the browser would scroll to the top of the page (since the `href="#"` event was not prevented).
1. When changing tabs in the dndv2 editor, the next tab would often be scrolled down halfway. Fixed so that when you go to the next tab, the editor scrolls to the top of the new tab.
1. In Studio, Newly added drag and drop components do not load properly, due to a Studio bug. I've added a workaround since the fix (https://github.com/edx/edx-platform/pull/11433) is not yet merged and won't be on Cypress or Dogwood.

JIRA: [YONK-264](https://openedx.atlassian.net/browse/YONK-264) / [SOL-1642](https://openedx.atlassian.net/browse/SOL-1642)

Sandbox: http://studio.master.dndv2.sandbox.opencraft.com/ and http://master.dndv2.sandbox.opencraft.com/ (shared by a few different open PRs, but it does currently include the changes from this PR)

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -293,6 +293,8 @@ function DragAndDropBlock(runtime, element, configuration) {
     if (!window.gettext) { window.gettext = function gettext_stub(string) { return string; }; }
 
     var $element = $(element);
+    element = $element[0]; // TODO: This line can be removed when we no longer support Dogwood.
+                           // It works around this Studio bug: https://github.com/edx/edx-platform/pull/11433
     // root: root node managed by the virtual DOM
     var $root = $element.find('.xblock--drag-and-drop');
     var root = $root[0];
--- a/drag_and_drop_v2/public/css/drag_and_drop_edit.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop_edit.css
@@ -188,7 +188,7 @@
 
 .xblock--drag-and-drop--editor .items-form .item-image-url {
     width: 81%;
-    margin-right: 1%;
+    margin: 0 1%;
 }
 
 .xblock--drag-and-drop--editor .items-form .item-width {
@@ -224,7 +224,6 @@
     border: 1px solid #156ab4;
     border-radius: 6px;
     padding: 5px 10px;
-    margin-top: 15px;
 }
 
 .xblock--drag-and-drop--editor .btn:hover {
--- a/drag_and_drop_v2/public/js/drag_and_drop_edit.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop_edit.js
@@ -81,6 +81,10 @@ function DragAndDropEditBlock(runtime, element, params) {
                     return success
                 },
 
+                scrollToTop: function() {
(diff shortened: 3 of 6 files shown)


bradenmacdonald:
@dianakhuang Can you please provide a second review for this PR too? (next week is OK if you're not available this week). This (and the first PR I also just pinged you on) are some follow-up fixes to bugs found after #34 merged.

FWIW, the first two bugs in this PR were introduced during the a11y review on #34, after you'd already given thumbs up.

mtyaka:
Looks good :+1:

bradenmacdonald:
@mtyaka I just pulled in one more one-line workaround, FYI - it was already on the sandbox. Without that fix, newly-added drag and drop components do not load in Studio until you refresh the page. We need this workaround until https://github.com/edx/edx-platform/pull/11433 is reviewed and merged.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -283,6 +283,7 @@ function DragAndDropBlock(runtime, element, configuration) {
     if (!window.gettext) { window.gettext = function gettext_stub(string) { return string; }; }
 
     var $element = $(element);
+    element = $element[0]; // <- workaround Studio bug until https://github.com/edx/edx-platform/pull/11433 is merged
```
dianakhuang: It looks like the edx-platform PR merged, so I don't think you need this workaround anymore.

dianakhuang:
Other than that, :+1:

drag_and_drop_v2/public/js/drag_and_drop.js
bradenmacdonald: @dianakhuang Do you think it's ok if we leave it in for a while though, in order to preserve Dogwood compatibility? We have clients who are using this block on Dogwood.

dianakhuang: I think it's fine to leave it in for compatibility reasons, but in that case, it'd be nice to have the comment updated to reflect exactly what needs to happen before it can be removed. I might be helpful to write the new comment as a `// TODO: ` as well to make it easier to find later and let people know that it's an action that should be taken later.

bradenmacdonald: Great, makes sense. I will update the comment accordingly before merging this.

bradenmacdonald:
Thanks for the review, @dianakhuang ! I've updated that comment and rebased. Will merge once the build is green.

PR #53, opened 2016-02-02
OC-1320 DnDv2 - Allow to set automatic alignment of dragged items in drop zones

pomegranited:
This PR adds an "align" field that it editable for each zone, with the options "none", "left", "right", and "center".  Align "none" corresponds to in the current default behaviour, where items are placed where the learner puts them via mouse, or centered in the zone via keyboard.

* Align "left" causes placed items to be stacked left to right in the zone, wrapping at the right edge of the zone.
* Align "right" works the same way as "left", but stacks from right to left.
* Align "center" causes placed items to be aligned in the center of the zone, but they are stacked top to bottom.

Items in "aligned" zones will not overlap each other, but if there's not enough space in the zone for all the items, they will overflow over the bottom of the zone, potentially overlapping other zones.

**JIRA tickets**: Implements OC-1320

**Discussions**:

1. The original specification requested that items placed in "aligned" zones be 
sequenced according the order they were placed by the learner.  This proved 
complex to implement, and so instead, items are sequenced in the order they are
configured by the instructor.

**Dependencies**: None

**Screenshots**:

Studio:
Updated DnDv2 Edit Zone screen:
edit zone align

LMS:
Example with multiple items placed in zones with mixed alignment:
mixed-alignment

Example with multiple items per zone, mixed alignment, with overlapping zones:
overlapping zones

Example with mixed alignment and numerical input:
numerical input

**Sandbox URLs**:

Studio: http://studio.master.dndv2.sandbox.opencraft.com/course/course-v1:OpenCraft+DnDv2+review
LMS: http://master.dndv2.sandbox.opencraft.com/courses/course-v1:OpenCraft+DnDv2+review/courseware/6fa41195e31b430398e755ca3bc61808/879aff4aa8c14ad385d896addcbd3d09/

**Partner information**: StGallen

**Merge deadline**: Thu Feb 4 EOD.  Apologies for the short notice!

**Testing instructions**:

1. Go to the course "Introduction to Drag and Drop v2" in Studio
2. In each Subsection, click Edit for the Drag and Drop exercise to see how the alignment is set up for each zone.

1. Go to the course "Introduction to Drag and Drop v2" in the LMS
2. In each Subsection, drag items to each zone to see how they are aligned.

*Expected behaviour*
* Alignment: none -- dropped items should remain where they're dropped (mouse), or be centered in the zone (keyboard)
* Alignment: left -- dropped items should be aligned with the left side of the zone.
* Alignment: center -- dropped items should be aligned in the center of the zone.
[...]

PR #57, opened 2016-02-11
Allow to set automatic alignment of dragged items in drop zones (second attempt)

smarnach:
This is a new version of #53, reopened for review after the original PR was merged and reverted.  See the other PR for all relevant information.  Please add any new comments to this PR, unless you are replying directly to a discussion on the old PR.

JIRA ticket: [OSPR-1139](https://openedx.atlassian.net/browse/OSPR-1139)
Sandbox: [Studio](http://studio.master.dndv2.sandbox.opencraft.com/) [LMS](http://master.dndv2.sandbox.opencraft.com/) (Updated 2016-03-30 to f7c3aeddc7e535c259c085d62179cf8a4b784251)


smarnach:
There are still some flaky tests in `test_sizing`; these should be fixed by #55.

bradenmacdonald:
@smarnach FYI, this will need to be rebased/updated now that #54 has merged. The `data-zone` attribute has changed to `data-uid`. The required updates to make the new tests in this PR pass again can be seen in 436d46f86ade8ca1328547e4c00455741ba3f649

Kelketek:
@bradenmacdonald This has been rebased, and tests pass locally. Should end up a green build.

bradenmacdonald:
@Kelketek Great! Did you fix that bug I reported as well?

Kelketek:
@bradenmacdonald No. I still need to do that.

Kelketek:
@itsjeyd This PR has been rebased and updated. Could you take a look at the latest iteration and its changes? The main thing I added, aside from rebasing, was a compatibility fix in studio for Firefox.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -467,6 +479,7 @@ function DragAndDropEditBlock(runtime, element, params) {
                                         correct: $el.find('.success-feedback').val(),
                                         incorrect: $el.find('.error-feedback').val()
                                     },
+
```
itsjeyd: @Kelketek Nit: There's no need for adding a blank line here I think.

Kelketek:
@itsjeyd Could you take another look now?

itsjeyd:
@Kelketek :+1:

Kelketek:
@bradenmacdonald Pinging you on this, since @itsjeyd gave me the thumbs up, and you asked me to do so. Do I merge or do we ping someone else, or?

bradenmacdonald:
@Kelketek This will go through the OSPR process. An OSPR ticket is already open for this PR - see description at the top. So we just wait for edx to assign a reviewer and approve it. In the meantime, please add a sandbox link (master.dndv2.sandbox.opencraft.com if it's been updated) and test instructions to the PR description.

I can update this on the client's instance in the meantime, since we can't wait for the upstream review to finish.

bradenmacdonald:
@Kelketek @itsjeyd There is a problem with this - I can't deploy it as-is:

If an item has a fixed width, its size is suddenly dependent on whether it's in an aligned zone or not.

In the screenshots below, the items are all configured to have a width of "50%" - but when zone alignment is turned on, they shrink when placed:
<table><tr><td>Non-aligned:<br></td><td>Aligned:<br></td></tr></table>

And even if a fixed width is not specified, if the zone is small, the default rule of "max-width: 33%" causes ridiculous layouts:
ridiculous

The main problem here is that if zone alignment is turned on, the placed blocks' DOM parent is now the zone div instead of the whole image container, so "width" and "max-width" are now relative to a (potentially) smaller parent.

You need to change [the sizing logic](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/1cd770d/drag_and_drop_v2/public/js/drag_and_drop.js#L92-L107) so that:
* if an item has a fixed width specified and is dropped onto an aligned zone, the dropped item's DIV gets assigned a percentage width that is adjusted proportionally to the width of the zone. (e.g. if an item is configured to have "33%" width relative to the background image and is dropped onto a zone that is 50% as wide as the image, the item should get an adjusted CSS width and max-width of "66%" to compensate for the fact that it's parent DIV is suddenly half as wide.
* If an item does not have a fixed width but is dropped onto an aligned zone, the default CSS "max-width: 33%" default rule must be overruled by a style attribute that sets an adjusted max-width that accounts for the potentially smaller width of the parent zone.

bradenmacdonald:
The bugs mentioned above are fixed, with new tests. @itsjeyd can you please review the last couple of commits, since they were added after your review?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -129,7 +157,7 @@ function DragNDropTemplates(url_name) {
             var item_description = h(
                 'div',
                 { id: item_description_id, className: 'sr' },
-                gettext('Correctly placed in: ') + getZoneTitle(item.zone, ctx)
+                gettext('Correctly placed in: ') + (zone.title || "Unknown Zone")
```
itsjeyd: @bradenmacdonald Could you use `getZoneTitle` here? If not, you might want to remove that method completely (it's not used anywhere).

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -89,12 +99,30 @@ function DragNDropTemplates(url_name) {
...
+            if (item.zone_align === 'none') {
+                style.left = item.x_percent + "%";
+                style.top = item.y_percent + "%";
+            } else {
+                // Allow for the input + button width for aligned items
+                if (item.input) {
+                    style.marginRight = '190px';
+                }
+            }
+            if (item.zone_align !== 'none') {
```
itsjeyd: @bradenmacdonald I don't think we need a separate `if` block here. Could you add the statements belonging to its body to the `else` block above?

itsjeyd:
@bradenmacdonald :+1: once comments are addressed. I also tested again for the use cases you described [above](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/57#issuecomment-188000389), it's working well.

drag_and_drop_v2/public/js/drag_and_drop.js
bradenmacdonald: @itsjeyd I'll just remove it, since it's more complex than needed here, where we already have the `zone` object available. Thanks for pointing that out.

drag_and_drop_v2/public/js/drag_and_drop.js
bradenmacdonald: @itsjeyd Good point. In earlier iterations there wasn't a clean way to combine these, but with the way I separated out the image-size handling at the end, these should now be combined.

bradenmacdonald:
@itsjeyd Thanks for the second look. I pushed a commit to fix your nits.

douglashall:
@bradenmacdonald I'm not seeing the new alignment configuration option on the sandbox [here](http://studio.master.dndv2.sandbox.opencraft.com/container/block-v1:OpenCraft+DnDv2+review+type@vertical+block@c98ca6ddc80645b1ab91ee876d5d7d1e#). Does this sandbox need to be updated to include these changes?

bradenmacdonald:
@douglashall Sorry, we've be re-using that sandbox for various PRs and this one has been on the shelf for a while. I've updated the sandbox now to use the version from this branch.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -52,19 +52,19 @@ function DragNDropTemplates(url_name) {
...
+    var getZone = function(zoneUID, ctx) {
         for (var i = 0; i < ctx.zones.length; i++) {
             if (ctx.zones[i].uid === zoneUID) {
-                return ctx.zones[i].title;
+                return ctx.zones[i];
             }
         }
-        return "Unknown Zone";  // This title should never be seen, so does not need i18n
-    }
+
```
douglashall: Nit: There is an extra blank line here.

douglashall:
:+1: One nit. Please squash commits before merging.

bradenmacdonald:
Thanks @douglashall! I've addressed your nit and then squashed and rebased 86d08a3dce76801a3037afc2824e07d9a6e34008 onto master. I also added this to the Changelog.

PR #58, opened 2016-02-15
Version bump to 2.0.1

bradenmacdonald:
Version bump so we can get the latest master version released with a tagged hash update.


Kelketek:
:+1:

PR #59, opened 2016-02-17
mattdrayer/xblock-translations: Initial attempt

mattdrayer:
This is a complementary changeset to go along with edx/edx-platform#11575

@asadiqbal08 @nedbat @cpennington @bradenmacdonald FYI


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -160,16 +173,16 @@ def items_without_answers():
             "url_name": getattr(self, 'url_name', ''),
             "display_zone_labels": self.data.get('displayLabels', False),
             "display_zone_borders": self.data.get('displayBorders', False),
-            "items": items_without_answers(),
-            "title": self.display_name,
+            "items": items_without_answers(self),
+            "title": self._(self.display_name),
```
cpennington: Do you really want to run this through i18n? This is dynamic, right? So if the instructor changes it, those strings aren't going to make it into the translation dictionary.

cpennington: This is an instance of a translatable (Text?) field, that I would want to get translated by the XBlock runtime (with translations being provided by the author)

bradenmacdonald: Translating user-provided content doesn't just have the problem of missing translations; it can also cause problems. e.g. French instructor writes "Location" (which means "Rental" in French), but after getting translated by this code, it would appear in the course as "Emplacement" or something, which has a completely different meaning.

nedbat: BTW, if you hook this repo up to edx-lint, then pylint will point these out.

bradenmacdonald: It is hooked up to pylint ([e.g.](https://travis-ci.org/edx-solutions/xblock-drag-and-drop-v2/builds/109975828)) but it doesn't seem to have caught those. pylint version / pylintrc problem perhaps?

BTW some of the tests are flaky but we have two open PRs that fix the flaky tests.

nedbat: edx-lint provides some plugins to pylint, as well as standardized pylintrc file so that we won't have wobbly standards across repos.  One of the plugins is to detect translations of non-literal strings.

mattdrayer: We can remove/change everything in this PR -- really I was more interested in demonstrating the i18n workflow itself, so whatever makes sense to tag/translate is fine with me.  We have a list of approximately two dozen XBlocks that we're going to need to configure for I18N, and I'm sure each one is going to have its own pecularities.

asadiqbal08:
@mattdrayer I had moved the `/conf/...` directory from 'xblock-drag-and-drop-v2' to inside xBlock app `/drag_and_drop_v2/` . You need to compile the .po files in order to generate .mo(s)

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -107,8 +109,14 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     has_score = True
 
     def _(self, text):
-        """ Translate text """
-        return self.runtime.service(self, "i18n").ugettext(text)
```
bradenmacdonald: ^ Am I right that the older, simpler code seen here is actually all that's going to be needed, in light of the latest changes to https://github.com/edx/edx-platform/pull/11575 ?

PR #60, opened 2016-02-18
Version bump to 2.0.2, added Changelog

bradenmacdonald:
This is a version bump so I can tag a new release and get the important fixes from #55 deployed to edx.org.

I also added a changelog and fixed some flaky tests observed while opening this PR.

### DIFF
--- a/tests/integration/test_interaction.py
+++ b/tests/integration/test_interaction.py
@@ -122,7 +122,12 @@ class InteractionTestBase(object):
     def send_input(self, item_value, value):
         element = self._get_item_by_value(item_value)
         self.wait_until_visible(element)
-        element.find_element_by_class_name('input').send_keys(value)
+        # Since virtual-dom may be updating DOM elements by replacing them completely, the
+        # following method must be used to wait for the input box to appear:
+        textbox_visible_selector = '.numerical-input[style*="display: block"] input'
+        self.wait_until_exists(textbox_visible_selector)
+        textbox = self._page.find_element_by_css_selector(textbox_visible_selector)
+        textbox.send_keys(value)
         element.find_element_by_class_name('submit-input').click()
 
     def assert_grabbed_item(self, item):
--- a/tests/integration/test_sizing.py
+++ b/tests/integration/test_sizing.py
@@ -90,7 +90,7 @@ class SizingTests(InteractionTestBase, BaseIntegrationTest):
         # A 400x300 image with automatic sizing should be constrained to the maximum width
         Expectation(item_id=5, zone_id=ZONE_50, width_percent=AUTO_MAX_WIDTH),
         # A 200x200 image with automatic sizing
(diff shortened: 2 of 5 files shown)


tests/integration/test_interaction.py
```diff
@@ -122,7 +122,12 @@ def move_item_to_zone(self, item_value, zone_id, action_key):
     def send_input(self, item_value, value):
         element = self._get_item_by_value(item_value)
         self.wait_until_visible(element)
-        element.find_element_by_class_name('input').send_keys(value)
+        # Since virtual-dom may be updating DOM elements by replacing them completely, the
+        # following method must be used to wait for the input box to appear:
+        textbox_visible_selector = '.numerical-input[style*="display: block"] input'
```
mtyaka: @bradenmacdonald Would it be possible to use [`is_displayed()`](http://selenium-python.readthedocs.org/api.html#selenium.webdriver.remote.webelement.WebElement.is_displayed) instead of the style selector?

tests/integration/test_sizing.py
```diff
@@ -129,19 +140,20 @@ def _check_width(self, item_description, item, container_width, expected_percent
...
         that range.
         """
-        width_percent = item.size["width"] / container_width * 100
+        width_pixels = item.size["width"]
+        width_percent = width_pixels / container_width * 100
         if isinstance(expected_percent, (list, tuple)):
             min_expected, max_expected = expected_percent
-            msg = "{} should have width of {}% - {}%. Actual: {:.2f}%".format(
-                item_description, min_expected, max_expected, width_percent
+            msg = "{} should have width of {}% - {}%. Actual: Actual: {}px ({:.2f}% of {}px)".format(
```
mtyaka: Doubled "Actual:"

mtyaka:
Added a couple of nits, but it looks good :+1:

tests/integration/test_interaction.py
bradenmacdonald: @mtyaka I don't think so? I was seeing flaky tests and I think one reason is that if we use this pattern (note `wait_until_visible` uses `is_displayed`):
```python
element = find_element_by_class_name('blah')
self.wait_until_visible(element)
```
Then sometimes between the first line and the second line, virtual_dom deletes the element and replaces it with a new one, which then causes this Selenium error: "The element is no longer attached to the DOM."

tests/integration/test_sizing.py
bradenmacdonald: Thanks. Fixed.

bradenmacdonald:
@dianakhuang Could you please review this small dndv2 PR as well? It's mostly a version bump so I can tag a new version to update in the edx-platform requirements file and include the fixes you reviewed in #54. I also fixed some flaky tests. (There are no changes to the XBlock code itself in this PR.)

dianakhuang:
:+1:

PR #62, opened 2016-02-29
Make Safari treat text areas with errors properly.

Kelketek:
@itsjeyd For your review.

Resolves https://openedx.atlassian.net/browse/SOL-1660

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop_edit.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop_edit.css
@@ -271,7 +271,9 @@
 
 .xblock--drag-and-drop--editor .tab .field-error {
     outline: none;
-    box-shadow: 0 0 10px darkred;
+    box-shadow: 0 0 10px 0 darkred;
+    /* Needed because Safari won't show the box-shadow for textareas otherwise. */
+    -webkit-appearance: none;
 }
 
 .xblock--drag-and-drop--editor .icon.add:before {


itsjeyd:
@Kelketek Tested on crossbrowsertesting.com using different versions of Safari, and double-checked locally with latest versions of Chrome, Chromium, Firefox. Your changes fix the bug for me. :+1: once you get a green build.

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -271,7 +271,8 @@
 
 .xblock--drag-and-drop--editor .tab .field-error {
     outline: none;
-    box-shadow: 0 0 10px darkred;
+    box-shadow: 0 0 10px 0 darkred;
+    -webkit-appearance: none;
```
bradenmacdonald: Nit: Please add a comment explaining why this style is here... Otherwise someone might remove it in a future refactor.

Kelketek:
@bradenmacdonald Added in the comment you requested. When tests pass and I merge, I'll need to update the hash in edx/edx-platform, right?

bradenmacdonald:
@Kelketek You need two thumbs - I can give the second thumb here since this is a simple fix (but note I haven't tested this yet so wait for my +1). Hold off on hash update since we have other bugfix PRs in flight, I believe.

bradenmacdonald:
:+1: when build is green.
Works for me. Code change is nice and simple. Tested in Safari and Chrome.

PR #63, opened 2016-03-04
Expand static image URLs for item background images.

smarnach:
Handle /static/… URLs correctly for item background images, since this is the only portable URL format.

We need to include both the expanded and the unexpanded URLs when passing items to the JS code, since we need the expanded URL for rendering and the original URL for editing in Studio.

**JIRA ticket**: [SOL-1661](https://openedx.atlassian.net/browse/SOL-1661)

**Sandbox**: [LMS](http://master.dndv2.sandbox.opencraft.com/courses/course-v1:TestX+TST-DNDV2-M+now/courseware/1e4f0db6718644ab891d3f57074e0090/042178f28a5a4a0f9a1b7aa93834e59b/) [Studio](http://studio.master.dndv2.sandbox.opencraft.com/container/block-v1:TestX+TST-DNDV2-M+now+type@vertical+block@1e63181ae0c74ef4beae234aa38f7798)

**Testing instructions**: Either use the above sandobx links to a DnDv2 instance configured with background images, or start over with a new unit:

1. Go to Content -> Files & Uploads on your test course in Studio and upload some images.  Keep the tab open to be able to copy the static URLs.
2. In a new browser tab, add a new DnDv2 XBlock to a unit, click "Edit", click "Continue" twice, and enter some image URLs from the previous step (the image descriptions are required as well).
3. Check that the background images display correctly in Studio and the LMS.
4. Change the URLs and check again.

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -857,7 +857,6 @@ function DragAndDropBlock(runtime, element, configuration) {
                     input.class_name = item_user_state.correct_input ? 'correct' : 'incorrect';
                 }
             }
-            var imageURL = item.imageURL || item.backgroundImage;  // Fall back on "backgroundImage" to be backward-compatible
             var grabbed = false;
             if (item.grabbed !== undefined) {
                 grabbed = item.grabbed;
@@ -870,9 +869,9 @@ function DragAndDropBlock(runtime, element, configuration) {
                 xhr_active: (item_user_state && item_user_state.submitting_location),
                 input: input,
                 displayName: item.displayName,
-                imageURL: imageURL,
+                imageURL: item.expandedImageURL,
                 imageDescription: item.imageDescription,
-                has_image: !!imageURL,
+                has_image: !!item.expandedImageURL,
                 grabbed: grabbed,
                 widthPercent: item.widthPercent, // widthPercent may be undefined (auto width)
                 imgNaturalWidth: item.imgNaturalWidth,
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -152,6 +152,12 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
                 del item['feedback']
                 del item['zone']
                 item['inputOptions'] = 'inputOptions' in item
+                # Fall back on "backgroundImage" to be backward-compatible.
+                image_url = item.get('imageURL') or item.get('backgroundImage')
+                if image_url:
+                    item['expandedImageURL'] = self._expand_static_url(image_url)
+                else:
+                    item['expandedImageURL'] = ''
             return items
 
         return {
--- a/tests/unit/data/html/config_out.json
+++ b/tests/unit/data/html/config_out.json
@@ -35,24 +35,28 @@
         {
           "displayName": "<b>1</b>",
           "imageURL": "",
+          "expandedImageURL": "",
           "id": 0,
           "inputOptions": false
         },
         {
           "displayName": "<i>2</i>",
           "imageURL": "",
+          "expandedImageURL": "",
           "id": 1,
           "inputOptions": true
         },
         {
           "displayName": "X",
           "imageURL": "",
+          "expandedImageURL": "",
           "id": 2,
           "inputOptions": false
         },
         {
           "displayName": "",
           "imageURL": "http://placehold.it/100x300",
+          "expandedImageURL": "http://placehold.it/100x300",
           "id": 3,
           "inputOptions": false
         }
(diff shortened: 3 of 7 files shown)


drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -871,6 +871,7 @@ function DragAndDropBlock(runtime, element, configuration) {
                 input: input,
                 displayName: item.displayName,
                 imageURL: imageURL,
```
bradenmacdonald: Do we need to keep both `imageURL` and `expandedImageURL` as separate attributes of `itemProperties` in this view? Seems like a potential source of bugs. You could probably get away with changing line 860 to

```
var imageURL = item.expandedImageURL || item.backgroundImage;  // Fall back on "backgroundImage" to be backward-compatible
```
and then just using `imageURL` everywhere from that point on.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -456,9 +456,22 @@ function DragAndDropEditBlock(runtime, element, params) {
...
+                                if (imageURL) {
+                                    var handlerUrl = runtime.handlerUrl(element, 'expand_static_url');
+                                    $.ajax({
+                                        type: 'POST',
+                                        url: handlerUrl,
+                                        data: JSON.stringify(imageURL),
+                                        success: function (result) {
+                                            expandedImageURL = result.url;
+                                        },
+                                        async: false
```
bradenmacdonald: I really don't like this synchronous call... If there were ten draggables with images this would be a very bad user experience, since there would be ten blocking network calls in sequence.

Can you not just submit the `imageURL` here? Why do you need to expand the URL at this point?  (Since there's no preview being shown, and `expandedImageURL` is a derived property that doesn't need to be stored, and the Python `submit` handler could compute the `expandedImageURL` instantly if it were somehow necessary).

drag_and_drop_v2/public/js/drag_and_drop.js
smarnach: I found it confusing to use the property name `imageURL` for different things in different places, but I'm happy to change it.  The actual image URL is indeed not needed for the view; I just left it there for completeness.  I don't think this would cause any further bugs, since neither of the URLs are ever changed on the view, and the image URL is only used in a single place.

This comment made me realise that I need to move the fallback to the old attribute name to the Python code (since the URL gets expanded there).

drag_and_drop_v2/public/js/drag_and_drop_edit.js
smarnach: @bradenmacdonald You are completely right, we don't need this here.  I didn't read the code correctly, and somehow assumed we need this because the XBlock gets updated when it is edited (but in fact it is completely reloaded), and in case of the target image we only need this for the preview, so I removed it.  (I didn't like the synchronous call either, but I'm glad I didn't spend much time figuring out a way to avoid it.)

Kelketek:
@smarnach You have test failures.

Kelketek:
@smarnach The task requirements say a test is needed. You may need to mock some things out but this should be doable.

smarnach:
@Kelketek The test failures look like some tests are relying on the order of dictionaries.  They are probably unrelated, but I'll look into them.

Regarding the new test, we already patch `replace_urls` in the workbench runtime.  In my opinion it would be enough to add a simple unit test checking that a correctly expanded URL is included when retrieving the configuration (which I will do now).

drag_and_drop_v2/public/js/drag_and_drop.js
bradenmacdonald: Up to you. You can also just use `expandedImageURL` throughout the JS code. My main point was we didn't need two similar variables, when only one is actually used. But this is not something I feel strongly about.

Fallback for the old attribute name is not really important, because the old block was never properly expanding these URLs anyways, so the old attribute will never contain short URLs. It's fine to expand it just in case though.

smarnach:
@Kelketek The test failures turned out to be related.  For the latest version, the tests are passing locally.  I also included an example of a static URL in the data to test it gets correctly expanded.  Could you please take another look?

drag_and_drop_v2/public/js/drag_and_drop.js
smarnach: I don't feel strongly about the variable names either, so I'll just leave it the way the current version happens to be.

The problem with the fallback was that `expandedImageURL` would be undefined if the old attribute name was used, but the JS code used it anyway in that case, so it was indeed a bug (fixed now).

Kelketek:
:+1:

mattdrayer:
@smarnach I followed the instructions noted above on the sandbox and everything appears to be in order -- code looks good and thanks much for the test coverage -- :+1:

smarnach:
@mattdrayer Thanks for the quick review!

PR #66, opened 2016-03-25
WL-373 i18n support drag and drop v2

asadiqbal08:
@mattdrayer @bradenmacdonald Can you please review the change set

Arabic Content Demo: 
screen shot 2016-03-25 at 4 20 48 pm

##### We can ignore test failure at that time, Having dependancy on [#335](https://github.com/edx/XBlock/pull/335)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
-Development version:
---------------------
+Version 2.0.5 (2016-03-30)
+--------------------------
 
+* Marked strings for translations (PR #66)
 * Added the ability to specify automatic alignment of dropped items. (PR #57)
 
 Version 2.0.4 (2016-03-10)
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -27,6 +27,7 @@ loader = ResourceLoader(__name__)
 # Classes ###########################################################
 
 @XBlock.wants('settings')
+@XBlock.needs('i18n')
 class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     """
     XBlock that implements a friendly Drag-and-Drop problem
@@ -106,10 +107,6 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     block_settings_key = 'drag-and-drop-v2'
     has_score = True
 
-    def _(self, text):
-        """ Translate text """
-        return self.runtime.service(self, "i18n").ugettext(text)
-
     @XBlock.supports("multi_device")  # Enable this block for use in the mobile app via webview
     def student_view(self, context):
         """
@@ -186,7 +183,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
         js_templates = loader.load_unicode('/templates/html/js_templates.html')
         help_texts = {
-            field_name: self._(field.help)
+            field_name: self.ugettext(field.help)
             for field_name, field in self.fields.viewitems() if hasattr(field, "help")
         }
         context = {
--- a/drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
+++ b/drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
@@ -1,0 +1,452 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: \n"
+"POT-Creation-Date: 2016-03-30 16:54+0500\n"
+"PO-Revision-Date: 2016-03-30 16:54+0500\n"
+"Last-Translator: \n"
+"Language-Team: \n"
+"Language: en\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#: default_data.py
+msgid ""
+"An isosceles triangle with three layers of similar height. It is shown "
+"upright, so the widest layer is located at the bottom, and the narrowest "
+"layer is located at the top."
+msgstr ""
+
+#: default_data.py
+msgid "The Top Zone"
+msgstr ""
+
+#: default_data.py
+msgid "The Middle Zone"
+msgstr ""
+
+#: default_data.py
+msgid "The Bottom Zone"
+msgstr ""
+
+#: default_data.py
+msgid "Use this zone to associate an item with the top layer of the triangle."
+msgstr ""
+
+#: default_data.py
+msgid ""
+"Use this zone to associate an item with the middle layer of the triangle."
+msgstr ""
+
+#: default_data.py
+msgid ""
+"Use this zone to associate an item with the bottom layer of the triangle."
+msgstr ""
+
+#: default_data.py
+#, python-brace-format
+msgid "Correct! This one belongs to {zone}."
+msgstr ""
+
+#: default_data.py
+msgid "No, this item does not belong here. Try again."
+msgstr ""
+
+#: default_data.py
+msgid "You silly, there are no zones for this one."
+msgstr ""
+
+#: default_data.py
+msgid "Drag the items onto the image above."
+msgstr ""
+
+#: default_data.py
+msgid "Good work! You have completed this drag and drop problem."
+msgstr ""
+
+#: default_data.py
+msgid "Goes to the top"
+msgstr ""
+
+#: default_data.py
+msgid "Goes to the middle"
+msgstr ""
+
+#: default_data.py
+msgid "Goes to the bottom"
+msgstr ""
+
+#: default_data.py
+msgid "I don't belong anywhere"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Title"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid ""
+"The title of the drag and drop problem. The title is displayed to learners."
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Drag and Drop"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Show title"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Display the title to the learner?"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Problem text"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "The description of the problem or instructions shown to the learner"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Show \"Problem\" heading"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Display the heading \"Problem\" above the problem text?"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Weight"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "The maximum score the learner can receive for the problem"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Item background color"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "The background color of draggable items in the problem."
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Item text color"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Text color to use for draggable items."
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Problem data"
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid ""
+"Information about zones, items, feedback, and background image for this "
+"problem. This information is derived from the input that a course author "
+"provides via the interactive editor when configuring the problem."
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid ""
+"Information about current positions of items that a learner has dropped on "
+"the target image."
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Indicates whether a learner has completed the problem at least once"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Text"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Description"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Describe this zone to non-visual users"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "width"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "height"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Alignment"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "none"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "left"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "center"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "right"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid ""
+"Align dropped items to the left, center, or right.  Default is no alignment "
+"(items stay exactly where the user drops them)."
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Zone"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Image URL (alternative to the text)"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "For example, http://example.com/image.png or /static/image.png"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid ""
+"Image description (should provide sufficient information to place the item "
+"even if the image did not load)"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Success Feedback"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Error Feedback"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid "Show advanced settings"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid ""
+"Preferred width as a percentage of the background image width (or blank for "
+"automatic width):"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid ""
+"Optional numerical value (if you set this, learners will be prompted for "
+"this value after dropping this item)"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid ""
+"Margin ± (when a numerical value is required, values entered by learners "
+"must not differ from the expected value by more than this margin; default "
+"is zero)"
+msgstr ""
+
+#: templates/html/drag_and_drop.html
+msgid "Loading drag and drop problem."
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid ""
+"Note: do not edit the problem if students have already completed it. Delete "
+"the problem and create a new one."
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Problem title"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Show title"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Maximum score"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Problem text"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Show \"Problem\" heading"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Introductory Feedback"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Final Feedback"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Zones"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Background URL"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid ""
+"For example, http://example.com/background.png or /static/background.png"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Change background"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Background description"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid ""
+"\n"
+"                            Please provide a description of the image for "
+"non-visual users.\n"
+"                            The description should provide sufficient "
+"information to allow anyone\n"
+"                            to solve the problem even without seeing the "
+"image.\n"
+"                        "
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Zone labels"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Display label names on the image"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Zone borders"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Display zone borders on the image"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Add a zone"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Items"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Background color"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Text color"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Add an item"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Continue"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Save"
+msgstr ""
+
+#: templates/html/drag_and_drop_edit.html
+msgid "Cancel"
+msgstr ""
+
+#: public/js/drag_and_drop.js
+msgid "ok"
+msgstr ""
+
+#: public/js/drag_and_drop.js
+msgid "Correctly placed in: "
+msgstr ""
+
+#: public/js/drag_and_drop.js
+msgid "Reset problem"
+msgstr ""
+
+#: public/js/drag_and_drop.js
+msgid "Feedback"
+msgstr ""
+
+#: public/js/drag_and_drop.js
+msgid "Keyboard Help"
+msgstr ""
+
+#: public/js/drag_and_drop.js
+msgid "You can complete this problem using only your keyboard."
+msgstr ""
+
+#: public/js/drag_and_drop.js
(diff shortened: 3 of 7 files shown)


mattdrayer:
@bradenmacdonald -- FYI, after some code dependency updating and juggling, we have a green build for this PR.  If you could take a look that would fantastic.

bradenmacdonald:
@mattdrayer Great! @itsjeyd will be taking a look at this on my behalf :)

mattdrayer:
Excellent -- @itsjeyd, let me know if you have any questions, thanks!

setup.py
```diff
@@ -35,5 +35,5 @@ def package_data(pkg, root_list):
     entry_points={
         'xblock.v1': 'drag-and-drop-v2 = drag_and_drop_v2:DragAndDropBlock',
     },
-    package_data=package_data("drag_and_drop_v2", ["static", "templates", "public"]),
+    package_data=package_data("drag_and_drop_v2", ["static", "templates", "public", "translations"]),
```
asadiqbal08: @mattdrayer is it required to update the version in setup.py for this xblock  ? currently it is pointing to `2.0.4`

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -167,15 +165,15 @@ def items_without_answers():
...
             "problem_text": self.question_text,
             "show_problem_header": self.show_question_header,
             "target_img_expanded_url": self.target_img_expanded_url,
-            "target_img_description": self.target_img_description,
-            "item_background_color": self.item_background_color or None,
-            "item_text_color": self.item_text_color or None,
-            "initial_feedback": self.data['feedback']['start'],
+            "target_img_description": self.ugettext(self.target_img_description),
+            "item_background_color": self.ugettext(self.item_background_color) or None,
+            "item_text_color": self.ugettext(self.item_text_color) or None,
```
itsjeyd: @mattdrayer `item_background_color` and `item_text_color` should not be translated: They are used to set CSS properties of DOM elements representing items, so translating them would break item styles.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -301,7 +299,7 @@ def do_attempt(self, attempt, suffix=''):
 
         self.runtime.publish(self, 'edx.drag_and_drop_v2.item.dropped', {
             'item_id': item['id'],
-            'location': zone.get("title"),
+            'location': self.ugettext(zone.get("title")),
```
itsjeyd: @mattdrayer If this is translated, is it possible that the event logs could end up containing entries that refer to the same zone but contain different values for `location` (e.g., "first zone"/"primera zona"/"erste Zone")? If so, it might be better to default to the untranslated value here.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -151,6 +148,7 @@ def items_without_answers():
             for item in items:
                 del item['feedback']
                 del item['zone']
+                item['displayName'] = self.ugettext(item['displayName'])
```
itsjeyd: @mattdrayer `displayName` may contain HTML markup (cf. [tests](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/master/tests/integration/data/test_html_data.json#L24)) which would end up in the translation files (and would need to be recreated in individual translations). The same goes for zone titles and initial and final feedback. I'm not sure if this is a problem; just pointing it out for now in case it's relevant.

itsjeyd:
@mattdrayer I made a first pass over the code and left a few comments.

I still need to do some manual testing (so far I've verified that the changes don't break existing/untranslated instances of DnDv2). If you could add instructions for that to the PR description (and outline the intended workflow for creating translations for DnDv2 problems), that would be great. You can assume that:

* edx-platform is at a version that includes the changes from https://github.com/edx/edx-platform/pull/11575
* XBlock is at version 0.4.7 (calls to `self.ugettext` from `DragAndDropBlock` don't error out). Once this PR merges, are you going to do a hash update for edx-platform to ensure that this dependency is satisfied automatically?

drag_and_drop_v2/drag_and_drop_v2.py
asadiqbal08: @itsjeyd If we have translations msgstr in .po files with for HTML markup then `self.ugettext` will perform the translation for those, otherwise it will be render as it is. 
Currently, I was seeing that, item displayName has the following values:

- "Goes to the top"
- "Goes to the middle"
- "Goes to the bottom"
- "I don't belong anywhere"

correct me If I am missing . 
Thanks

asadiqbal08: We have some initial default values in default_data.py. I think We can translate these initial static data.

drag_and_drop_v2/drag_and_drop_v2.py
bradenmacdonald: I would add: as a general rule, any time you are wrapping `ugettext` around something other than a constant string is probably not correct. AFAIK, the only time it is correct is when it's translating a value that could not possibly be translated earlier and where all possible values of that variable are guaranteed to have been previously marked with `ugettext_noop()`. A lot of the uses below look wrong to me, since they are wrapping user-provided strings.

Example from my [previous comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/59#discussion_r53235814): 'Translating user-provided content doesn't just have the problem of missing translations; it can also cause problems. e.g. French instructor writes "Location" (which means "Rental" in French), but after getting translated by this code, it would appear in the course as "Emplacement" or something, which has a completely different meaning.'

setup.py
bradenmacdonald: In order to update the version used in edx-platform, you'll have to do a version bump yes - see previous version bump PRs, because you have to update a few separate things - I think setup.py, circle.yml, and Changelog.md

drag_and_drop_v2/drag_and_drop_v2.py
itsjeyd: @asadiqbal08 

> We have some initial default values in default_data.py. I think We can translate these initial static data. [...]

Since these defaults live in a separate module (i.e., they are not part of the `DragAndDropBlock` class), I don't think you can translate them using the new `ugettext` mechanism from `XBlock`. Also, zone titles can (and will :)) be modified by course authors; as soon as that happens, translations for default values won't apply anymore (even if an author only uses the default zones that are created when adding a new DnDv2 block to a unit, and doesn't add any additional zones).

asadiqbal08: @itsjeyd can you please highlight the places where `ugettext` is not required anymore? I am going to update this PR with version as well.

asadiqbal08:
###### Note: New release tag 2.0.5 will be required after this merge.

drag_and_drop_v2/drag_and_drop_v2.py
itsjeyd: @asadiqbal08 See @bradenmacdonald's [comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/66#discussion_r57737584) -- you shouldn't call `self.ugettext` on user-provided strings. These include the string in this line (`item[displayName]`) and most of the strings below; calling it on `field.help` is fine since help texts for fields are defined in the code ([example](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/asadiqbal08/WL-373/drag_and_drop_v2/drag_and_drop_v2.py#L37)) and can't be changed by course authors.

asadiqbal08:
@itsjeyd kindly take a look

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
```diff
@@ -0,0 +1,168 @@
...
+
+#: drag_and_drop_v2.py
+msgid ""
+"Information about current positions of items that a learner has dropped on "
+"the target image."
+msgstr ""
+
+#: drag_and_drop_v2.py
+msgid "Indicates whether a learner has completed the problem at least once"
+msgstr ""
```
itsjeyd: @asadiqbal08 I noticed that this file does not contain entries for strings from `js_templates.html`. The templates in this file are Handlebars templates, and they use non-standard markup for strings that should be translated (see [here](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/templates/html/js_templates.html#L15) for an example). That's why the strings they contain are not being added to the translation file.

You didn't specify how you are generating translation files, but maybe you could try and see if you can modify the `makemessages` command to pick up strings from `js_templates.html`? See [here](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#customizing-makemessages) for more info. Having translations available for these strings would be helpful for course authors.

itsjeyd:
@asadiqbal08 I had another look. Your changes to `drag_and_drop_v2.py` look good, but I had another question about the translation file (inline comment above).

Once that's addressed could you please add some instructions for manual testing to the PR description? (See [this comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/66#issuecomment-202882294) for assumptions you can make.) What's the intended workflow for generating and using translation files with DnDv2?

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
asadiqbal08: @itsjeyd I am using [poedit](https://poedit.net/) to create these .po files from python source files. I am not considering .html files in this process because we cannot refer the `ugettext runtime service` in templates. 
If you require to add the .html strings in text.po then I can do it.

asadiqbal08:
@itsjeyd 
**Instruction for manual testing :**
- mark strings for translations with `_` ( see utils.py `_()` method ) where we cannot utlize the runtime i18n service e.g. XBlock field text etc. 
- mark strings with `self.ugettext` method where we can utilize the i18n runtime service. 
- I am using [POEdit](https://poedit.net/) to create empty .po file for `en` language. It take input python source files and look around `_` and `ugettext` keywords to extract strings from source files. you can create translations for any preferred language for testing e.g. `Arabic` , `French` etc. 


As we know, [#11575](https://github.com/edx/edx-platform/pull/11575) has been merged so edx-platform will look around the XBlock messages catalogs for translations if present. 

Yes, We will update the hash tag in edx-platform after this merge.

let me know if you have anything else from my side ?

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
itsjeyd: @asadiqbal08 

> I am not considering .html files in this process because we cannot refer the ugettext runtime service in templates.

Strings that are marked for translation in `js_templates.html` are passed to `gettext` on the client via [this Handlebars helper](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/public/js/drag_and_drop_edit.js#L9), so if translations are available for them, I *think* they should get picked up dynamically.

But if there is currently no method for automatically extracting strings from all relevant file types<sup>1</sup>, it might make sense to address my comment about `js_templates.html` at another time.

CC'ing @bradenmacdonald to see what he thinks.

<sup>1</sup>That's what I'm getting from your [comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/66#issuecomment-203799017) below.

itsjeyd:
@asadiqbal08 

> As we know, #11575 has been merged so edx-platform will look around the XBlock messages catalogs for translations if present.

> Yes, We will update the hash tag in edx-platform after this merge.

If I understand correctly, with these comments you are addressing this portion of my earlier comment:

> XBlock is at version 0.4.7 (calls to self.ugettext from DragAndDropBlock don't error out). Once this PR merges, are you going to do a hash update for edx-platform to ensure that this dependency is satisfied automatically?

I just want to confirm that we are talking about the same hash tag here: My question was about the hash tag for the `XBlock` repository. When I first tried to load a DnDv2 block (on a Dogwood devstack) with the `asadiqbal08/WL-373` branch of the `xblock-drag-and-drop-v2` repository checked out, I got an `AttributeError` about the `ugettext` method. I had to manually update the `XBlock` repository to version 0.4.7 to get rid of this error.

---

You'll need to rebase this branch; it has conflicts with the base branch.

asadiqbal08:
@itsjeyd I think @mattdrayer has recently [incremented-the-xblock-release-version](https://github.com/edx/XBlock/pull/339) and This PR is also pointing to xblock-sdk==v0.1.2. Kindly look [XBlock-SDK](https://github.com/edx/xblock-sdk/pull/105)
However I was talking about the version in `github.txt` for edx-platform. @mattdrayer please share your thoughts here.

itsjeyd:
@asadiqbal08 

> I think @mattdrayer has recently incremented-the-xblock-release-version and This PR is also pointing to xblock-sdk==v0.1.2.

I'm aware that xblock-sdk is pointing to version 0.4.7 of the XBlock repository, but on Devstack, the LMS and Studio act as runtimes for DnDv2. xblock-sdk is not installed on Devstack by default, so DnDv2 uses whatever version of XBlock came with edx-platform (see [here](https://github.com/edx/edx-platform/blob/master/requirements/edx/github.txt#L75)).

> However I was talking about the version in github.txt for edx-platform.

Yes, the hash for DnDv2 will need to be updated in edx-platform as well.

asadiqbal08:
thanks @itsjeyd  , good point. Yes, We should update the XBlock reference to 0.4.7 in github.txt . @mattdrayer ^^

mattdrayer:
Ah, right -- I have submitted #12003 to update the edx-platform XBlock dependency to 0.4.7

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
bradenmacdonald: I agree with @itsjeyd; we should have a process for extracting strings from the HTML files for translation. Most XBlocks are going to have translatable strings in their templates, so this is an important feature to have. If we use something like Django's `makemessages` tool, it is already designed to include translatable strings from HTML files. If you use poedit, you have to go to PoEdit menu > Preferences > Extractors, and configure a new extractor that supports .html files. However, I don't think poedit is a tool that every XBlock author will be using; it would be nice to have a simple script optimized for XBlocks that will use gettext to pull the translatable strings out of .py and .html files and save them into the correctly-named catalog. (It's basically just a matter of calling `xgettext` with the correct arguments)

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -186,7 +183,7 @@ def studio_view(self, context):
 
         js_templates = loader.load_unicode('/templates/html/js_templates.html')
         help_texts = {
-            field_name: self._(field.help)
+            field_name: self.ugettext(field.help)
```
itsjeyd: @asadiqbal08 @mattdrayer 

I used Poedit and the `text.po` file for English to create `text.po` and `text.mo` files for French with a few translations for help texts, and put them in `/edx/app/edxapp/xblock-drag-and-drop-v2/drag_and_drop_v2/translations/fr/LC_MESSAGES/` (this is where DnDv2 is mounted on my Devstack). Then I started Studio and accessed the EDIT dialog of a unit containing a DnDv2 block. The strings didn't seem to get translated, they were still displayed in English. (Studio itself is being translated correctly.)

If I'm missing something, could you please post detailed instructions for testing this PR in Studio?

asadiqbal08: @itsjeyd You are right at that point. Xblocks Message catalog is not loading in xblock `studio_view` while it is perfectly working for `student_view.` 
I had figured out this problem recently and I am working on a [PR
](https://github.com/edx/edx-platform/pull/11987) , After this merge you can see the translations in xblocks `edit` view. 
Let me know If you have any other questions ? 
cc: @mattdrayer 
Translation

asadiqbal08: For just testing, you can mark any string(s) in `student_view` with `self.ugettext`. You will get the translated string(s)

itsjeyd: @asadiqbal08 

> For just testing, you can mark any string(s) in student_view with self.ugettext. You will get the translated string(s)

Yes, I did that yesterday and it worked fine :)

Thanks for the heads-up about translation support for `studio_view` being missing. I think the only thing that's left to address now is [this comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/66#discussion_r58084737).

asadiqbal08: @itsjeyd Yes currently I was looking into that. Actually I was running this command 
`django-admin makemessages --locale=en --extension .html` to extract string from HTML files, It is working fine but not extracting string from `js_templates.html`. It seems me that, `js_templates.html` is using `handlebars {{i18n}}` tags.

itsjeyd: @asadiqbal08 This might help with the strings in `js_templates.html`: [Customizing the makemessages command](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#customizing-makemessages).

Changelog.md
```diff
@@ -3,6 +3,11 @@ Development version:
 
 * Added the ability to specify automatic alignment of dropped items. (PR #57)
 
+Version 2.0.5 (2016-03-30)
+--------------------------
+
+* Marked strings for translations (PR #66)
```
bradenmacdonald: The line above (in "Development Version" needs to be moved into this section since it would be included in 2.0.5.)

bradenmacdonald:
Hey guys, here's my perspective on what's needed to get this merged:

1. Fix the `.po` file included in this changeset, so that it includes strings from `js_templates.html`. Do it manually if necessary, but ideally find an automatic solution and document it. (Maybe [this](https://docs.djangoproject.com/en/1.9/topics/i18n/translation/#customizing-makemessages) ? )
1. Double-check that all user-visible strings shipped with this block (all labels and default data) are marked with `ugettext` or `ugettext_noop`, and that no variables/user-data are marked with `ugettext` (unless we know for sure that the contents of the variable are guaranteed to be a static string that's been marked with `ugettext_noop`).
1. Document how to extract strings and update the .po files in this repo. Ideally you could add the appropriate `xgettext` commands to a Makefile so one can just type "make messages" to extract strings from the all the .py and .html files, and update all locales which are present. However, since [django templates require complex pre-processing before they work with xgettext](https://github.com/django/django/blob/6448873197fa4e3df3f5f03201538dc57d7643d6/django/utils/translation/trans_real.py#L571-L576), we will likely end up requiring developers to install Django and using some sort of wrapper around django `makemessages`. This new command should include the handlebars {{i18n}} messages, and should be something that is easily re-usable in other XBlocks (maybe make it part of the SDK or xblock-utils).

I think it's sensible to split part 3 out to a separate story (and perhaps a discovery story?), since I'm sure it's more work than you were expecting for this PR.

bradenmacdonald:
It would also be really nice to include in this repo an example translation to `eo` (Esperanto) using the [edX Dummy Translation tool](https://github.com/edx/i18n-tools). Then anyone who wants to hack on this can just switch to Esperanto and confirm that it's all working, without needing to generate a translation first.

asadiqbal08:
@bradenmacdonald @itsjeyd I have incorporated your suggestions. Kindly Take a look. 

cc: @mattdrayer

bradenmacdonald:
@asadiqbal08 I don't have time for more than a quick look, but here's what I was able to see:

* I notice that the "Keyboard Help" link is marked as `gettext('...')` but it doesn't show up in the .po files nor does it get translated. Same with the "Feedback" heading. I understand that JS client-side `gettext()` calls may not be working yet (?), but I think we should add those to the .po files at least.

* Also a lot of string from the Studio editor seem to be missing from the diff and your po files: "Introductory Feedback", "Maximum Score", "Background URL", "Zone labels", "Display label names on the image", "should provide sufficient information to place the item even if the image did not load", and others... Does the studio editor UI appear fully translated to you when you test it? I could be missing something, so will let @itsjeyd comment further and take it from here.

itsjeyd:
@asadiqbal08 I agree with @bradenmacdonald that it would be good to add strings marked with `gettext` (from `drag_and_drop.js` and `drag_and_drop_edit.js`) to the .po files, as well as strings from the remaining .html files (`drag_and_drop.html` and `drag_and_drop_edit.html`). In terms of the Studio editor UI appearing fully translated, you already [mentioned](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/66#discussion_r58182349) that there's a separate [PR](https://github.com/edx/edx-platform/pull/11987) for enabling this. So as far as I can tell, it'll be a matter of making sure that all strings that are part of the Studio editor UI are translated correctly when that PR merges (and not just the help texts that are passed to `self.ugettext` in `studio_view`).

asadiqbal08:
@itsjeyd @bradenmacdonald
Runtime service in this PR will only look around the strings that are related to `help` texts. As I had marked these strings with `self.ugettext` runtime service. You can see the logic [Here](https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/modulestore/django.py#L286) 

Few things that I want to highlight. 
- JS client-side gettext() calls is not working yet.
- Templates strings will not be translated with the runtime logic. 

However as per request, I have updated the .PO file with strings that are related to *.html, *.js files and also created a `eo` directory for testing. 
@itsjeyd You are right, After the merge of [PR#11987](https://github.com/edx/edx-platform/pull/11987) you can see the `help` text translated.

itsjeyd:
@asadiqbal08 So to sum up, the following items seem to be out of scope for this PR:

* Making JS client-side `gettext()` calls work: **no PR** (?)
* Making strings from HTML templates such as `drag_and_drop.html` and `drag_and_drop_edit.html` appear translated: **no PR** (?)
* Making strings that are passed to `ugettext` in `studio_view` appear translated: [PR](https://github.com/edx/edx-platform/pull/11987)
* Find a way to automatically extract all relevant strings and update the .po files in this repository, and document it: **no PR, need to do discovery first**

Does that summarize the current situation accurately? Are there any plans (or even PRs) for addressing the first two items?

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
```diff
@@ -0,0 +1,547 @@
...
+msgstr ""
+"Project-Id-Version: \n"
+"POT-Creation-Date: 2016-03-30 16:54+0500\n"
+"PO-Revision-Date: 2016-03-30 16:54+0500\n"
+"Last-Translator: \n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Language: en\n"
```
itsjeyd: @asadiqbal08 Should this be "eo" instead of "en"?

asadiqbal08:
Shortly and currently we are only utilizing `Modulei18nService` in python side However I will tag @mattdrayer here to share his thoughts.

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
asadiqbal08: I created this file with the help of [i18n-tool](https://github.com/edx/i18n-tools). I ran `i18n_tool dummy` command. Yes it should be `eo`

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
```diff
@@ -0,0 +1,547 @@
...
+msgid ""
+"Optional numerical value (if you set this, learners will be prompted for "
+"this value after dropping this item)"
+msgstr ""
+"Öptïönäl nümérïçäl välüé (ïf ýöü sét thïs, léärnérs wïll ßé prömptéd för "
+"thïs välüé äftér dröppïng thïs ïtém) Ⱡ'σяєм ιρѕυм ∂σłσя ѕ#"
+
+#: templates/html/js_templates.html
+msgid ""
+"Margin B1 (when a numerical value is required, values entered by learners "
```
itsjeyd: @asadiqbal08 `B1` ⟶ `±`

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
```diff
@@ -0,0 +1,547 @@
...
+"must not differ from the expected value by more than this margin; default is"
+" zero)"
+msgstr ""
+"Märgïn B1 (whén ä nümérïçäl välüé ïs réqüïréd, välüés éntéréd ßý léärnérs "
```
itsjeyd: @asadiqbal08 `B1` ⟶ `±`

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
```diff
@@ -0,0 +1,456 @@
...
+
+#: templates/html/js_templates.html
+msgid ""
+"Optional numerical value (if you set this, learners will be prompted for "
+"this value after dropping this item)"
+msgstr ""
+
+#: templates/html/js_templates.html
+msgid ""
+"Margin B1 (when a numerical value is required, values entered by learners "
```
itsjeyd: @asadiqbal08 `B1` ⟶ `±`

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
```diff
@@ -0,0 +1,456 @@
...
+"Language-Team: \n"
+"Language: en\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Generator: Poedit 1.8.7\n"
+"X-Poedit-Basepath: ../../..\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"X-Poedit-SearchPath-0: default_data.py\n"
+"X-Poedit-SearchPath-1: drag_and_drop_v2.py\n"
```
itsjeyd: @asadiqbal08 Could you remove Poedit-specific info from this block (`X-Generator`, `X-Poedit-Basepath`, `X-Poedit-SearchPath-0`, `X-Poedit-SearchPath-1`)?

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
```diff
@@ -0,0 +1,547 @@
...
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Language: eo\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"X-Generator: Poedit 1.8.7\n"
+"X-Poedit-Basepath: ../../..\n"
+"X-Poedit-SearchPath-0: default_data.py\n"
+"X-Poedit-SearchPath-1: drag_and_drop_v2.py\n"
```
itsjeyd: @asadiqbal08 The same comment about removing Poedit-specific info applies here.

itsjeyd:
@asadiqbal08 Thanks for creating the remaining stories in JIRA. I left some additional comments that should be easy to address. :+1: from me once you've done that.

Before merging please double-check if you need another +1 -- maybe from @mattdrayer?

asadiqbal08:
@mattdrayer we have gotten thumbs-up from @itsjeyd but your one is required :)

mattdrayer:
AH!  Sorry, I forgot to do this yesterday :smile:

:+1:!!!

asadiqbal08:
@bradenmacdonald can you please update the release tag ? 
Jenkins is complaining about release version on [PR#12147](https://github.com/edx/edx-platform/pull/12147)

mattdrayer:
Just a heads up that we are all set with the release tag at this point -- have a nice vacation @bradenmacdonald 😄

PR #67, opened 2016-05-03
Removes role="application"

pomegranited:
Removes role="application" from xblock wrapper and tests.

**Sandbox URL**: 

* Sandbox *with* `role="application"`: http://pr12338.sandbox.opencraft.com/
* Sandbox *without* `role="application"`: http://studio.pr12340.sandbox.opencraft.com/

**Testing instructions**:

1. To run automated tests:

```bash
$ sh install_test_deps.sh
$ python run_tests.py
```

2. Login to the above sandboxes with `staff@example.com`/`edx`, and visit course named "Testing DnDv2 Accessibility".  
    Each unit contains an DnDv2 xblock, which you can test with screen readers.  

    For example, the triangle background image uses the alt text "An isosceles triangle with three layers of similar height. It is shown upright, so the widest layer is located at the bottom, and the narrowest layer is located at the top.".

**Author Notes and Concerns**

I was not able to reproduce the accessibility issue noted in SOL-1770 on any of the screen readers I tried.  I tested the Dndv2 XBlock both before and after removing the `role="application"` attribute, and got the same result for both code sets.

These screen readers successfully read all the text on the page, including instructions and image alt tags:

* [VoiceOver](https://www.apple.com/voiceover/info/guide/_1124.html): Mac, built in tool
* [Narrator](http://windows.microsoft.com/en-au/windows-10/getstarted-hear-text-read-aloud): Windows 10, built-in tool
* [TalkBack](https://support.google.com/accessibility/android/answer/6007100): Android

The screen readers below had their own idiosyncracies, but showed no difference in behaviour between the two code sets:

* [JAWS](http://www.freedomscientific.com/Products/Blindness/JAWS) - couldn't get it to read any of the text on the page, except for elements I could tab to. Maybe because I only had the free trial version?
* [NVDA](http://www.nvaccess.org/) - successfully reads any text or image alt text I can mouse over, but doesn't just read out the page.

**Reviewers**
- [ ] @itsjeyd 
- [ ] @cptvitamin

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -279,7 +279,7 @@ function DragNDropTemplates(url_name) {
                     problemHeader,
                     h('p', {innerHTML: ctx.problem_html}),
                 ]),
-                h('section.drag-container', { attributes: { role: 'application' } }, [
+                h('section.drag-container', {}, [
                     h(
                         'div.item-bank',
                         renderCollection(itemTemplate, items_in_bank, ctx)
--- a/tests/integration/test_render.py
+++ b/tests/integration/test_render.py
@@ -161,7 +161,7 @@ class TestDragAndDropRender(BaseIntegrationTest):
     def test_drag_container(self):
         self.load_scenario()
         item_bank = self._page.find_element_by_css_selector('.drag-container')
-        self.assertEqual(item_bank.get_attribute('role'), 'application')
+        self.assertIsNone(item_bank.get_attribute('role'))
 
     def test_zones(self):
         self.load_scenario()


pomegranited:
@itsjeyd If you're happy with this change, could you please ping @cptvitamin so he knows it's ready to for review?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -279,7 +279,7 @@ function DragNDropTemplates(url_name) {
                     problemHeader,
                     h('p', {innerHTML: ctx.problem_html}),
                 ]),
-                h('section.drag-container', { attributes: { role: 'application' } }, [
+                h('section.drag-container', { attributes: { } }, [
```
itsjeyd: @pomegranited Any specific reason to keep the `attributes: {}` portion of the second argument to `h`?

pomegranited: @itsjeyd No good reason no.. removed with [b5ab955](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/67/commits/b5ab955f96ef1adb6d14699597872ad705f3355c).

itsjeyd:
@pomegranited I tested with Firefox and Orca. For me, Orca refuses to read the entire page (nothing happens when pressing the associated keyboard shortcut), so I couldn't verify the behavior for that specific use case. But I was able to use [structural navigation](https://help.gnome.org/users/orca/stable/commands_structural_navigation.html.en) to navigate to the background image and have Orca read the alt text.  

The code looks good (thanks for the update re: my comment).

itsjeyd:
@pomegranited :+1:

itsjeyd:
@cptvitamin This is ready for you to have a look. See the PR description for sandbox links. Thanks!

pomegranited:
Hi @cptvitamin, have you had a chance to look at this change?

cptvitamin:
@pomegranited This worked just fine without the role=application, so I say let's go with this.  I had a couple other pieces of feedback I wanted to share:

1. The aria-live region needs to be scoped to just the div that contains the notification messages. Because of the way the DOM is being updated, the screen reader thinks that entire region has been updated and reads the *entire* contents of that div, which is a lot of content.  Let's try scoping it to just the div with the class of popup.  The only problem there is that there is a display:none on that div.  You can't actually toggle the visibility of a live region, it has to be present in the DOM at page load, so you'll have to add another div as a child that actually does the show/hide.
2. When tabbing through the focusable controls, the Windows user is never exposed to the alt text for the background image, which is critical to understanding the context of the problem.  I think a reasonable solution would be to add an ID to this element and then reference it via aria-describedby on all the dropzones.  The screen reader user would move focus to the dropzone, hear its description, and then hear the description of its meaning in the greater context of the problem.
3. In certain screen readers, the keyboard controls are not exposed.  On Windows and NVDA< only the CTRL + M keycombo works to drag and drop the items (many users will not be aware of this combination).  It would be very cool to add an additional aria-describedby that references that bit of help text that is available in the Keyboard help modal: "Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on." on the draggable items and "Press "Enter", "Space", "Ctrl-m", or "⌘-m" to drop the item on the current zone." on the dropzones.

cptvitamin:
@pomegranited one last thing I forgot to mention was that when I went to a dropzone that had an item in it, it was not read back to me, so I had no idea if a dropzone had any items in it or not.

cptvitamin:
nope, I support the removal of role=application here.

pomegranited:
@cptvitamin Lovely, thank you for such a complete description of the accessibility issues with this XBlock.  I'm merging this change, and we'll address the other issues in a subsequent PR.

CC @bradenmacdonald

bradenmacdonald:
Merging now, though note that these changes won't appear in any edx release until we do a version bump and update the requirements file.

PR #75, opened 2016-07-07
Mode field

e-kolpakov:
**Description:** This PR adds a mode field to DnDv2 with to possible values: "standard" and "assessment" + studio editing capabilities.

**Dependencies:** Based on https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/74

**Sandbox LMS URL**: http://dndv2.opencraft.hosting/
**Sandbox Studio URL**: http://studio-dndv2.opencraft.hosting/

**Testing instructions:**
1. Install XBlock into LMS/Studio
2. Enable in Advanced settings
3. Create a course, section, subsection and unit
4. Add DnDv2 block to a unit
5. Click edit. *Expected result*: Editor popup opens; "Problem mode" select with two values "standard" and "assessment" are present, with "standard" selected
6. Change "Problem mode", save DnDv2 (note the edit dialog have three steps)
7. Click "Edit" again. *Expected result*: Editor popup opens, "Problem mode" is present and previously saved value is selected.

Reviewers:

* [ ] @haikuginger 
* [ ] edX reviewer[s] TBD

PR #74, opened 2016-07-06
Remove support for numerical input.

mtyaka:
This PR removes support for numeric input feature. Nobody is using that feature anymore and it adds a lot of complexity to the codebase.

**JIRA tickets**: https://openedx.atlassian.net/browse/SOL-1870

**Dependencies**: None

**Sandbox LMS URL**: http://dndv2.opencraft.hosting/
**Sandbox Studio URL**: http://studio-dndv2.opencraft.hosting/

**Testing instructions**:

This PR does not add any new features and makes no user-visible changes (assuming users were not using the numeric input feature) so there is no specific functionality to test.
Testers should verify that DnDv2 continues to function correctly.

**Reviewers**
- [x] @e-kolpakov 
- [ ] edX reviewer[s] TBD

PR #77, opened 2016-07-08
Problem mode field and edit capabilities

e-kolpakov:
Reincarnation of #75 

**Description:** This PR adds a mode field to DnDv2 with to possible values: "standard" and "assessment" + studio editing capabilities.

This is the first of a series of PRs that will be implementing the new "[Assessment mode](https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit)". This story is only adding the editor UI, and does not yet implement functionality changes, which are coming in subsequent stories.

**Dependencies:** Based on https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/74
**JIRA:** [SOL-1933](https://openedx.atlassian.net/browse/SOL-1933)

**Sandbox LMS URL**: http://dndv2-sandbox2.opencraft.hosting/
**Sandbox Studio URL**: http://studio-dndv2-sandbox2.opencraft.hosting/

**Screenshot:**
image

**Testing instructions:**
1. Install XBlock into LMS/Studio
2. Enable in Advanced settings
3. Create a course, section, subsection and unit
4. Add DnDv2 block to a unit
5. Click edit. *Expected result*: Editor popup opens; "Problem mode" select with two values "standard" and "assessment" are present, with "standard" selected
6. Change "Problem mode", save DnDv2 (note the edit dialog have three steps)
7. Click "Edit" again. *Expected result*: Editor popup opens, "Problem mode" is present and previously saved value is selected.

Reviewers:

* [x] @haikuginger 
* [x] @cahrens
* [x] @staubina


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=77&repo=edx-solutions%2Fxblock-drag-and-drop-v2

e-kolpakov:
@haikuginger I've accidentally merged #75, so I had to revert it n #76 and now created it anew as #77 :)

This is basically the same changeset, just squashed and rebased against most recent master, so could you please cast +1 here as well? You can take another look of course :)

haikuginger:
@e-kolpakov, it looks like your merge/unmerge caused #74 to be merged (and then reverted) as well, but unlike in #75, you're not including those commits in your PR, although some of its changes are in your README.md. Can you either remove those README modifications from your branch, or rebase on top of the branch for #74 so that the commit messages are accurate?

We'll also need to reopen #74; not sure of the process on that.

Thanks!

haikuginger:
Casting a 👍 here in addition to the original one on #75. I've re-tested the functionality and it appears to be working; the PR structure noted above is a nit, albeit one I'd still like taken care of. 🙂 Given that this works without #78, I think the best solution would be to just remove the README bits that #78 produces, and re-squash onto master.

e-kolpakov:
@marcotuts @bradenmacdonald reincarnation of #75

bradenmacdonald:
@e-kolpakov Can you please add a screenshot to the PR description? I'd like us to include screenshots in any PRs we open that touch UI.

marcotuts:
[Aside] Not relevant to this PR necessarily, but should "Continue" in the editor modal be "Save" to echo other components in Studio?

haikuginger:
@marcotuts: To configure the XBlock, there are actually three progressive pages; on the last one, the Continue button changes to Save.

marcotuts:
ah! got it thanks @haikuginger. 

I still wasn't able to see the mode option but am now able to see the multiple pages after clicking the Continue button.

e-kolpakov:
@cahrens @staubina This PR is ready for your review

staubina:
Code seems straightforward and testing went as expected.

👍

marcotuts:
This is out of scope for this PR but it would be great to explore two items in the future:
1.) Tabs for the editing modes / steps up top instead of the multi-part editing experience (similar to ORA)
2.) Updates to the styling to echo the other settings which incoporate help text strings.

-----
3.) In the shorter term, I think it would be helpful to add a support string to clarify what Feedback & assessment mode are in context if possible, perhaps just underneath that mode setting dropdown?   Without this the authoring is still possible, so it doesn't need to  block the merge of this PR necessarily. 

thanks!

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -32,13 +32,30 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
...
         display_name=_("Title"),
         help=_("The title of the drag and drop problem. The title is displayed to learners."),
         scope=Scope.settings,
         default=_("Drag and Drop"),
     )
 
+    mode = String(
+        display_name=_("Mode"),
+        help=_(
+            "Standard mode: feedback is provided to learner right after an item is dropped to a zone. "
```
cahrens: If this help documentation will be displayed to course authors (currently it is not), then I would recommend having someone from Documentation help wordsmith it.

bradenmacdonald: @pdesjardins ^ are you our go-to guy for dndv2.1 docs requests?

pdesjardins: Yes, I'm covering DnDv2.

"Standard mode: the problem provides immediate feedback each time a learner drops an item on a target zone."

"Assessment mode: the problem provides feedback only after a learner drops all available items on target zones."

I can go through the rest of the file, when do you need that?

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -21,6 +21,16 @@
                       <span class="sr">{{ help_texts.show_title }}</span>
                     </label>
 
+                    <label class="h3" for="problem-mode" title="{{ help_texts.mode }}">{% trans "Problem mode" %}</label>
+                    <select id="problem-mode">
```
cahrens: Take care with using ids like this. We have to make sure that ids are unique on the page, even if there are multiple instances of the xblock within the unit. In this particular case it is OK because we only allow a single component to be edited at a time, but seeing an id like this still makes me cringe.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -21,6 +21,16 @@
                       <span class="sr">{{ help_texts.show_title }}</span>
                     </label>
 
+                    <label class="h3" for="problem-mode" title="{{ help_texts.mode }}">{% trans "Problem mode" %}</label>
```
cahrens: Why isn't the display name as defined on the xblock field used instead of hardcoding/duplicating the display name?

cahrens:
I had a few questions on the template, but otherwise 👍

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
e-kolpakov: @cahrens noted - I've added an item to our internal backlog to convert ids to classes or something better suited the task.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
e-kolpakov: @cahrens I wasn't involved in previous a11y review, so I'm afraid I can't answer these questions - basically I've followed an established pattern. I've created a task in our internal backlog to investigate/address this issue.

e-kolpakov:
@cahrens The questions you've brought are important, but addressing them right away might not be a good idea as:

* It might not be what we want (I would assume we had titles and sr spans for a reason - after all it passed a11y review)
* Addressing them in scope of this PR feels a little wrong - the scope of changes to make will likely be even larger than what we have in this PR now.

So, I've created tasks to investigate those questions and address them (if needed) in future PRs.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @pdesjardins @cahrens @bradenmacdonald merged this PR before pushed the update for the docstring - sent as a separate PR #79. Since it's just a copy update I'm going to merge it as soon as CI build passes.

bradenmacdonald:
@marcotuts Can you please verify this on http://studio-dndv2-sandbox2.opencraft.hosting/ ? The Mode field is definitely updated on the sandbox now.

The help text will appear as a tooltip if you hover over the "Problem Mode" setting, and we have a ticket opened to improve that to display like other XBlock/XModule settings do (below the field rather than as a tooltip).

screen shot 2016-07-19 at 12 41 53 pm

bradenmacdonald:
@sstack22 ^ Same question, instead of Marco :)

sstack22:
@bradenmacdonald - verified! @pdesjardins - question on the tool tip text. Do you think we should add a note to clarify that either mode can be used for a graded assignment? I think there might be a bit of confusion, since it's referred to as "assessment" mode.

pdesjardins:
@e-kolpakov and @bradenmacdonald, I was just looking at the sandbox (in studio) and found that in assessment mode, dropping an item on the wrong target sent the item back to the item area. I expected the item to remain on the incorrect target until all items were dropped, then tell me that I was wrong and use up one of my allowed attempts. I can try an item on each zone until I find a correct zone. Is that the expected behavior for assessment mode? I don't see why there would need to be an allowed number of attempts for the problem since the only way to complete the problem is to get all the items correct.

bradenmacdonald:
@pdesjardins Please try http://studio-dndv2-sandbox8.opencraft.hosting/ (from SOL-1944). This PR only implemented a small part of the feature, so its sandbox is not fully functional.

PR #80, opened 2016-07-19
Remove support for numerical input

mtyaka:
**NOTE**

The changes in this PR were initially submitted in #74, which was merged into master by accident before the review process was finished. The PR was later re-opened as #78, the changes were properly reviewed and approved, but after rebasing that PR on top of master github detected that the commits had already been previously merged into master and automatically closed that issue. The commits were indeed present in master, but the changes were somehow semi-manually reverted in #76, so this PR simply re-creates the same commits that were in already reviewed and approved in #78.

**Background**

This PR removes support for numeric input feature. Nobody is using that feature anymore and it adds a lot of complexity to the codebase.

The numeric input feature was a way to prompt students to enter a numeric value after dragging an item onto the target.

**Screenshot of the input field which is now removed**:
screen shot 2016-07-13 at 11 18 50 am

**JIRA ticket**: https://openedx.atlassian.net/browse/SOL-1870

**Dependencies**: None

**Sandbox LMS URL**: http://dndv2.opencraft.hosting/
**Sandbox Studio URL**: http://studio-dndv2.opencraft.hosting/

**Testing instructions**:

This PR does not add any new features and makes no user-visible changes (assuming users were not using the numeric input feature) so there is no specific functionality to test.
Testers should verify that DnDv2 continues to function correctly.

**Reviewers**
- [x] Solutions: @e-kolpakov 
- [x] TNL: @staubina

PR #83, opened 2016-07-20
[SOL-1909] Multiple Drop Targets per Item

haikuginger:
This PR implements SOL-1909, and allows instructors to select multiple valid options for each item in a Drag and Drop v2 problem. It is fully backwards-compatible with existing problem sets.

We approach this by changing the `zone` attribute on each item to a `zones` attribute containing a list, which can contain an arbitrary number of zone IDs.  Additionally, a compatibility layer has been introduced such that existing datasets with `zone` are mutated before being passed to the frontend to have a list containing the existing single item (in cases where there's an ID value) or an empty list (in cases where `zone` is set to `"none"`). This won't migrate existing datasets automatically, but they will be updated the next time the XBlock configuration is saved. 

**JIRA tickets**: Implements [SOL-1909](https://openedx.atlassian.net/browse/SOL-1909)

**Dependencies**: None

**Screenshots**:

screen shot 2016-07-22 at 11 24 47 am

screen shot 2016-07-22 at 11 24 35 am

screen shot 2016-07-20 at 2 07 35 pm

screen shot 2016-07-20 at 2 07 23 pm

screen shot 2016-07-20 at 2 07 08 pm

**Sandbox URL**:

- LMS: http://dndv2-sandbox4.opencraft.hosting/

- Studio: http://studio-dndv2-sandbox4.opencraft.hosting/

**Testing instructions**:

1. In Studio, create a course or update a course to include the DnDv2 XBlock

2. In Studio, observe that the dropdown select menu for Zone selection has changed to be multiple checkbox controls.

3. In Studio, observe that existing configurations are retained and presented appropriately for the new control set.

4. In Studio, create a new item that can be assigned to multiple zones.

5. In LMS, observe that pre-existing items have appropriate behavior when placed on the zone they were previously assigned to.

6. In LMS, observe that the item you created can be dragged to any of the zones you selected for it.

7. In LMS, review the XBlock demo course and observe that it works without modification, and that changes to take advantage of the feature allowing multiple correct answers works immediately.

**Reviewers**
- [x] @itsjeyd
- [x] a11y: @cptvitamin and/or @clrux
- [x] Product: @sstack22

Also tagging @pdesjardins as a docs FYI.


edx-webhook:
Thanks for the pull request, @haikuginger! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=83&repo=edx-solutions%2Fxblock-drag-and-drop-v2

drag_and_drop_v2/default_data.py
```diff
@@ -97,6 +98,20 @@
...
+      "displayName": _("Goes anywhere on the triangle"),
+      "feedback": {
+        "correct": ITEM_ANY_ZONE_FEEDBACK,
+        "incorrect": ITEM_INCORRECT_FEEDBACK
+      },
+      "zones": [
+        TOP_ZONE_ID,
+        BOTTOM_ZONE_ID,
+        MIDDLE_ZONE_ID
+      ],
```
itsjeyd: @haikuginger Can you update the other items to specify a list of `"zones"` as well?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -166,7 +166,10 @@ def items_without_answers():
             items = copy.deepcopy(self.data.get('items', ''))
             for item in items:
                 del item['feedback']
-                del item['zone']
+                # Switching to `pop` from `del`, as we no longer have any guarantee of
+                # either `zone` or `zones` being present.
```
itsjeyd: @haikuginger Thanks for adding this comment!

Right now it reads a bit like a commit message. How about something like:

> Use `pop` instead of `del` to remove target zone info. This is necessary because we don't have any guarantee of either `zone` or `zones` being present (legacy instances of this block allow only a single `zone` per item; new instances allow authors to specify a list of `zones`).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -233,8 +236,18 @@ def studio_view(self, context):
         for js_url in js_urls:
             fragment.add_javascript_url(self.runtime.local_resource_url(self, js_url))
 
+        # Do a bit of manipulation so we get the appearance of a list of zone options on
+        # items that still have just a single zone stored
+        data = copy.deepcopy(self.data)
```
itsjeyd: @haikuginger Can we do without copying `self.data` here? My understanding is that we don't necessarily want to preserve `item["zone"]` entries for `data` belonging to old instances; we just want these instances to keep working :) So it would be OK for this code to replace `item["zone"]` with `item["zones"]` I think (moving us closer to being able to remove backward compatibility hacks over time).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -321,6 +334,15 @@ def reset(self, data, suffix=''):
         self.item_state = {}
         return self._get_user_state()
 
+    def _is_attempt_correct(self, attempt):
+        """
+        Bringing the logic to determine whether a placement is correct out of do_attempt to
+        improve clarity on the process for cases where we have multiple valid options.
```
itsjeyd: @haikuginger This also reads a bit like a commit message. I think you can just stick to explaining what the method does ("Check if item was placed correctly.").

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -321,6 +334,15 @@ def reset(self, data, suffix=''):
...
+        """
+        valid_options = self._get_valid_item_zones(attempt['val'])
```
itsjeyd: @haikuginger `valid_options` is a bit generic. I think it would make sense to rename this to `correct_zones`, and rename `_get_valid_item_zones` to `_get_correct_item_zones` (or even just `_get_item_zones`, since item definitions do not explicitly store information about incorrect zones).

---

You can remove the blank line below (L343).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -373,12 +395,15 @@ def _get_user_state(self):
...
         for item_id, item in item_state.iteritems():
-            definition = self._get_item_definition(int(item_id))
             # If information about zone is missing
             # (because problem was completed before a11y enhancements were implemented),
             # deduce zone in which item is placed from definition:
             if item.get('zone') is None:
-                item['zone'] = definition.get('zone', 'unknown')
+                valid_zones = self._get_valid_item_zones(int(item_id))
+                if valid_zones:
+                    item['zone'] = valid_zones[0]
```
itsjeyd: @haikuginger It might be worth mentioning that support for multiple target zones was implemented *after* the a11y enhancements mentioned above, and that this means that `valid_zones` will at most contain a single zone for items from problems that were "completed before a11y enhancements were implemented". This is not immediately obvious from looking at the code alone (if it were possible for items without `zone` info to be associated with multiple target zones, using `valid_zones[0]` would not necessarily return the zone that a student actually placed an item in).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -408,6 +433,24 @@ def _get_item_definition(self, item_id):
         """
         return next(i for i in self.data['items'] if i['id'] == item_id)
 
+    def _get_valid_item_zones(self, item_id):
+        """
+        Returns a list of the zones that are valid options for the item. If
+        we have an existing configured list of options, return that. If we
+        have an existing single item, return that encapsulated in a list.
+        If we don't have either, which would be weird, return an empty list.
```
itsjeyd: @haikuginger Please add a line break and an empty line after the first sentence of the docstring (not all docstrings in this file currently do this, but for newly added code we should try and follow edX's [Python Guidelines](https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#docstrings)).

For the rest of the docstring, it would be good to make the terminology more specific (to better reflect the fact that the method deals with zones, not options or items):

> If the item is associated with a list of zones, return that. If the item is associated with a single zone, return that encapsulated in a list.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -408,6 +433,24 @@ def _get_item_definition(self, item_id):
...
+        item = self._get_item_definition(item_id)
+
+        if item.get('zones') is not None:
+            return item.get('zones')
+
+        elif item.get('zone') is not None and item.get('zone') != 'none':
+            return [item.get('zone')]
+
+        else:
+            return []
```
itsjeyd: @haikuginger Nit: This code is nice and readable -- no need for all that vertical whitespace :)

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -457,11 +500,12 @@ def _is_finished(self):
...
+
         for item in self.data['items']:
-            if item['zone'] != 'none':
+            item_id = item['id']
+            if self._get_valid_item_zones(item_id) != []:
                 total_count += 1
-                item_id = str(item['id'])
-                if item_id in item_state:
+                if str(item_id) in item_state:
                     completed_count += 1
```
itsjeyd: @haikuginger Aside from the return value, the logic in this method is identical to the logic in `_get_grade` above. It would be worth moving it to a separate method (something like `_get_item_stats`) I think. Then you could just call this method from `_get_grade` and `_is_finished` like this:

```python
correct_count, total_count = self._get_item_stats()
```

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -31,7 +31,7 @@ function DragAndDropEditBlock(runtime, element, params) {
                     _fn.tpl = {
                         zoneInput: Handlebars.compile($("#zone-input-tpl", element).html()),
                         zoneElement: Handlebars.compile($("#zone-element-tpl", element).html()),
-                        zoneDropdown: Handlebars.compile($("#zone-dropdown-tpl", element).html()),
+                        zoneCheckbox: Handlebars.compile($("#zone-checkbox-tpl", element).html()),
```
itsjeyd: @haikuginger In many cases it'll be more than one checkbox -- maybe generalize to `zoneCheckboxes`?

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -357,26 +357,20 @@ function DragAndDropEditBlock(runtime, element, params) {
...
                         var zoneObjects = _fn.build.form.zone.zoneObjects;
 
                         zoneObjects.forEach(function(zoneObj) {
-                            dropdown.push(template({
-                                uid: zoneObj.uid,
+                            checkboxes.push(template({
+                                zoneUid: zoneObj.uid,
                                 title: zoneObj.title,
-                                selected: (zoneObj.uid == selectedUID) ? 'selected' : '',
+                                checked: $.inArray(zoneObj.uid, selectedZones) != -1 ? 'checked' : '',
```
itsjeyd: @haikuginger The previous version of the code wasn't doing it, but we generally want to use strict comparison (`===` and `!==`).

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -469,7 +462,9 @@ function DragAndDropEditBlock(runtime, element, params) {
                             if (name.length > 0 || imageURL.length > 0) {
                                 var data = {
                                     displayName: name,
-                                    zone: $el.find('.zone-select').val(),
+                                    zones: $.map($el.find('.zone-checkbox:checked'), function(c){
```
itsjeyd: @haikuginger This is definitely readable but slightly terse. You could store the checkbox elements in a variable (like the code does for `name`, `imageURL`, etc.), and do `function(checkbox) {...}` instead of `function(c) {...}`.

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: So, we definitely can, and it'd work. However, as it stands, we already get the new settings as soon as the instructor hits "save" in the settings dialog - it's a POST, not a PATCH, so `data` gets overridden entirely, including removing items that aren't in the POST (i.e., `item['zone']`). Given that, I'd prefer to only actually make the change in the database once the user's explicitly made a decision to save.

Obviously, that's something that can go either way, especially since the user isn't consciously making a choice to move to the new format when they hit Save, but I'd prefer to avoid making the change completely silently just because the Studio view was opened. Seems to go against the "GET doesn't change data" rule.

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: Good call.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
haikuginger: Absolutely correct, except this particular template only renders a single checkbox. Similarly, `zoneDropdown`, when it was there, only rendered a single selection item. So, I think `zoneCheckbox` in the singular is the better fit in this particular case; the duplication of `zoneCheckbox` for each zone is handled around line 365.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
haikuginger: Good call. I don't necessarily see it as an issue on this case, because `$.inArray` by definition returns an int, but it's always good to be robust.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
haikuginger: I think I might not be familiar enough with JS to follow this comment fully; I can definitely give `c` a better variable name, but I don't understand the construct you're suggesting - do you mean something like this?

```javascript
var checked_checkboxes = $el.find('.zone-checkbox:checked')
var data = {
    zones: $.map(checkboxes, function(checkbox){
        return checkbox.value();
    }
}
```

drag_and_drop_v2/default_data.py
haikuginger: Sure!

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: Agreed; outside the context of "this is the change that we're making", the comment doesn't really make much sense.

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: Will do.

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: Agreed.

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: So, I definitely could make that change, and it's something I thought about, but it kind of seems like that might be a good abstraction to design as part of adding Assessment Mode, because it seems likely to turn `_get_grade` and `_is_finished` on their respective heads anyway.

drag_and_drop_v2/drag_and_drop_v2.py
bradenmacdonald: @haikuginger 
> making the change completely silently just because the Studio view was opened

Was that actually happening? I don't remember off the top of my head whether an XBlock write to a `Scope.content` field from a Studio `student_view` has any effect or not; there's a chance it does not. (Studio uses a different runtime for previews vs. XBlock edits. And the LMS student_view definitely cannot modify Scope.content fields.)

itsjeyd: @haikuginger That makes sense, I'm fine with keeping the existing approach. Perhaps just adjust/extend the comment above to mention why you're copying `self.data`.

drag_and_drop_v2/drag_and_drop_v2.py
itsjeyd: @haikuginger Hm, I don't think that grade calculation will change in assessment mode (more specifically, *how* grades are calculated should stay the same; *when* they are calculated will change), and standard mode will keep using the current version of `_is_finished`. Are there any specific changes that you think would make it necessary to touch a `_get_item_stats` method again if we implemented it now?

drag_and_drop_v2/public/js/drag_and_drop_edit.js
itsjeyd: @haikuginger 

> except this particular template only renders a single checkbox

I missed that somehow -- totally fine to keep `zoneCheckbox` then!

drag_and_drop_v2/public/js/drag_and_drop_edit.js
itsjeyd: @haikuginger Yes, that's exactly what I meant :) You won't need the `var` though; you can just add the declaration to the existing list of declarations above the `if` (keeping the order in which the fields appear):

```js
                            var $el = $(el),
                                name = $el.find('.item-text').val(),
                                $zones = $el.find('.zone-checkbox:checked'),
                                imageURL = $el.find('.item-image-url').val(),
                                imageDescription = $el.find('.item-image-description').val();

                            if (name.length > 0 || imageURL.length > 0) {
                                var data = {
                                    displayName: name,
                                    zones: $.map($zones, function(zone) {
                                        return zone.value;
                                    }),
                                    // ...
```

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: So I guess the question is, "what will the behavior of submitted, but incorrect, items be in Assessment Mode?" If they're added to `item_state`, but tagged as incorrect, then the logic for `_get_grade` would have to adapt to account for the fact that not all items in `item_state` are correct when running in assessment mode.

If we're fine with leaving incorrect answers out of `item_state`, then yeah, refactoring the logic  into another method makes sense.

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: >Was that actually happening? I don't remember off the top of my head whether an XBlock write to a Scope.content field from a Studio student_view has any effect or not; there's a chance it does not.

So, it wouldn't just be opening the Studio view - it'd be if they opened the Studio view and then opened the editor for the block. Looking at the existing code, it looks like the database methods are completely abstracted - I can't tell if a method needs to be accessed in a particular way for an attribute change to be committed.

I'll see if I can do some further testing to verify the behavior; if it works like I think it does, then I'll add the comment @itsjeyd mentioned; if not, then I'll remove the `copy.deepcopy()` call.

haikuginger: Okay, fun investigation on this one! This may be old news to you guys, but it looks like only when a function is marked as a handler (for example, via the `@XBlock.json_handler` decorator) will changes to the XBlock state be saved automatically. Because `studio_view()` is not a handler, we can mess with `self` as much as we want, and no changes will be saved to the backing store.

I'm going to remove the `copy.deepcopy` call, but I'm going to add a comment noting exactly why we can get away with that.

bradenmacdonald: That's along the lines of what I suspected, but I wasn't totally sure. Thanks for checking!

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: Found a way to go with this that preserves current behavior and gives us a comprehensive starting place for the future.

README.md
```diff
@@ -131,17 +131,16 @@ items, as well as the drag items themselves. A drag item can contain
 either text or an image. You can define custom success and error
 feedback for each item. The feedback text is displayed in a popup
 after the learner drops the item on a zone - the success feedback is
-shown if the item is dropped on the correct zone, while the error
+shown if the item is dropped on a correct zone, while the error
 feedback is shown when dropping the item on an incorrect drop zone.
 
-![Zone dropdown](/doc/img/edit-view-zone-dropdown.png)
```
itsjeyd: @haikuginger This was the only place where `edit-view-zone-dropdown.png` was referenced; you can remove that file from the repo now.

README.md
```diff
@@ -131,17 +131,16 @@ items, as well as the drag items themselves. A drag item can contain
...
+You can select any number of zones for an item to belong to using
+the checkboxes; all zones defined in the previous step are available.
+You can leave none of the checkboxes checked in order to create a
```
itsjeyd: @haikuginger "You can leave all checkboxes unchecked", perhaps?

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -82,45 +82,58 @@
...
-            <label for="item-{{id}}-zone">{{i18n "Zone"}}</label>
-            <select id="item-{{id}}-zone"
-                    class="zone-select">{{ dropdown }}</select>
             <a href="#" class="remove-item hidden">
                 <div class="icon remove"></div>
             </a>
         </div>
         <div class="row">
-            <label for="item-{{id}}-image-url">{{i18n "Image URL (alternative to the text)"}}</label>
+            <fieldset id="item-{{ id }}-zone">
```
itsjeyd: @haikuginger `item-{{ id }}-zone` doesn't seem to be referenced anywhere. If that's the case you can remove it.

drag_and_drop_v2/default_data.py
```diff
@@ -83,19 +88,34 @@
...
+      "zones": [
+        BOTTOM_ZONE_ID
+      ],
       "imageURL": "",
       "id": 2,
     },
     {
+      "displayName": _("Goes anywhere"),
+      "feedback": {
+        "correct": ITEM_ANY_ZONE_FEEDBACK
```
itsjeyd: @haikuginger This is missing an `"incorrect": "",` entry; please add that above this line.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -263,11 +280,11 @@ def do_attempt(self, attempt, suffix=''):
 
         state = None
         zone = None
-        feedback = item['feedback']['incorrect']
+        feedback = item['feedback'].get('incorrect', '')
```
itsjeyd: @haikuginger Once you update the default exercise as mentioned above (i.e., once correct and incorrect feedback is set for all items), you should be able to revert this change.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -321,6 +338,14 @@ def reset(self, data, suffix=''):
         self.item_state = {}
         return self._get_user_state()
 
+    def _is_attempt_correct(self, attempt):
+        """
+        Check if the item was placed correctly.
+        """
+        correct_zones = self._get_item_zones(attempt['val'])
+
```
itsjeyd: @haikuginger This is looking great now! :) Since this method is so short, this empty line should be removed I think.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -432,39 +482,38 @@ def _get_zone_by_uid(self, uid):
...
         item_state = self._get_item_state()
 
-        for item in self.data['items']:
-            if item['zone'] != 'none':
-                total_count += 1
-                item_id = str(item['id'])
-                if item_id in item_state:
-                    correct_count += 1
+        required_items = [str(item['id']) for item in all_items if self._get_item_zones(item['id']) != []]
+        completed_items = [item for item in required_items if item in item_state]
```
itsjeyd: @haikuginger That's great that you went ahead and introduced this method :)

I think you can simplify it a bit though: Since #80, placing an item on a correct zone is equivalent to "completing" it. (Before, there was an option for course authors to define items that required a numeric input; those items were considered "completed" iff they had been placed on a correct zone *and* a numeric input had been submitted for them.) So we don't need a "completed" stat; `_get_grade` and `_is_finished` just need `correct_count` and `total_count` to produce an appropriate result.

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -187,11 +188,18 @@
...
+}
+
 .xblock--drag-and-drop--editor .items-form .item-image-url {
     width: 81%;
     margin: 0 1%;
 }
 
+.xblock--drag-and-drop--editor .items-form .item-text {
+    margin: 0 1%;
+}
```
itsjeyd: @haikuginger DRY: You can remove this rule and add the selector to the rule in L181 above.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -465,11 +458,14 @@ function DragAndDropEditBlock(runtime, element, params) {
                                 name = $el.find('.item-text').val(),
                                 imageURL = $el.find('.item-image-url').val(),
                                 imageDescription = $el.find('.item-image-description').val();
+                                checkedCheckboxes = $el.find('.zone-checkbox:checked')
```
itsjeyd: @haikuginger I think it would be worth renaming this to `selectedZones`, to focus more on the information that this code is supposed to retrieve (and focus less on the current UI choice of using checkboxes for zone selection, which is arbitrary).

Also, you'll need to add `;` at the end of this line and replace `;` in the previous line with `,`.

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -82,45 +82,58 @@
...
+            {{ title }}
+        </label>
+    </div>
 </script>
 
 <script id="item-input-tpl" type="text/html">
     <div class="item">
         <div class="row">
-            <label for="item-{{id}}-text">{{i18n "Text"}}</label>
+            <label for="item-{{id}}-text" class="h3 zone-checkbox-label">{{i18n "Text"}}</label>
```
itsjeyd: @haikuginger Probably just a typo -- if we need an additional class to style this, it would make sense to call it `zone-text-label` I think.

tests/unit/data/html/data.json
```diff
@@ -37,7 +37,10 @@
         "incorrect": "No <i>2</i>",
         "correct": "Yes <i>2</i>"
       },
-      "zone": "Zone <b>2</b>",
+      "zones": [
+        "Zone <b>2</b>",
+        "Zone <i>1</i>"
+      ],
```
itsjeyd: @haikuginger We should probably switch to using `"zones"` instead of `"zone"` for the other items as well. The same comment applies to `tests/unit/data/plain/data.json`.

tests/unit/data/old/data.json
```diff
@@ -40,7 +40,10 @@
         "incorrect": "No 2",
         "correct": "Yes 2"
       },
-      "zone": "Zone 2",
+      "zones": [
+        "Zone 2",
+        "Zone 1"
+      ],
```
itsjeyd: @haikuginger I'm not sure if we should modify this file. There are no existing problems using the old data format that have multiple zones defined for a single item, but this change seems to suggest otherwise (so it's a little confusing).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -432,39 +481,37 @@ def _get_zone_by_uid(self, uid):
             if zone["uid"] == uid:
                 return zone
 
-    def _get_grade(self):
+    def _get_item_stats(self):
         """
-        Returns the student's grade for this block.
+        Returns a 3-tuple representing the number of correct items placed, the number of completed
+        items, right or wrong, and the total number of items that must be completed.
```
itsjeyd: @haikuginger This docstring is outdated now, could you please update it? To keep the terminology consistent I think it would be good to talk about "placed items" instead of "completed items" in the context of this method:

> Returns a tuple representing the number of correctly-placed items, and the total number of items that must be placed on the board (non-decoy items).

drag_and_drop_v2/drag_and_drop_v2.py
itsjeyd: @haikuginger `completed_items` ⟶ `placed_items`

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -82,45 +82,59 @@
...
 </script>
 
 <script id="item-input-tpl" type="text/html">
     <div class="item">
         <div class="row">
-            <label for="item-{{id}}-text">{{i18n "Text"}}</label>
+            <label for="item-{{id}}-text" class="h3">{{i18n "Text"}}</label>
             <input type="text"
                    id="item-{{id}}-text"
+                   placeholder="{{ i18n 'Use text that is clear and descriptive of the item to be placed'}}"
```
itsjeyd: @haikuginger Extra space before `i18n`

tests/unit/data/html/data.json
haikuginger: ✅

itsjeyd:
@haikuginger :+1:

- [x] I tested both locally and on the sandbox and everything works as advertised
- [x] I read through the code
- [x] Includes documentation

haikuginger:
Tagging for upstream review:

TNL: @cahrens and/or @staubina
a11y: @cptvitamin and/or @edx/edx-accessibility
Product: @sstack22
Also tagging @pdesjardins

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -82,45 +82,59 @@
...
-            <label for="item-{{id}}-image-url">{{i18n "Image URL (alternative to the text)"}}</label>
+            <fieldset>
+                <legend class="h3">
+                    {{ i18n "Zones" }}
+                </legend>
+                {{ checkboxes }}
+            </fieldset>
+        </div>
+        <div class="row">
+            <label for="item-{{id}}-image-url" class="h3">{{i18n "Image URL (alternative to the text)"}}</label>
```
cahrens: Nit: nest the input in the label so that the "for" attribute is not needed.

cahrens:
I'm having trouble finding the testing of the new functionality in the unit test changes.  Can you point me to where the test coverage is for an item that can be dropped in multiple zones?

Otherwise 👍

haikuginger:
@cahrens: The basic tests have been migrated to use `zones` rather than `zone`, and a couple of the items have multiple zones available, showing that we're properly looking for the zone name being contained in the list rather than being equal to the item. If we want an explicit test showing a single item being placed once, reset, and then placed again in a different zone, I can definitely add that.

cahrens:
@haikuginger I think that an explicit test like you described would be a great addition. Thanks!

haikuginger:
@cahrens: those items have been addressed. Passes all tests locally, so I'll babysit Travis in case it needs a bump.

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -116,12 +116,13 @@
...
         <div class="row">
-            <label for="item-{{id}}-image-url" class="h3">{{i18n "Image URL (alternative to the text)"}}</label>
-            <input type="text"
-                   id="item-{{id}}-image-url"
-                   placeholder="{{i18n 'For example, http://example.com/image.png or /static/image.png'}}"
-                   class="item-image-url"
-                   value="{{ imageURL }}" />
+            <label class="h3">{{i18n "Image URL (alternative to the text)"}}
+                <input type="text"
+                       id="item-{{id}}-image-url"
```
cahrens: Is the ID still needed?

cahrens:
Thanks, I don't need to re-review.

drag_and_drop_v2/templates/html/js_templates.html
haikuginger: No; I'll go ahead and pull that. Looks like it was only used here. Thanks!

sstack22:
@haikuginger - 👍 on this PR. thank you!

staubina:
👍 Tested and review the code, looks good to me.

clrux:
#### General

* Only Ctrl + M works on a PC (NVDA + JAWS). Pressing Enter/Return to select a button does not work. This might be intentional and correct, but the Help instructions may then be confusing. This isn't a problem on Mac's.
* When more than one button is placed in a dropzone, they overlap/stack. When the results are announced, only the last/top button in each dropzone is announced. It would be ideal if these buttons could somehow offset, showing all of the options in a dropzone.
* Using the up/down arrow keys to read in JAWS doesn't work. All I hear is "link".

##### Bucket Demo

* It is impossible to make an incorrect choice. Placing a word with two letters on the bucket with "Four" doesn't work. If this is graded, it seems like folks would always get 100% via process of elimination.

###### Voiceover

* Voiceover announces an incorrect placement as correct.
* When correctly placing a button, Voiceover doesn't announce the correct word/button. It's represented by a funky glyph in the textual output.
* I wasn't able to hear the description of the droppable area/image.

###### NVDA

* Incorrectly placed buttons aren't announced at all. Instead, the lengthy description of the droppable area and the currently placed buttons are announced.

###### JAWS

* Incorrectly placed buttons aren't announced at all.

##### Canada Demo

* Might just be the demo configuration, but the droppable regions hint to the answer ("BC Bottom leftmost province, button"). Not being Canadian I was able to figure out that BC meant British Columbia.

###### Voiceover

* Voiceover doesn't announce correctly placed buttons.
* I wasn't able to hear the description of the droppable area/image.

###### NVDA

* No indication of an incorrect or correct placement.

###### JAWS

* No indication of an incorrect or correct placement.

##### Multiple Blocks

###### Voiceover

* I wasn't able to hear the description of the droppable area/image.

###### NVDA

* Correct and incorrect dropped buttons aren't announced. And placing a button incorrectly announces itself as correctly placed.

###### JAWS

* I wasn't able to hear the description of the droppable area/image.
* Correct and incorrect dropped buttons aren't announced.

##### RGB Demo

###### Voiceover

* I wasn't able to hear the description of the droppable area/image.

###### NVDA

* Correct and incorrect dropped buttons aren't announced. And placing a button incorrectly announces itself as correctly placed.

###### JAWS

* I wasn't able to hear the description of the droppable area/image.
* Correct and incorrect dropped buttons aren't announced.

##### Backwards Compatibility

###### Voiceover

* I wasn't able to hear the description of the droppable area/image.

###### NVDA

* Correct and incorrect dropped buttons aren't announced. And placing a button incorrectly announces itself as correctly placed.

###### JAWS

* I wasn't able to hear the description of the droppable area/image.
* Correct and incorrect dropped buttons aren't announced.

##### Unconfigured

###### Voiceover

* I wasn't able to hear the description of the droppable area/image.

###### NVDA

* Correct and incorrect dropped buttons aren't announced. And placing a button incorrectly announces itself as correctly placed.

###### JAWS

* I wasn't able to hear the description of the droppable area/image.
* Correct and incorrect dropped buttons aren't announced.

##### Multiple Target Zones

* I can't interact with this demo. Maybe because all of the options have been placed?
* Correctly/Already placed buttons aren't announced in their dropzones.
* No description of the dropzone/image area in Voiceover.
* Pressing command + M on a Mac only minimizes the browser window.

clrux:
@haikuginger Sorry for the crazy feedback this morning. I neglected to scope the review to this PR (having too much fun playing around, I guess). Some things I discovered are covered under SOL-1807 so I will exclude them here.

Things work as expected. I am able to drop multiple items in a single dropzone. The only issue I came across was that the dropped items are stacked on top of each other. For sighted keyboard users, it is difficult to see all of the options in a dropzone as all but the top item is obscured. (And as you are aware from SOL-1807, placed/dropped items aren't announced back.)

Two possible solutions that came to mind:

* when dropped, use random coordinates from the absolutely positioned dropzone, or perhaps
* drop the items in a cascaded fashion; the results are tabbable and on focus and the z-index is changed to be higher than the others (reset on blur, or on another item focus)

haikuginger:
@clrux: I agree that it can definitely be a bit confusing. We do provide an "alignment" option for each zone; if that's enabled, then any dropped items will be moved to a predetermined (non-overlapping) position inside the zone. This isn't turned on by default, though, resulting in that behavior that you're seeing. I think the best route may be to add documentation suggesting that instructors building tests where multiple items go on a single target zone should enable the alignment feature for that zone, using the appropriate choice (left, center, or right).

bradenmacdonald:
@haikuginger As we did in another PR, can't we just force the "alignment" algorithm to control the placement of dropped items when they are dropped by keyboard controls, regardless of the alignment setting of that zone?  IMHO that solves all our problems and requires virtually no new code or author effort.

bradenmacdonald:
^ Turns out my suggestion above is tricky to implement.

@sstack22 We have this issue that for keyboard users, items stack on top of each other when placed. One simple solution would be to remove the feature where "items stay exactly where they are dropped" and instead only allow the "aligned" mode, where items get centered in the zone that they are dropped in:
screen shot 2016-07-28 at 1 33 43 pm

I changed the bucket demo at http://dndv2.opencraft.hosting/courses/course-v1:OpenCraft+DNDv2+DEMO/courseware/6915ee3dd6ab403d8c05c0ea3180a0ee/21c574a894d64b428d417a4a260304a5/ to show an example of this.

What do you think? Is the ability for items to stay where they are dropped important to any course authors?

(Whatever fix we make, this is an existing issue so I think can be outside the scope of this PR)

sstack22:
@bradenmacdonald - i think this is a great solution, and makes sense. I don't anticipate that course teams would have concerns.

clrux:
@haikuginger Is there a use-case for them overlapping as a default?

clrux:
[This solution](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/83#issuecomment-236017205) is very nice! What happens if there are more dropped items than the size of the dropzone?

bradenmacdonald:
@clrux I don't think there's any use case for them overlapping.

When there are more dropped items than the size of the dropzone, it's not the greatest:
screen shot 2016-07-29 at 11 59 59 am

However, that can only happen in "assessment mode". I'm going to open a new ticket for addressing the overlapping drops by turning on the automatic placement of items at all times, as I suggested.

Can I then consider you +1 on this PR? I think we're ready to merge.

clrux:
👍

PR #84, opened 2016-07-25
Move items between zones in assessment mode

mtyaka:
**Description**

This PR implements the following use case:

> As a learner, I want to be able to move items between drop targets before checking (submitting) my answer in assessment mode. This should be possible for any item, regardless of whether I placed it on a correct or incorrect zone before. I also want to be able to drag items back to the item bank without having to use the reset button.

*UX*

The existing implementation does not allow learners to move an item to a different zone once it has been placed on the board. We will keep this behavior for drag-and-drop problems in standard mode.

In assessment mode, learners should be able to move items that have already been placed on the board, regardless of whether they are located in a correct or incorrect zone. To do this, learners will be able to use the same workflows that are already established for moving items from the item bank to the board:

- Items can be moved to a different zone using the mouse. Workflow:
  1. Pick item up by clicking it.
  2. Drag item to desired zone.
  3. Drop item.
- Items can be moved to a different zone using the keyboard. Workflow:
  1. (Shift-)TAB to item.
  2. Pick item up by pressing "Enter", "Space", "Ctrl-m", or "⌘-m".
  3. TAB to desired zone.
  4. Drop item by pressing "Enter", "Space", "Ctrl-m", or "⌘-m".
  5. Press "Esc" to cancel drop operation.

With no item selected, learners will be able to (Shift-)TAB to items that have been placed on the board.

Once an item has been picked up, pressing (Shift-)TAB will only navigate between zones and the item bank.

**JIRA tickets**: [SOL-1943](https://openedx.atlassian.net/browse/SOL-1943)

**Discussions**: [Discovery document for DnDv2 assessment mode](https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit) (relevant section: 2.2.2)

**Dependencies**: None

**Sandbox URLs** (installed version de0312bfbba02b850c9032a6db63c728924505f5):

- LMS: http://dndv2-sandbox5.opencraft.hosting/
- Studio: http://studio-dndv2-sandbox5.opencraft.hosting/

DnD example is located at http://dndv2-sandbox5.opencraft.hosting/courses/course-v1:OpenCraftX+OC1+1/courseware/bdeaa08976fa4645a277edcaf5e8148c/e8ec2f5c98f4436abf532cccd6a71668/1?activate_block_id=block-v1%3AOpenCraftX%2BOC1%2B1%2Btype%40vertical%2Bblock%40e3c62b19090948b8ad89d3386bff4435

**Testing instructions**:

1. Add a DnDv2 block to a unit in Studio.
2. Click "EDIT" and change "Problem mode" to "assessment".
3. Drag items from the bank into drop zones. You may place some items in wrong zones.
4. Drag and drop already placed items between zones, both with mouse and keyboard.
5. Observe that the items "stick" in the placed position (are not moved back to the bank or previous zone). Observe that per-item feedback is not displayed.
6. Drag some items back to the bank with both mouse and keyboard.
6. Change "Problem mode" back to "standard".
7. Make sure everything still works correctly in standard mode.

If you put multiple items into the same zone with keyboard in assessment mode, you can only see the topmost item, the rest of the items are hidden behind it (the same thing can happen when dragging with the mouse, but when dragging with the mouse you usually don't drop the item in *exactly* the same place, so parts of overlapping items are usually still visible).

Using the existing zone alignment feature (where you can specify how dropped items automatically align inside zones) prevents this problem.

**Reviewers**
- [x] OpenCraft/Solutions: @haikuginger  
- [ ] a11y: @edx/edx-accessibility
- [ ] Product: @sstack22

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -30,6 +30,29 @@ function DragAndDropTemplates(configuration) {
         }
     };
 
+    var bankItemWidthStyles = function(item, ctx) {
+        var style = {};
+        if (item.widthPercent) {
+            // The item bank container is often wider than the background image, and the
+            // widthPercent is specified relative to the background image so we have to
+            // convert it to pixels. But if the browser window / mobile screen is not as
+            // wide as the image, then the background image will be scaled down and this
+            // pixel value would be too large, so we also specify it as a max-width
+            // percentage.
+            style.width = (item.widthPercent / 100 * ctx.bg_image_width) + "px";
+            style.maxWidth = item.widthPercent + "%";
+        }
+        return style;
+    };
+
+    var itemContentTemplate = function(item) {
+        var item_content_html = item.displayName;
+        if (item.imageURL) {
+            item_content_html = '<img src="' + item.imageURL + '" alt="' + item.imageDescription + '" />';
+        }
+        return h('div', { innerHTML: item_content_html, className: "item-content" });
+    };
+
     var itemTemplate = function(item, ctx) {
         // Define properties
         var className = (item.class_name) ? item.class_name : "";
@@ -40,12 +63,15 @@ function DragAndDropTemplates(configuration) {
         if (item.widthPercent) {
             className += " specified-width";  // The author has specified a width for this item.
         }
+        if (item.grabbed_with) {
+            className += " grabbed-with-" + item.grabbed_with;
+        }
         var attributes = {
             'role': 'button',
             'draggable': !item.drag_disabled,
             'aria-grabbed': item.grabbed,
             'data-value': item.value,
-            'data-drag-disabled': item.drag_disabled
+            'tabindex': item.focusable ? 0 : undefined
         };
         var style = {};
         if (item.background_color) {
@@ -83,28 +109,13 @@ function DragAndDropTemplates(configuration) {
                 // ^ Hack to detect image width at runtime and make webkit consistent with Firefox
             }
         } else {
-            // If an item has not been placed it must be possible to move focus to it using the keyboard:
-            attributes.tabindex = 0;
-            if (item.widthPercent) {
-                // The item bank container is often wider than the background image, and the
-                // widthPercent is specified relative to the background image so we have to
-                // convert it to pixels. But if the browser window / mobile screen is not as
-                // wide as the image, then the background image will be scaled down and this
-                // pixel value would be too large, so we also specify it as a max-width
-                // percentage.
-                style.width = (item.widthPercent / 100 * ctx.bg_image_width) + "px";
-                style.maxWidth = item.widthPercent + "%";
-            }
+            $.extend(style, bankItemWidthStyles(item, ctx));
         }
         // Define children
         var children = [
             itemSpinnerTemplate(item.xhr_active)
         ];
-        var item_content_html = item.displayName;
-        if (item.imageURL) {
-            item_content_html = '<img src="' + item.imageURL + '" alt="' + item.imageDescription + '" />';
-        }
-        var item_content = h('div', { innerHTML: item_content_html, className: "item-content" });
+        var item_content = itemContentTemplate(item);
         if (item.is_placed) {
             // Insert information about zone in which this item has been placed
             var item_description_id = configuration.url_name + '-item-' + item.value + '-description';
@@ -143,6 +154,35 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
+    // When an item is dragged out of the bank, a hidden placeholder of the same width and height as
+    // the original item is rendered in the bank. The function of the placeholder is to take up the
+    // same amount of space as the original item so that the bank does not collapse when you've dragged
+    // all items out.
+    var itemPlaceholderTemplate = function(item, ctx) {
+        var className = "";
+        if (item.has_image) {
+            className += " " + "option-with-image";
+        }
+        if (item.widthPercent) {
+            className += " specified-width";  // The author has specified a width for this item.
+        }
+        var style = bankItemWidthStyles(item, ctx);
+        // Placeholder should never be visible.
+        style.visibility = 'hidden';
+        return (
+            h(
+                'div.option',
+                {
+                    key: 'placeholder-' + item.value,
+                    className: className,
+                    attributes: {draggable: false},
+                    style: style
+                },
+                itemContentTemplate(item)
+            )
+        );
+    };
+
     var zoneTemplate = function(zone, ctx) {
         var className = ctx.display_zone_labels ? 'zone-name' : 'zone-name sr';
         var selector = ctx.display_zone_borders ? 'div.zone.zone-with-borders' : 'div.zone';
@@ -161,6 +201,7 @@ function DragAndDropTemplates(configuration) {
             h(
                 selector,
                 {
+                    key: zone.prefixed_uid,
                     id: zone.prefixed_uid,
                     attributes: {
                         'tabindex': 0,
@@ -246,6 +287,15 @@ function DragAndDropTemplates(configuration) {
         var items_in_bank = $.grep(ctx.items, is_item_placed, true);
         var is_item_placed_unaligned = function(i) { return i.zone_align === 'none'; };
         var items_placed_unaligned = $.grep(items_placed, is_item_placed_unaligned);
+        var item_bank_properties = {};
+        if (ctx.item_bank_focusable) {
+            item_bank_properties.attributes = {
+                'tabindex': 0,
+                'dropzone': 'move',
+                'aria-dropeffect': 'move',
+                'role': 'button'
+            };
+        }
         return (
             h('section.themed-xblock.xblock--drag-and-drop', [
                 problemTitle,
@@ -254,10 +304,11 @@ function DragAndDropTemplates(configuration) {
                     h('p', {innerHTML: ctx.problem_html}),
                 ]),
                 h('section.drag-container', {}, [
-                    h(
-                        'div.item-bank',
-                        renderCollection(itemTemplate, items_in_bank, ctx)
-                    ),
+                    h('div.item-bank', item_bank_properties, [
+                        h('p', { className: 'zone-description sr' }, gettext('Item Bank')),
+                        renderCollection(itemTemplate, items_in_bank, ctx),
+                        renderCollection(itemPlaceholderTemplate, items_placed, ctx)
+                    ]),
                     h('div.target',
                         {
                             attributes: {
@@ -281,10 +332,9 @@ function DragAndDropTemplates(configuration) {
                                 h('img.target-img', {src: ctx.target_img_src, alt: ctx.target_img_description}),
                             ]
                         ),
-                        renderCollection(zoneTemplate, ctx.zones, ctx),
-                        renderCollection(itemTemplate, items_placed_unaligned, ctx)
-                    ]
-                    ),
+                        renderCollection(itemTemplate, items_placed_unaligned, ctx),
+                        renderCollection(zoneTemplate, ctx.zones, ctx)
+                    ]),
                 ]),
                 keyboardHelpTemplate(ctx),
                 feedbackTemplate(ctx),
@@ -327,7 +377,6 @@ function DragAndDropBlock(runtime, element, configuration) {
     var TAB = 9;
     var M = 77;
 
-    var placementMode = false;
     var $selectedItem;
     var $focusedElement;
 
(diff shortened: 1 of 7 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=84

tests/integration/test_interaction.py
```diff
@@ -406,27 +438,33 @@ class AssessmentInteractionTest(DefaultAssessmentDataTestMixin, InteractionTestB
...
+    #     self.parameterized_final_feedback_and_reset(
+    #         self.items_map, self.feedback, action_key=action_key, assessment_mode=True
+    #     )
 
-    @data(False, True)
-    def test_keyboard_help(self, use_keyboard):
-        self.interact_with_keyboard_help(use_keyboard=use_keyboard)
+    # @data(False, True)
+    # def test_keyboard_help(self, use_keyboard):
+    #     self.interact_with_keyboard_help(use_keyboard=use_keyboard)
```
haikuginger: Is there a reason we've got several tests commented out here?

mtyaka: @haikuginger There isn't, thanks for catching this. I commented them out while debugging a specific test and then forgot to uncomment them :/.

haikuginger:
So, I've checked this out - all tests pass locally, and the interaction definitely works as intended. I do have a small interaction concern, though - there's no way to remove an item from the board once it's been placed, besides pressing the reset button. In general, this isn't a huge issue, except when it comes to decoy items.

Right now, when an item is dropped on an area that isn't an active zone, it jumps back to where it was placed before. I kind of wonder if they should jump back to the palette instead. Either that, or we should possibly make the palette a full zone that we can drag to in its own right.

mtyaka:
@haikuginger Thanks a lot for the review. That's a very good point. I think it would make sense to be able to drag items back to the bank.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -722,16 +736,23 @@ function DragAndDropBlock(runtime, element, configuration) {
         updateDOM();
     };
 
-    var setGrabbedState = function(item_id, grabbed) {
-        for (var i = 0; i < configuration.items.length; i++) {
-            if (configuration.items[i].id === item_id) {
-                configuration.items[i].grabbed = grabbed;
+    var setGrabbedState = function(item_id, grabbed, interaction_type) {
+        for (var i = 0, item; i < configuration.items.length; i++) {
+            item = configuration.items[i];
```
bradenmacdonald: Nit: Just use `configuration.items.forEach(function(item) {` which is nicer and used elsewhere in the code.

bradenmacdonald:
Yes, one should be able to drag items back to the bank.

> If you put multiple items into the same zone with keyboard in assessment mode, you can only see the topmost item, the rest of the items are hidden behind it 

Can we use the existing "alignment" code to avoid this? It already has an algorithm for moving items around within a zone so that they don't overlap, if I remember correctly.

mtyaka:
> Can we use the existing "alignment" code to avoid this? It already has an algorithm for moving items around within a zone so that they don't overlap, if I remember correctly.

I was not familiar with this feature, but it seems to be working well and prevents the problem.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: Thanks, changed.

mtyaka:
@haikuginger I implemented ability to drag items back to the bank, please take a look.

haikuginger:
:+1:

- [x] I tested this: I ran automated tests locally and confirmed all tests passed. I also installed the XBlock in a devstack and manually tested the interaction. Buttons can now be dragged between targets in Assessment Mode, and they can also be dragged back to the item bank.
- [x] I read through the code
- [x] Includes documentation

mtyaka:
Rebased cd7e92e43cb82c77678c8931df06b86eda1a8701 onto master

cahrens:
When I used the keyboard to move items, I ended up with "Goes to the top" completely behind "Goes to the middle". Probably keyboard moves always go to the same place?

image

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -143,6 +154,35 @@ function DragAndDropTemplates(configuration) {
...
+        var className = "";
+        if (item.has_image) {
+            className += " " + "option-with-image";
+        }
+        if (item.widthPercent) {
+            className += " specified-width";  // The author has specified a width for this item.
+        }
+        var style = bankItemWidthStyles(item, ctx);
+        // Placeholder should never be visible.
+        style.visibility = 'hidden';
```
cahrens: They will be invisible to screen readers as well?

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
```diff
@@ -409,6 +409,10 @@ msgid "ok"
 msgstr ""
 
 #: public/js/drag_and_drop.js
+msgid "Item Bank"
+msgstr ""
+
+#: public/js/drag_and_drop.js
 msgid "Placed in: "
```
cahrens: Shouldn't the placeholder be part of the string for translation?

Placed in: {}

tests/integration/test_interaction.py
```diff
@@ -110,56 +131,69 @@ def place_item(self, item_value, zone_id, action_key=None):
             self.move_item_to_zone(item_value, zone_id, action_key)
 
     def drag_item_to_zone(self, item_value, zone_id):
-        element = self._get_unplaced_item_by_value(item_value)
```
cahrens: Add documentation that zone_id of None signifies the item bank.

tests/integration/test_interaction.py
```diff
@@ -110,56 +131,69 @@ def place_item(self, item_value, zone_id, action_key=None):
...
+        element = self._get_item_by_value(item_value)
+        if zone_id is None:
+            target = self._get_item_bank()
+        else:
+            target = self._get_zone_by_id(zone_id)
         action_chains = ActionChains(self.browser)
         action_chains.drag_and_drop(element, target).perform()
 
     def move_item_to_zone(self, item_value, zone_id, action_key):
         # Get zone position
```
cahrens: Again, a note about zone_id being None would be helpful.

tests/integration/test_interaction.py
```diff
@@ -110,56 +131,69 @@ def place_item(self, item_value, zone_id, action_key=None):
...
-        zone_position = self._get_zone_position(zone_id)
+        if zone_id is None:
+            zone_position = len(self.all_zones)
```
cahrens: How does this work for the zone_position?

tests/integration/test_interaction.py
```diff
@@ -424,6 +472,12 @@ def test_item_no_feedback_on_bad_move(self, action_key):
         )
 
     @data(None, Keys.RETURN, Keys.SPACE, Keys.CONTROL+'m', Keys.COMMAND+'m')
+    def test_move_items_between_zones(self, action_key):
+        self.parameterized_move_items_between_zones(
```
cahrens: Is there test coverage that items can't be moved once dropped for assessment_mode=False?

cahrens:
Just a few questions (see above), and I noted a situation in which an item can get completely hidden behind another element.

bradenmacdonald:
@cahrens The problem with overlapping drop targets was a pre-existing issue AFAIK, and after it was brought up on another PR I created https://openedx.atlassian.net/browse/SOL-1979 - so we will deal with it in an upcoming PR.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: @cahrens Yes, elements with `visibility: hidden` are ignored by screen readers.

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
mtyaka: @cahrens It should, thanks for catching that. It's not related to the changes in this PR, but I'll go ahead and update it now anyway.

tests/integration/test_interaction.py
mtyaka: Makes sense, I will add a docstring to explain the meaning of `zone_id=None`.

tests/integration/test_interaction.py
mtyaka: Good point, will do.

tests/integration/test_interaction.py
mtyaka: @cahrens Yeah, this code is quite confusing.

In keyboard placement mode pressing the Tab key cycles focus between zones and the item bank, with the item bank being the last one to gain focus.

`zone_position` is not really the best variable name for what we are doing here, I will rename it to `tab_press_count` and add some comments to explain what the code is doing.

tests/integration/test_interaction.py
mtyaka: We were only testing that indirectly (by asserting dropped item has no `tabindex`, has a `fade` class, etc).
I added some tests to explicitly test that items cannot be moved between zones or back to the bank in standard mode.

mtyaka:
@cahrens The item overlap issue will be addressed by https://openedx.atlassian.net/browse/SOL-1979. I addressed all of your other comments, please have a look.

staubina:
My only concern is that focus goes to the bottom of the page when selecting a drop zone before selecting something to drag.  Is this expected behavior? Could we instead focus on the problem again?

Other than that 👍

mtyaka:
@staubina Thanks for the review! I believe that's the expected behavior. It has worked that way before the changes in this PR.

cahrens:
👍 , though I would think that the scrolling behavior @staubina mentioned will have to be fixed at some point. If you don't think it makes sense to do as part of this PR, can you update an appropriate ticket with this information?

Note that the keyboard focus is correct (it's still on the zone you selected), just the spurious scroll occurs.

mtyaka:
@staubina @cahrens Sorry, I think we might have been thinking about different issues. What kind of scrolling behavior are you seeing?

Is it when you press the space bar while the focus is on a zone without any item being selected?

cahrens:
@mtyaka That's what I was talking about. @staubina were you talking about the same thing?

staubina:
@cahrens @mtyaka Yes that is the issue I was referring to.

mtyaka:
@cahrens @staubina I see. Scrolling the page down when you press space bar is the default behavior in most browsers. When you pick up an item we hijack the space bar from the browser so that it drops the item rather than scroll the page.

When no item is picked up, space bar is not hijacked so pressing it produces the standard scroll down effect. I can see why this can be distracting while working on a Drag and Drop exercise, especially if you think you've grabbed an item when you haven't, so I think it would make sense to hijack the space bar every time one of the zones is focused, regardless of whether an item has been picked up or not.

I will check to see whether that's easy to do. If easy, I'll add it to this PR, if not I'll create a separate ticket.

bradenmacdonald:
@mtyaka I think it's simply a matter of calling `preventDefault()` or similar in the event handler, to stop the spacebar event from bubbling up to the body and causing the page scroll.

mtyaka:
Thanks for the suggestion @bradenmacdonald, `preventDefault()` did it.

@cahrens @staubina I added some code to prevent default browser scroll from happening when pressing space bar while focused on a zone with no item selected.

mtyaka:
@sstack22 Please review when you get a chance.

staubina:
@mtyaka Thanks for making that change.  Did you update the Sandbox?

:+1:

mtyaka:
@staubina The sandbox hasn't been updated yet, I'll do it in a couple of minutes.

mtyaka:
@staubina Sandbox has been updated.

staubina:
Looks good to me
👍

sstack22:
@mtyaka  - a couple small notes:

1) When you move an item after having placed it, the text disappears and there is only a loading sign. Is it possible to have the text appear as well (similar to the initial drop?)
2) I was able to drop an item on the left side background. If I try to drop an item on the right side, the item is automatically placed in the related zone. I would have expected the same behavior for the left side background.

screen shot 2016-08-03 at 11 07 00 am

Otherwise, looks good.

mtyaka:
Thanks for the review @sstack22!

> 1) When you move an item after having placed it, the text disappears and there is only a loading sign. Is it possible to have the text appear as well (similar to the initial drop?)

Thanks for pointing that out, I never noticed the way text temporarily disappeared. It is now fixed and the sandbox has been updated with the fix.

> 2) I was able to drop an item on the left side background. If I try to drop an item on the right side, the item is automatically placed in the related zone. I would have expected the same behavior for the left side background.

The way we currently determine whether the item was dropped into a zone is by looking at the position of mouse pointer at the time you release the item. If the mouse pointer is inside a zone, the drop is successful, but if the mouse pointer is outside the zone, the drop is reverted.
So when you dropped the item to the left edge, you were probably holding it somewhere on the right side, so that when you released it, mouse pointer was still inside the zone. It is easier to understand the behavior when the zone is highlighted:

screen shot 2016-08-03 at 23 42 55

sstack22:
@mtyaka - thanks for the clarification! I confirmed the fix for #1. This looks great, 👍

mtyaka:
Squashed from 30173c470b7a758ce4b96ec3261bc8ee7cc8b7b0

PR #73, opened 2016-06-23
Adding usability for keyboard controls (SOL-1807)

dovinmu:
Edit: this pull request is covered by the OpenCraft contributor agreement. @antoviaque 

This should add in the following functionality:

1. 'When tabbing through the focusable controls, the Windows user is never exposed to the alt text for the background image, which is critical to understanding the context of the problem. I think a reasonable solution would be to add an ID to this element and then reference it via aria-describedby on all the dropzones. The screen reader user would move focus to the dropzone, hear its description, and then hear the description of its meaning in the greater context of the problem.'

2. 'In certain screen readers, the keyboard controls are not exposed. On Windows and NVDA< only the CTRL + M keycombo works to drag and drop the items (many users will not be aware of this combination). It would be very cool to add an additional aria-describedby that references that bit of help text that is available in the Keyboard help modal: "Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on." on the draggable items and "Press "Enter", "Space", "Ctrl-m", or "⌘-m" to drop the item on the current zone." on the dropzones.'

3. 'when I went to a dropzone that had an item in it, it was not read back to me, so I had no idea if a dropzone had any items in it or not.'

**Sandbox URLs**

Studio: http://studio-dndv2-sandbox7.opencraft.hosting/
LMS: http://dndv2-sandbox7.opencraft.hosting/

The sandbox is at version de47111bb93a892778fa32fb35d5d6ee851be293

PR #85, opened 2016-07-26
[SOL-1806] Moved aria-live attributes to popup wrapper

e-kolpakov:
**Description:** This PR modifies ARIA attributes so that only feedback is read to screen-reader user when an item is dropped into a zone, instead of full "working field"

**JIRA:** [SOL-1806](https://openedx.atlassian.net/browse/SOL-1806)

**Testing instructions:**
0. Enable Screen Reader + make sure your browser support it (e.g. default Ubuntu screen reader does not work with Chrome, but works with Firefox).
0.a. Get used to it speaking out loud everything + learn to notice important parts of it.
1. Create Drag and Drop v2 XBlock in a course. Default configuration should be enough.
2. Open unit with DnDv2 in LMS.
3. Use keyboard or mouse to drop items into correct zones. *Expected behavior:* only feedback is read out loud. *Previous behavior:* almost everything were read out loud (all/some zones and items, feedback)
4. (Unconfirmed - see concerns below) Use keyboard or mouse to drop items into *incorrect* zones. *Expected behavior:* only feedback is read out loud. *Previous behavior:* almost everything were read out loud (all/some zones and items, feedback)

**Sandboxes:** 
LMS: http://dndv2-sandbox3.opencraft.hosting/
Studio: http://studio-dndv2-sandbox3.opencraft.hosting/

Sandboxes use 07ce4db8ba068050d771338b3d10d341a20b2b70 version

**Reviewers**

- [x] @mtyaka 
- [ ] a11y: @cptvitamin
- [ ] Product: @sstack22

**Author concerns:**

Default Ubuntu screen reader seems to ignore error feedback popups (despite them being absolutely identical to normal popups). This behavior exists in current master, so I'm not sure if it is a real issue; and if so, should it be fixed in scope of this PR.

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -227,7 +227,6 @@ function DragAndDropTemplates(configuration) {
     };
 
     var feedbackTemplate = function(ctx) {
-        var properties = { attributes: { 'aria-live': 'polite' } };
         var messages = ctx.overall_feedback_messages || [];
         var feedback_display = messages.length > 0 ? 'block' : 'none';
         var feedback_messages = messages.map(function(message) {
@@ -239,11 +238,49 @@ function DragAndDropTemplates(configuration) {
         });
 
         return (
-            h('section.feedback', properties, [
-                h('h3.title1', { style: { display: feedback_display } }, gettext('Feedback')),
-                h('div.messages', { style: { display: feedback_display } }, feedback_messages)
+            h('section.feedback', {}, [
+                h(
+                    "div.feedback-content",
+                    { attributes: { 'aria-live': 'polite' } },
+                    [
+                        h('h3.title1', { style: { display: feedback_display } }, gettext('Feedback')),
+                        h('div.messages', { style: { display: feedback_display } }, feedback_messages),
+                    ]
+                )
             ])
         );
+    };
+
+    var popupTemplate = function(ctx) {
+        var popupSelector = 'div.popup';
+        if (ctx.popup_html && !ctx.last_action_correct) {
+            popupSelector += '.popup-incorrect';
+        }
+
+        return (
+            h(
+                "div.popup-wrapper",
+                {
+                    attributes: {
+                        'aria-live': 'polite',
+                        'aria-atomic': 'true',
+                        'aria-relevant': 'additions',
+                    }
+                },
+                [
+                    h(
+                        popupSelector,
+                        {
+                            style: {display: ctx.popup_html ? 'block' : 'none'},
+                        },
+                        [
+                            h('div.close.icon-remove-sign.fa-times-circle'),
+                            h('p.popup-content', {innerHTML: ctx.popup_html}),
+                        ]
+                    )
+                ]
+            )
+        )
     };
 
     var keyboardHelpPopupTemplate = function(ctx) {
@@ -324,10 +361,7 @@ function DragAndDropTemplates(configuration) {
     var mainTemplate = function(ctx) {
         var problemTitle = ctx.show_title ? h('h2.problem-title', {innerHTML: ctx.title_html}) : null;
         var problemHeader = ctx.show_problem_header ? h('h3.title1', gettext('Problem')) : null;
-        var popupSelector = 'div.popup';
-        if (ctx.popup_html && !ctx.last_action_correct) {
-            popupSelector += '.popup-incorrect';
-        }
+
         // Render only items_in_bank and items_placed_unaligned here;
         // items placed in aligned zones will be rendered by zoneTemplate.
         var is_item_placed = function(i) { return i.is_placed; };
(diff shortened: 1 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=85&repo=edx-solutions%2Fxblock-drag-and-drop-v2

sstack22:
@e-kolpakov - FYI the sandbox isn't loading for me

bradenmacdonald:
@e-kolpakov FYI, when you list the "reviewers" in the PR description, it pings all of them - so probably better to fill that section out once our internal review is completed, and it's ready for review?

cahrens:
Yes, I would appreciate following @bradenmacdonald's suggestion. It would make it clearer when it is ready for TNL review.

mtyaka:
@e-kolpakov I've tested this with VoiceOver in Safari on OS X. I noticed a similar (the same?) problem where incorrect popups are not read out, but that only happens while the board is empty - after you place at least one item in a correct zone, the incorrect popups start working as expected.

I found that the problem was the "Reset problem" button. When you drop the first item onto the board, it shows for a fraction of a second until the item is reverted back to the bank and the button hidden. The button seem to distract VoiceOver which starts reading out "Reset problem" instead of the new contents of the popup.
I replaced `display_reset_button: Object.keys(state.items).length > 0` with this code so that the button only shows when at least one item is in *correct* location:

```
      var display_reset_button = false;
      for (var key in state.items) {
        if (state.items.hasOwnProperty(key)) {
          if (!state.items[key].submitting_location) {
            display_reset_button = true;
            break;
          }
        }
      }
```

After I made this change incorrect popups started working fine, but now after placing the first correct item, VoiceOver would start reading out "Reset problem" instead of the contents of the correct popup. The reset button does have `aria-live="off"`, but VoiceOver seems to ignore that for some reason. To fix this problem, I removed `aria-live="polite"` attribute from the `section.feedback` wrapper which includes the reset button, the feedback title, and the feedback content, and instead wrapped the feedback title and feedback content in an extra div with `aria-live="polite"`.

After this change, all popups work correctly for me, except in the case where you drag the same item to two wrong locations successively (the first time you drop it into the wrong location, VoiceOver reads out "The item does not belong here", but after you drag it again and drop it on another (wrong) region, the feedback message doesn't change, so VoiceOver doesn't read it out). This would be hard to fix and since this problem was already present prior to this PR, I think we can ignore it for now.

e-kolpakov:
@mtyaka Implemented the suggested fixes; however, my SR still does not read "error" feedback. Instead it reads zone description twice. Probably a bug in Ubuntu screen reader + Firefox

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -250,24 +284,9 @@ function DragNDropTemplates(url_name) {
...
-                                popupSelector,
-                                {
-                                    style: {display: ctx.popup_html ? 'block' : 'none'},
-                                },
-                                [
-                                    h('div.close.icon-remove-sign.fa-times-circle'),
-                                    h('p.popup-content', {innerHTML: ctx.popup_html}),
-                                ]
-                            ),
+                            popupTemplate(ctx),
```
mtyaka: @e-kolpakov Nice, this looks much cleaner now.

mtyaka:
:+1:

- [x] I tested this with VoiceOver on Safari
- [x] I read through the code
- [x] this change does not require documentation

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -179,20 +179,57 @@ function DragNDropTemplates(url_name) {
...
         var feedback_display = ctx.feedback_html ? 'block' : 'none';
         var reset_button_display = ctx.display_reset_button ? 'block' : 'none';
-        var properties = { attributes: { 'aria-live': 'polite' } };
         return (
-            h('section.feedback', properties, [
+            h('section.feedback', {}, [
                 h(
                     'button.reset-button.unbutton.link-button',
-                    { style: { display: reset_button_display }, attributes: { tabindex: 0 }, 'aria-live': 'off'},
+                    { style: { display: reset_button_display }, attributes: { tabindex: 0, 'aria-live': 'off' } },
```
mtyaka: Nice catch. I tested whether moving `arial-live="polite"` back to the `section.feedback` container makes any difference with this change, but it looks like it doesn't. VoiceOver still reads out "Reset problem" the first time it appears, even though `aria-live="off"` is set correctly, so moving the button out of `aria-live="polite"` aria as you did here still makes sense.

staubina:
The code changes seem straight forward, but I am having trouble recreating this on the Sandbox. Can you set up the Drag and drop problem you used for testing?

e-kolpakov:
@staubina I've updated the course with the default instance of DnDv2. Note there's a stock "Drag and Drop" CAPA problem - it is not the one we're working with here.

cahrens:
I've read through the code and will defer to a11y review on this one.

staubina:
I again tried to parse through this with a screen reader, but having not used it often, I am going to also defer to accessibility on this PR.

e-kolpakov:
@cptvitamin Friendly reminder to review this PR.

e-kolpakov:
@sstack22 FYI - waits for your and a11y review. This PR is about a11y as well, so I believe we should not merge it withot a11y review. Might be a good idea to include reviewing this PR and #73 to upcoming UX/RTL/a11y task.

bradenmacdonald:
@e-kolpakov  Per @cptvitamin, this can merge without his review for now; I'm waiting to find out if @sstack22 wants to review.

Please rebase and get a green build in preparation for merge.

bradenmacdonald:
Rebased from 71796ad1032f43a0b389aae4ceac0bb65c0acc00 and squashed to 1cb7b9d6673eb552c25718f0daa8cc81d8917a22 . There were some rebase conflicts, so I did a quick post-rebase manual test of both standard + assessment mode using VoiceOver, and everything seems good. Merging now.

PR #82, opened 2016-07-20
Keep items in dropped position in assessment mode.

mtyaka:
**Description**

This PR implements the following use case:

> As a learner, I want to place all items belonging to a drag-and-drop problem in assessment mode on the board first, and then check (submit) my answer by clicking a button.

Items dropped onto wrong zones in standard mode are immediately removed from the board and sent back to the bank.

In assessment mode, items should stick in the dropped position until the entire problem is explicitly submitted by clicking a button.

Note that the submit button and associated xhr handler will be implemented in a future PR.

*UX*

The workflow for placing items on the board should not change in assessment mode. However, items that were placed on an incorrect zone should stay on the board in assessment mode (at least until the learner checks their answer) instead of returning to the item bank.

Also, a drag-and-drop problem in assessment mode should not show per-item "Success Feedback" or "Error Feedback" to the learner as they place items on the board.

**JIRA tickets**: [SOL-1942](https://openedx.atlassian.net/browse/SOL-1942)

**Discussions**: [Discovery document for DnDv2 assessment mode](https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit) (relevant section: 2.2.1)

**Dependencies**: None

**Screenshots**:

Items stick in dropped position even when dropped onto a wrong zone:

screen shot 2016-07-20 at 19 40 22

**Sandbox URLs** (ff7ddfd2ee734594459f5e96c18e453b8cc3ca92): 
- [Studio](http://studio-dndv2.opencraft.hosting/container/block-v1:OpenCraft+DNDv2+DEMO+type@vertical+block@90948135d1fb45afbb87622eb07f8f40)
- [LMS](http://dndv2.opencraft.hosting/courses/course-v1:OpenCraft+DNDv2+DEMO/courseware/6915ee3dd6ab403d8c05c0ea3180a0ee/8c5328b8b478441bac5d588bf76f0da2/1?activate_block_id=block-v1%3AOpenCraft%2BDNDv2%2BDEMO%2Btype%40vertical%2Bblock%4090948135d1fb45afbb87622eb07f8f40)

**Testing instructions**:

1. Add a DnDv2 block to a unit in Studio.
2. Click "EDIT" and change "Problem mode" to "assessment".
3. Drag items from the bank into drop zones. Place some items in wrong zones.
4. Observe that the items "stick" in the placed position (are not moved back to the bank). Observe that per-item feedbacks are not displayed.
[...]

PR #81, opened 2016-07-19
Add option for limiting number of attempts in assessment mode to Studio editor

itsjeyd:
**Description**

This PR implements the following use case<sup>1</sup>:

> As a course author, I want to be able to limit the number of times a learner may attempt a drag-and-drop problem in assessment mode. More specifically, I want to be able to specify that a learner may try a problem one time, n times, or as many times as they wish (unlimited attempts).

<sup>1</sup>Note that learner-facing functionality related to limiting the number of available attempts will be covered by future PRs.

*UX*

For drag-and-drop problems in standard mode, the interactive Studio editor does not show an option for setting the maximum number of attempts. 

For drag-and-drop problems in assessment mode, the interactive Studio editor shows a numeric input field that allows course authors to specify the maximum number of attempts. The field is hidden until a course author sets the mode of a drag-and-drop problem to "assessment mode" (as described above).

* By leaving the maximum number of attempts unset, course authors can indicate that the number of times learners may attempt the problem is unlimited. 
* By setting the maximum number of attempts to a positive number `n`, course authors can indicate that learners may attempt the problem exactly `n` times. 

The value of the “Maximum attempts” setting defaults to “not set” (unlimited attempts).

Switching modes does not affect the current value of the “Maximum attempts” field.

**JIRA tickets**: [SOL-1941](https://openedx.atlassian.net/browse/SOL-1941)

**Discussions**: The feature implemented in this PR was discussed with @marcotuts as part of his review of the [discovery document](https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit) for DnDv2 assessment mode.

* Relevant section: 2.1.2
* Relevant comment threads (have since been "marked as resolved"):
  * https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit?disco=AAAAAsmOs3E
  * https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit?disco=AAAAAsmOs3I

**Dependencies**: None

**Screenshots**:

Standard mode:

edit-view-standard

Assessment mode:

edit-view

**Sandbox URL**:

Studio: http://studio-dndv2-sandbox2.opencraft.hosting/
LMS: http://dndv2-sandbox2.opencraft.hosting/

[...]

PR #86, opened 2016-07-27
[SOL-1944] Submit Answer UI

e-kolpakov:
**Description:** This PR introduces concept of attempts to DnDv2 and adds UI elements to submit an answer and display number of attempts.

**JIRA:** Part of [SOL-1944](https://openedx.atlassian.net/browse/SOL-1944)

**Discussions**: [Discovery document for DnDv2 assessment mode](https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit) (relevant section: 2.2.3) - covers only UI part of it, click handlers and backend processing will be implemented as a separate PR.

**Sandbox URLs**:

- LMS: http://dndv2-sandbox6.opencraft.hosting/
- Studio: http://studio-dndv2-sandbox6.opencraft.hosting/

Currently pinned to 5754ac6

**Screenshots:**

1. Student view - can't submit as no items are placed

image

2. Student view - can submit (+ hover over reset):

image

Note "Reset problem" and "Keyboard help" are reworked to look more like the upcoming CAPA Problem redesign and edx-pattern-library in general.

**Testing instructions:**
1. Create/edit DnDv2; put it into assessment mode
2. Set "Max attempts" to zero.
3. Publish, open in LMS.
4. Note that "Submit" button is displayed. Make sure problem is in its initial state (reset if needed). Submit button should be disabled. Attempts info should not be displayed.
5. Put any item to correct zone. "Submit" button should become enabled.
6. Return to Studio, set "Max attempts" to any positive integer. Publish.
7. Return to LMS. Note attempts info is displayed. Should read "You have used 0 of <number_from_step_6> attempts."

*Not* implemented as part of this PR - expected behavior:
1. "Submit" button do nothing.
2. Used attempts are always zero.

**Author concerns:**

This PR copies some classes from edx-pattern-library into DnDv2 own css. Other options considered seems like a greater evil:

* edx-pattern-library is only available in `lms-main-v2.css` in edX-platform, and that file is not added to courseware page
* Including edx-pattern-library into DnDv2 is feasible, but problematic: needs converting css into sass/scss, installing edx-pattern-library as node module, build system or script and finally committing both source sass/scss and generated files. All in all - too much trouble (and definitely too much for this "feature" PR)

On the other hand, classes copied are more or less colocated, so when edx-pattern-library classes become available on courseware pages it should be sufficient to just drop their copied counterparts from DnDv2 css.

**Reviewers**
- [x] @mtyaka 
- [ ] a11y: @cptvitamin
- [x] Product: @sstack22

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -127,6 +127,12 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         default={},
     )
 
+    num_attempts = Integer(
+        help=_("Number of attempts learner used"),
+        scope=Scope.user_state,
+        default=0
+    )
+
     completed = Boolean(
         help=_("Indicates whether a learner has completed the problem at least once"),
         scope=Scope.user_state,
@@ -191,6 +197,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
         return {
             "mode": self.mode,
+            "max_attempts": self.max_attempts,
             "zones": self._get_zones(),
             # SDK doesn't supply url_name.
             "url_name": getattr(self, 'url_name', ''),
@@ -433,6 +440,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         return {
             'items': item_state,
             'finished': is_finished,
+            'num_attempts': self.num_attempts,
             'overall_feedback': self.data['feedback']['finish' if is_finished else 'start'],
         }
 
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -326,13 +326,174 @@
     outline: 2px solid white;
 }
 
-/*** KEYBOARD HELP ***/
+/*** edX pattern library components ***/
+
+/* reset stock edx-platform button styles */
+.xblock--drag-and-drop button,
+.xblock--drag-and-drop button:hover,
+.xblock--drag-and-drop button.is-hovered,
+.xblock--drag-and-drop button:focus,
+.xblock--drag-and-drop button.is-focused,
+.xblock--drag-and-drop button:active,
+.xblock--drag-and-drop button.is-active {
+    box-shadow: none;
+    text-shadow: none;
+    background-image: none;
+}
+
+.xblock--drag-and-drop .btn-default,
+.xblock--drag-and-drop .btn-brand {
+    display: inline-block;
+    font-weight: normal;
+    background: #0079bc;
+    color: #fcfcfc;
+    -webkit-transition: color 0.125s ease-in-out 0s, border-color 0.125s ease-in-out 0s, background 0.125s ease-in-out 0s, box-shadow 0.125s ease-in-out 0s;
+    -moz-transition: color 0.125s ease-in-out 0s, border-color 0.125s ease-in-out 0s, background 0.125s ease-in-out 0s, box-shadow 0.125s ease-in-out 0s;
+    transition: color 0.125s ease-in-out 0s, border-color 0.125s ease-in-out 0s, background 0.125s ease-in-out 0s, box-shadow 0.125s ease-in-out 0s;
+    border-radius: 3px;
+    border: 1px solid #0079bc;
+    padding: 0.625em 1.25em;
+    font-size: 1em;
+}
+
+.xblock--drag-and-drop .btn-default {
+    border-color: transparent;
+    background: transparent;
+    color: #0074b4
+}
+
+.xblock--drag-and-drop .btn-brand {
+  border-color: #0075b4;
+  background: white;
+  color: #0075b4;
+}
+
+.xblock--drag-and-drop .btn-small {
+  padding: 0.625em 0.625em;
+  font-size: 0.875em;
+}
+
+.xblock--drag-and-drop .btn-default:hover,
+.xblock--drag-and-drop .btn-default.is-hovered,
+.xblock--drag-and-drop .btn-default:focus,
+.xblock--drag-and-drop .btn-default.is-focused {
+    background-color: transparent;
+    border: 1px solid #0074b4;
+    color: #0074b4
+}
+
+.xblock--drag-and-drop .btn-default:active,
+.xblock--drag-and-drop .btn-default.is-pressed,
+.xblock--drag-and-drop .btn-default.is-active {
+    border-color: #0074b4;
+    color: #0074b4
+}
+
+.xblock--drag-and-drop .btn-default:disabled,
+.xblock--drag-and-drop .btn-default.is-disabled {
+    color: #6b6969
+}
+
+.xblock--drag-and-drop .btn-brand {
+    border-color: #0074b4;
+    background: #0074b4;
+    color: #fcfcfc
+}
+
+.xblock--drag-and-drop .btn-brand:hover,
+.xblock--drag-and-drop .btn-brand.is-hovered,
+.xblock--drag-and-drop .btn-brand:focus,
+.xblock--drag-and-drop .btn-brand.is-focused {
+    border-color: #008bd8;
+    background-color: #008bd8;
+    color: #fcfcfc
+}
+
+.xblock--drag-and-drop .btn-brand:active,
+.xblock--drag-and-drop .btn-brand.is-pressed,
+.xblock--drag-and-drop .btn-brand.is-active {
+    border-color: #0074b4;
+    background: #0074b4
+}
+
+.xblock--drag-and-drop .btn-brand:disabled,
+.xblock--drag-and-drop .btn-brand.is-disabled {
+    border-color: #d2d0d0;
+    background: #f2f3f3;
+    color: #676666
+}
+
+
+/*** ACTIONS TOOLBAR ***/
+
+.xblock--drag-and-drop .actions-toolbar {
+    min-height: 3.75em;
+    width: auto;
+    position: relative;
+}
+
+.xblock--drag-and-drop .actions-toolbar .action-toolbar-item {
+    display: inline-block;
+    margin: 10px 0;
+}
+
+.xblock--drag-and-drop .attempts-used {
+    margin-left: 0.675em;
+    font-size: 0.875em;
+    color: #666;
+}
+
+.xblock--drag-and-drop .actions-toolbar .action-toolbar-item.sidebar-buttons {
+    text-align: left;
+    display: block;
+}
+
+@media (min-width: 768px) {
+    .xblock--drag-and-drop .actions-toolbar .action-toolbar-item.sidebar-buttons {
+        float: right;
+        margin: 0;
+        padding-right: -5px;
+    }
+}
+
+.xblock--drag-and-drop .sidebar-buttons .sidebar-button-wrapper {
+    border-right: 1px solid #ddd;
+    border-collapse: collapse;
+    padding: 0 5px;
+    display: inline-block;
+    height: 100%;
+}
+
+.xblock--drag-and-drop .sidebar-buttons .sidebar-button-wrapper:first-child {
+    padding-left: 0;
+}
+
+.xblock--drag-and-drop .sidebar-buttons .sidebar-button-wrapper:last-child {
+    border-right: none;
+    padding-right: 0;
+}
+
+.xblock--drag-and-drop .sidebar-buttons .btn-brand {
+    display: inline-block;
+    padding: 3px 5px;
+}
 
 .xblock--drag-and-drop .keyboard-help {
     margin-top: 3px;
     margin-bottom: 6px;
 }
 
+.xblock--drag-and-drop .btn-icon {
+    display: block;
+}
+
+.xblock--drag-and-drop .reset-button {
+    margin-top: 3px;
+}
+
+/*** ACTIONS TOOLBAR END ***/
+
+/*** KEYBOARD HELP ***/
 .xblock--drag-and-drop .keyboard-help-dialog {
     position: fixed;
     left: 50%;
@@ -382,26 +543,7 @@
     margin-left: 2%;
 }
 
-.xblock--drag-and-drop .link-button {
-    padding: 0;
-    margin: 0;
-    cursor: pointer;
-    color: #2d74b3;
-    font-weight: normal;
-    font-size: 12pt;
-    background: none;
-    box-shadow: none;
-    border: none;
-}
-
-.xblock--drag-and-drop .reset-button {
-    float: right;
-    margin-top: 3px;
-}
-
-.xblock--drag-and-drop .link-button:focus {
-    outline: 2px solid #2d74b3;
-}
+/*** KEYBOARD HELP END ***/
 
 /* Make sure screen-reader content is hidden in the workbench: */
 .xblock--drag-and-drop .sr {
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -228,49 +228,85 @@ function DragAndDropTemplates(configuration) {
 
     var feedbackTemplate = function(ctx) {
         var feedback_display = ctx.feedback_html ? 'block' : 'none';
-        var reset_button_display = ctx.display_reset_button ? 'block' : 'none';
         var properties = { attributes: { 'aria-live': 'polite' } };
         return (
             h('section.feedback', properties, [
-                h(
-                    'button.reset-button.unbutton.link-button',
-                    { style: { display: reset_button_display }, attributes: { tabindex: 0 }, 'aria-live': 'off'},
-                    gettext('Reset problem')
-                ),
                 h('h3.title1', { style: { display: feedback_display } }, gettext('Feedback')),
                 h('p.message', { style: { display: feedback_display }, innerHTML: ctx.feedback_html })
             ])
         );
     };
 
-    var keyboardHelpTemplate = function(ctx) {
-        var dialog_attributes = { role: 'dialog', 'aria-labelledby': 'modal-window-title' };
-        var dialog_style = {};
+    var keyboardHelpPopupTemplate = function(ctx) {
+        var labelledby_id = 'modal-window-title-'+configuration.url_name;
         return (
-            h('section.keyboard-help', [
-                h('button.keyboard-help-button.unbutton.link-button', { attributes: { tabindex: 0 } }, gettext('Keyboard Help')),
-                h('div.keyboard-help-dialog', [
-                    h('div.modal-window-overlay'),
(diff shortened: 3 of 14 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=86&repo=edx-solutions%2Fxblock-drag-and-drop-v2

e-kolpakov:
@mtyaka for some reason tests are extremely flaky - at least one `parameterized_final_feedback_and_reset` or zone align test fail, but not a single one failed two times in a row + everything passes locally. I don't expect to change anything, so please ignore the failing CI build. Meanwhile, I'll try to make it pass.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -58,6 +58,16 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         default=STANDARD_MODE
     )
 
+    max_attempts = Integer(
+        display_name=_("Maximum number of attempts"),
+        help=_(
+            "Limits the number of attempts learner can use to complete the problem. "
+            "For unlimited attempts, set to zero. This setting have no effect in standard mode."
```
mtyaka: typo: "have no effect" -> "has no effect"

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -178,51 +187,79 @@ function DragNDropTemplates(url_name) {
...
-                                h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')),
-                                h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" to drop the item on the current zone.')),
-                                h('li', gettext('Press "Esc" if you want to cancel the drop operation (for example, to select a different item).')),
-                            ])
-                        ]),
-                        h('div.modal-actions', [
-                            h('button.modal-dismiss-button', gettext("OK"))
+            h('div.keyboard-help-dialog', [
+                h('div.modal-window-overlay'),
+                h('div.modal-window', { attributes: { role: 'dialog', 'aria-labelledby': 'modal-window-title' }, style: {} }, [
```
mtyaka: Redundant `style: {}` property.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -178,51 +187,79 @@ function DragNDropTemplates(url_name) {
...
     };
 
+    var submitAnswerTemplate = function(ctx) {
+        var attemptsUsedDisplay = ctx.max_attempts > 0 ? 'inline': 'none';
+        var button_enabled = ctx.items.some(function(item) { return item.is_placed}) &&
+            (ctx.max_attempts == 0 || ctx.max_attempts > ctx.num_attempts);
+
+        return (
+          h("section.submit-answer", {}, [
+              h("button.btn-brand.submit-answer-button", { disabled: !button_enabled}, gettext("Submit")),
```
mtyaka: nit: extra space between `{` and `dissabled`.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -178,51 +187,79 @@ function DragNDropTemplates(url_name) {
...
+              h(
+                  "span.attempts-used", {'style': {'display': attemptsUsedDisplay}},
```
mtyaka: nit: `style` would be more consistent than `'style'` (same for `'display'`)

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -178,51 +187,79 @@ function DragNDropTemplates(url_name) {
...
+                h(
+                    'button.unbutton.btn-brand.btn-small.keyboard-help-button.link-button', { attributes: { tabindex: 0 } },
+                    [
+                        h("i.btn-icon.fa.fa-question"),
+                        gettext('Keyboard Help')
+                    ]
+                ),
+                h(
+                    'button.unbutton.btn-brand..btn-small.reset-button.link-button',
+                    { disabled: ctx.disable_reset_button, attributes: { tabindex: 0 }, 'aria-live': 'off'},
```
mtyaka: `aria-live` should be inside `attributes`. Maybe we don't need it at all since this is not a live region anymore?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -178,51 +187,79 @@ function DragNDropTemplates(url_name) {
...
+            h("section.sidebar-buttons", {}, [
+                h(
+                    'button.unbutton.btn-brand.btn-small.keyboard-help-button.link-button', { attributes: { tabindex: 0 } },
+                    [
+                        h("i.btn-icon.fa.fa-question"),
+                        gettext('Keyboard Help')
+                    ]
+                ),
+                h(
+                    'button.unbutton.btn-brand..btn-small.reset-button.link-button',
```
mtyaka: Double dot in `btn-brand..btn-small`

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -317,13 +317,105 @@
...
+
+.xblock--drag-and-drop .attempts-used {
+    margin-left: 5px;
+    font-size: 0.875rem;
+    color: #666;
+}
+
+.xblock--drag-and-drop .sidebar-buttons {
+    float: right;
+    display: flex;
```
mtyaka: I don't think `display: flex` does anything in combination with `float`.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -274,23 +311,30 @@ function DragNDropTemplates(url_name) {
...
-                    ]
-                    ),
+                    ]),
                 ]),
-                keyboardHelpTemplate(ctx),
+                h("section.actions-toolbar", {}, [
+                    sidebarTemplate(ctx), // sidebar has float:right, so putting it before the content to float
+                    (ctx.show_submit_answer ? submitAnswerTemplate(ctx) : null)
+                ]),
+                h("div", {style: {clear: 'both'}}), // div to clear floating
```
mtyaka: Can we get rid of this redundant element? Perhaps instead of floating the sidebar position it absolutely with `right: 0`.

drag_and_drop_v2/drag_and_drop_v2.py
mtyaka: Actually, this field and related studio edit functionality have already been implemented in https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/81, you should probably remove this code from this PR.

mtyaka:
@e-kolpakov Thanks for the updates! FYI both #81 and #82 have been merged into master, so if you rebase you can remove the extra commits from this PR.

I have been testing this locally. The code looks good and it's working correctly. I only have a few comments regarding the styles of the buttons:

1. In the Studio, the buttons and the "attempts" text look really tiny (see screenshot).
2. I think it would look better if the margin/padding between the submit button and the "You have used ..." text would be larger, closer to the screenshots from https://github.com/edx/edx-platform/pull/13052. Same thing for "Keyboard Help"/"Reset" buttons - I would increase the padding between button text and the vertical line that divides the buttons.
3. The vertical line between "Keyboard Help" and "Reset" could be a bit brighter. It looks brighter in CAPA PR. Also, if you look closely, the top and bottom of the line are not straight (they are affected by buttons round borders).
4. I think it would work better if disabled "Reset" button didn't have a grey background (the text should be greyed out, but the background should remain white). I believe that's how it works in the linked CAPA PR. The grey background draws too much attention to the disabled "Reset" button.

**Tiny buttons in the Studio**:

screen shot 2016-07-31 at 19 45 51

**Buttons in the LMS**:

The sizes are good, but notice the issues outlined above:
- the paddings could be increased
- the "Reset" button should not have a grey background
- the line between the right floated buttons is too dark and has rounded endings

screen shot 2016-07-31 at 20 17 08

e-kolpakov:
Rebased from a909bd8

e-kolpakov:
@mtyaka I've addressed the issues you have outlined - please review one more time.

mtyaka:
Thanks @e-kolpakov, it looks very nice!

:+1:

- [x] I tested to make sure the buttons work correctly and look pretty
- [x] I read through the code
- [ ] ~~Includes documentation~~ (We will update the README (and our demo course) once assessment mode functionality is complete.)

e-kolpakov:
@cahrens, @staubina, @cptvitamin, @sstack22 this Pr is ready for your review.

cahrens:
@alisan @chris-mike, look, they've copied our new button bar design! 💯

cahrens:
I believe the number of attempts remaining should be linked to the Submit button through aria-describedby. @cptvitamin can confirm.

cahrens:
In Studio the icons are italicized.

image

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -187,51 +187,82 @@ function DragAndDropTemplates(configuration) {
...
-                                h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')),
-                                h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" to drop the item on the current zone.')),
-                                h('li', gettext('Press "Esc" if you want to cancel the drop operation (for example, to select a different item).')),
-                            ])
-                        ]),
-                        h('div.modal-actions', [
-                            h('button.modal-dismiss-button', gettext("OK"))
+            h('div.keyboard-help-dialog', [
+                h('div.modal-window-overlay'),
+                h('div.modal-window', {attributes: {role: 'dialog', 'aria-labelledby': 'modal-window-title'}}, [
```
cahrens: Incorrect aria-labelledby.

image

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -187,51 +187,82 @@ function DragAndDropTemplates(configuration) {
...
+
+    var sidebarButtonTemplate = function(buttonClass, iconClass, buttonText, disabled) {
+        var disabled = disabled || false;
+        return (
+            h('span.sidebar-button-wrapper', {}, [
+                h(
+                    'button.unbutton.btn-default.btn-small.'+buttonClass,
+                    {disabled: disabled, attributes: {tabindex: 0}},
+                    [
+                        h("i.btn-icon.fa."+iconClass),
```
cahrens: As mentioned elsewhere, icons should be spans instead of i tags. And add aria-hidden (if there is not visible text associated with the icon, add screenreader text as well).

cahrens:
You should make the Submit/Help/Reset bar responsive for small displays (Mobile). When wrapping is necessary, the Help/Reset buttons should appear above the Submit button.

image

e-kolpakov:
@cahrens Fixed everything you mentioned except focus for submit button - there are no event handler for it yet (I'm going to add it in a follow up PR no later than this week, hopefully before the release is cut), so there's just no place to put focus-handling logic.

But I've thought about it and it looks like there are two options: if everything is correct, focus should go to feedback section; otherwise, return to first item in item bank.

One side effect of Keyboard Help and Reset appearing before the submit is that submit gets tab focus *after* Keyboard Help and Reset buttons. It feels a little wrong that element to the left receives focus after elements to the right, but I was unable to find a solution that would satisfy all the conditions:

1. Kbd Help and Reset stick to the right when there's enough width to display them and submit on one line
2. Kbd Help and Reset appears above Submit when there aren't
3. Submit receives focus before Kbd Help and Reset

So, since I've self-imposed condition 3, I just decided to sacrifice it. But if you've got an idea of how it can be achieved - I would be grateful to hear it. Current solution just uses float: right on KbdHelp wrapper block that appears in html markup before Submit. Obvious solution of using tabindex to control the order does not work well, as *all* the elements on page have no tabindex, so after zones are tabbed through focus does not land at Submit/KbdHelp/Reset at all.

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @cahrens technically I've just moved this code around (so this is an issue with previous versions of the code), but I've fixed it anyway.

staubina:
I mentioned this in https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/84, but thought it warranted mention here as well.

The focus goes to the bottom of the page when selecting a drop zone before selecting something to drag. Is this expected behavior? Is there a Bug for this?

e-kolpakov:
@staubina I believe it would be best to address this problem as part of #84, so that we don't end up with two different solutions.

cahrens:
@e-kolpakov Regarding focus traversal order, we actually WANT focus to go first to the Help and Reset buttons. That is the behavior that we will have for capa.

The reason is that for screenreader users, it is important that they know about the availability of the additional options (especially for capa, where there may be a Hint button) before they press Submit. 

So for consistency, please restore the behavior of focus moving first to Help and Reset (which is also why they should wrap above the Submit button for smaller screens).

e-kolpakov:
@cahrens I believe it already does just this, but the sandbox is not updated yet - I'll let you know when it is done.

staubina:
> @staubina I believe it would be best to address this problem as part of #84, so that we don't end up with two different solutions.

@e-kolpakov As long as it is documented somewhere if we want to fix it.

e-kolpakov:
@staubina I'm afraid I'm not quite sure what you mean here. I definitely don't know if it is documented somewhere and if we want to fix it or not (and I don't think I'm to decide this). What I meant is to handle this potential problem as part of #84, including creating a task to fix it if needed.

cahrens:
@e-kolpakov I do believe the focus traversal order was correct before-- I was just responding to your comment. I think I misunderstood it, as I thought you had implemented your third option.

e-kolpakov:
@cahrens no, I was trying to explain that I was not able to do that :)

e-kolpakov:
@cahrens I've just updated the sandbox

cahrens:
For focus after pressing Reset, it might be better to put the focus up on the problem description (so it is read again by the screen reader). @cptvitamin can advise on this detail in his review.

cahrens:
I still have a concern about the button bar on small devices-- see my note above about wrapping.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -202,7 +202,7 @@ function DragAndDropTemplates(configuration) {
                 h('div.modal-window-overlay'),
                 h('div.modal-window', {attributes: {role: 'dialog', 'aria-labelledby': 'modal-window-title'}}, [
                     h('div.modal-header', [
-                        h('h2.modal-window-title', gettext('Keyboard Help'))
+                        h('h2.modal-window-title#modal-window-title', gettext('Keyboard Help'))
```
cahrens: This caused a non-unique ID when 2 problems are on the same page. It is very easy to look for these sort of issues using Chrome and the aXe developer tool plugin.

image

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -238,15 +238,14 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
-    var sidebarButtonTemplate = function(buttonClass, iconClass, buttonText, disabled) {
-        var disabled = disabled || false;
+    var sidebarButtonTemplate = function(buttonClass, iconClass, buttonText, disabled, tabindex) {
```
cahrens: Doesn't look like this tabindex variable is needed.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -187,51 +187,81 @@ function DragAndDropTemplates(configuration) {
...
-                                h('li', gettext('Press "Esc" if you want to cancel the drop operation (for example, to select a different item).')),
-                            ])
-                        ]),
-                        h('div.modal-actions', [
-                            h('button.modal-dismiss-button', gettext("OK"))
+            h('div.keyboard-help-dialog', [
+                h('div.modal-window-overlay'),
+                h('div.modal-window', {attributes: {role: 'dialog', 'aria-labelledby': 'modal-window-title'}}, [
+                    h('div.modal-header', [
+                        h('h2.modal-window-title#modal-window-title', gettext('Keyboard Help'))
```
cahrens: This caused a non-unique ID when 2 problems are on the same page. It is very easy to look for these sort of issues using Chrome and the aXe developer tool plugin.

image

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -187,51 +187,81 @@ function DragAndDropTemplates(configuration) {
...
+              h(
+                  "span.attempts-used", {style: {display: attemptsUsedDisplay}},
+                  gettext("You have used {used} of {total} attempts.")
+                      .replace("{used}", ctx.num_attempts).replace("{total}", ctx.max_attempts)
+              )
+          ])
+        );
+    };
+
+    var sidebarButtonTemplate = function(buttonClass, iconClass, buttonText, disabled, tabindex) {
```
cahrens: Doesn't look like this tabindex variable is needed.

e-kolpakov:
@cahrens 

* Focus on description after reset - I've tried to do that, but for some reason neither Chrome nor Firefox set focus on description section, and AFAIK screen reader does not read it. Maybe I'm doing something wrong (I haven't spent more than 15 minutes trying to achieve this), so I would wait for @cptvitamin comments on this.
* Buttons bar on small screens - implemented (looked up in https://github.com/edx/edx-platform/pull/13052)
* Other review notes - addressed

staubina:
My issue was corrected in another PR
👍

cahrens:
@e-kolpakov Can you update the sandbox? I'm still seeing the help and reset buttons be right-aligned on wrap.

e-kolpakov:
@cahrens sure - give me 15 minutes.

e-kolpakov:
@cahrens updated.

cahrens:
I'm still seeing issues with wrapping.

image

e-kolpakov:
@cahrens that's because I'm just copying WIP styles from CAPA problems redesign: https://github.com/edx/edx-platform/pull/13052#issuecomment-237292836 :)

cahrens:
Nit: the Help/Reset buttons stay left-aligned for a time during the "wrapping" phase. See image below. Perhaps @alisan617 will come up with a better approach in the capa work, and then you guys can copy it then.

image

cahrens:
👍

e-kolpakov:
@cahrens Thank you!

sstack22:
Sorry for not commenting sooner - 👍

bradenmacdonald:
Merging now. Thanks all!

Due to scheduling requirements, full a11y review of this work is being tracked separately as [SOL-1978](https://openedx.atlassian.net/browse/SOL-1978)

PR #87, opened 2016-08-02
Click handlers and backend processing for assessment mode

e-kolpakov:
**Description:** This PR actually implements submitting attempts for Assessment Mode. As of this PR, the minimum viable product implementation of Assessment Mode in Drag and Drop v2 is complete.

**Dependencies:** #86 (now merged)

**JIRA:** Part of [SOL-1944](https://openedx.atlassian.net/browse/SOL-1944)

**Discussions**: [Discovery document for DnDv2 assessment mode](https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit) (relevant section: 2.2.3)

**Sandbox URLs**: 
* LMS: http://dndv2-sandbox8.opencraft.hosting/
* Studio: http://studio-dndv2-sandbox8.opencraft.hosting/

Updated 2016-08-12 to commit 5b837e7

**Testing instructions:**
1. Create/edit DnDv2; put it into assessment mode
2. Set "Max attempts" to N: N > 2
3. Publish, open in LMS.
4. Note that "Submit" button is displayed. Make sure problem is in its initial state (reset if needed). Submit button should be disabled. Attempts info should not be displayed.
5. Put any item to correct zone. "Submit" button should become enabled.
6. Click "Submit". Expected to happen:
6.1. Submit button shows spinner until AJAX request is sent
6.2. When AJAX is completed, incorrect items are returned to item bank, other items stay where they were.
6.3. Feedback message is updated to contain info about correctly placed items, misplaced items and items that should be put, but was not. If there were any misplaced items, feedback should also contain note about them being returned to item bank. If "initial" overall feedback is specified, it should be shown as well.
6.4. Attempts info should increase the number of used attempts.
7. Use N-1 attempts to test various combinations of correct, misplaced and missing items (note pluralization should be appropriate). Same things as in (6) should happen.
8. Submit final attempt. Expected to happen:
8.1. Same as 6.1. - spinner
8.2. All items stay where they were (misplaced are not returned to bank, as opposed to 6.2)
8.3. All of 6.3 + note about using final attempt and final grade are shown.
8.4. Attempts info should read "Used N of N attempts".
8.5. Submit and Reset buttons are permanently disabled.

**Reviewers**
- [x] @mtyaka 
- [ ] a11y: @cptvitamin
- [ ] Product: @sstack22

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,11 +96,19 @@ and Drop component to a lesson, then click the `EDIT` button.
 
 ![Edit view](/doc/img/edit-view.png)
 
-In the first step, you can set some basic properties of the component,
-such as the title, the mode, the maximum number of attempts, the maximum score,
+In the first step, you can set some basic properties of the component, such as
+the title, the problem mode, the maximum number of attempts, the maximum score,
 the problem text to render above the background image, the introductory feedback
 (shown initially), and the final feedback (shown after the learner successfully
-completes the drag and drop problem).
+completes the drag and drop problem, or when the learner runs out of attempts).
+
+There are two problem modes available:
+
+* **Standard**: In this mode, the learner gets immediate feedback on each
+  attempt to place an item, and the number of attempts is not limited.
+* **Assessment**: In this mode, the learner places all items on the board and
+  then clicks a "Submit" button to get feedback.  The number of attempts can be
+  limited.
 
 ![Drop zone edit](/doc/img/edit-view-zones.png)
 
@@ -126,13 +134,14 @@ potentially, overlap the zones below.
 
 ![Drag item edit](/doc/img/edit-view-items.png)
 
-In the final step, you define the background and text color for drag
-items, as well as the drag items themselves. A drag item can contain
-either text or an image. You can define custom success and error
-feedback for each item. The feedback text is displayed in a popup
-after the learner drops the item on a zone - the success feedback is
-shown if the item is dropped on a correct zone, while the error
-feedback is shown when dropping the item on an incorrect drop zone.
+In the final step, you define the background and text color for drag items, as
+well as the drag items themselves. A drag item can contain either text or an
+image. You can define custom success and error feedback for each item. In
+standard mode, the feedback text is displayed in a popup after the learner drops
+the item on a zone - the success feedback is shown if the item is dropped on a
+correct zone, while the error feedback is shown when dropping the item on an
+incorrect drop zone.  In assessment mode, the success and error feedback texts
+are not used.
 
 You can select any number of zones for an item to belong to using
 the checkboxes; all zones defined in the previous step are available.
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
 #
-
+""" Drag and Drop v2 XBlock """
 # Imports ###########################################################
 
-import json
-import webob
 import copy
+import json
 import urllib
+import webob
 
 from xblock.core import XBlock
 from xblock.exceptions import JsonHandlerError
@@ -15,7 +15,7 @@ from xblock.fragment import Fragment
 from xblockutils.resources import ResourceLoader
 from xblockutils.settings import XBlockWithSettingsMixin, ThemableXBlockMixin
 
-from .utils import _  # pylint: disable=unused-import
+from .utils import _, DummyTranslationService, FeedbackMessage, FeedbackMessages
 from .default_data import DEFAULT_DATA
 
 
@@ -34,6 +34,22 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     """
     STANDARD_MODE = "standard"
     ASSESSMENT_MODE = "assessment"
+
+    SOLUTION_CORRECT = "correct"
+    SOLUTION_PARTIAL = "partial"
+    SOLUTION_INCORRECT = "incorrect"
+
+    GRADE_FEEDBACK_CLASSES = {
+        SOLUTION_CORRECT: FeedbackMessages.MessageClasses.CORRECT_SOLUTION,
+        SOLUTION_PARTIAL: FeedbackMessages.MessageClasses.PARTIAL_SOLUTION,
+        SOLUTION_INCORRECT: FeedbackMessages.MessageClasses.INCORRECT_SOLUTION,
+    }
+
+    PROBLEM_FEEDBACK_CLASSES = {
+        SOLUTION_CORRECT: FeedbackMessages.MessageClasses.CORRECT_SOLUTION,
+        SOLUTION_PARTIAL: None,
+        SOLUTION_INCORRECT: None
+    }
 
     display_name = String(
         display_name=_("Title"),
@@ -127,7 +143,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         default={},
     )
 
-    num_attempts = Integer(
+    attempts = Integer(
         help=_("Number of attempts learner used"),
         scope=Scope.user_state,
         default=0
@@ -137,6 +153,12 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         help=_("Indicates whether a learner has completed the problem at least once"),
         scope=Scope.user_state,
         default=False,
+    )
+
+    grade = Float(
+        help=_("Keeps maximum achieved score by student"),
+        scope=Scope.user_state,
+        default=0
     )
 
     block_settings_key = 'drag-and-drop-v2'
@@ -179,6 +201,9 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
 
         def items_without_answers():
+            """
+            Removes feedback and answer from items
+            """
             items = copy.deepcopy(self.data.get('items', ''))
             for item in items:
                 del item['feedback']
@@ -278,6 +303,9 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
     @XBlock.json_handler
     def studio_submit(self, submissions, suffix=''):
+        """
+        Handles studio save.
+        """
         self.display_name = submissions['display_name']
         self.mode = submissions['mode']
         self.max_attempts = submissions['max_attempts']
@@ -294,49 +322,285 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         }
 
     @XBlock.json_handler
-    def do_attempt(self, attempt, suffix=''):
-        item = self._get_item_definition(attempt['val'])
+    def drop_item(self, item_attempt, suffix=''):
+        """
+        Handles dropping item into a zone.
+        """
+        self._validate_drop_item(item_attempt)
 
-        state = None
-        zone = None
-        feedback = item['feedback']['incorrect']
-        overall_feedback = None
-        is_correct = False
-
-        if self._is_attempt_correct(attempt):  # Student placed item in a correct zone
-            is_correct = True
-            feedback = item['feedback']['correct']
-            state = {
-                'zone': attempt['zone'],
-                'correct': True,
-                'x_percent': attempt['x_percent'],
-                'y_percent': attempt['y_percent'],
-            }
-
-        if state:
-            self.item_state[str(item['id'])] = state
-            zone = self._get_zone_by_uid(state['zone'])
+        if self.mode == self.ASSESSMENT_MODE:
+            return self._drop_item_assessment(item_attempt)
+        elif self.mode == self.STANDARD_MODE:
+            return self._drop_item_standard(item_attempt)
         else:
-            zone = self._get_zone_by_uid(attempt['zone'])
+            raise JsonHandlerError(
+                500,
+                self.i18n_service.gettext("Unknown DnDv2 mode {mode} - course is misconfigured").format(self.mode)
+            )
+
+    @XBlock.json_handler
+    def do_attempt(self, data, suffix=''):
+        """
+        Checks submitted solution and returns feedback.
+
+        Raises:
+             * JsonHandlerError with 400 error code in standard mode.
+             * JsonHandlerError with 409 error code if no more attempts left
+        """
+        self._validate_do_attempt()
+
+        self.attempts += 1
+        self._mark_complete_and_publish_grade()  # must happen before _get_feedback
+
+        overall_feedback_msgs, misplaced_ids = self._get_feedback()
+
+        for item_id in misplaced_ids:
+            del self.item_state[item_id]
+
+        return {
+            'attempts': self.attempts,
+            'misplaced_items': list(misplaced_ids),
+            'overall_feedback': self._present_overall_feedback(overall_feedback_msgs)
+        }
+
+    @XBlock.json_handler
+    def publish_event(self, data, suffix=''):
+        """
+        Handler to publish XBlock event from frontend
+        """
+        try:
+            event_type = data.pop('event_type')
+        except KeyError:
+            return {'result': 'error', 'message': 'Missing event_type in JSON data'}
+
+        self.runtime.publish(self, event_type, data)
+        return {'result': 'success'}
+
+    @XBlock.json_handler
+    def reset(self, data, suffix=''):
+        """
+        Resets problem to initial state
+        """
+        self.item_state = {}
+        return self._get_user_state()
+
+    @XBlock.json_handler
+    def expand_static_url(self, url, suffix=''):
+        """ AJAX-accessible handler for expanding URLs to static [image] files """
+        return {'url': self._expand_static_url(url)}
+
+    @property
+    def i18n_service(self):
+        """ Obtains translation service """
+        i18n_service = self.runtime.service(self, "i18n")
+        if i18n_service:
+            return i18n_service
+        else:
+            return DummyTranslationService()
+
+    @property
+    def target_img_expanded_url(self):
+        """ Get the expanded URL to the target image (the image items are dragged onto). """
+        if self.data.get("targetImg"):
+            return self._expand_static_url(self.data["targetImg"])
+        else:
+            return self.default_background_image_url
+
+    @property
+    def target_img_description(self):
+        """ Get the description for the target image (the image items are dragged onto). """
+        return self.data.get("targetImgDescription", "")
+
+    @property
+    def default_background_image_url(self):
+        """ The URL to the default background image, shown when no custom background is used """
+        return self.runtime.local_resource_url(self, "public/img/triangle.png")
+
+    @property
+    def attempts_remain(self):
+        """
+        Checks if current student still have more attempts.
+        """
+        return self.max_attempts is None or self.max_attempts == 0 or self.attempts < self.max_attempts
+
+    @XBlock.handler
+    def get_user_state(self, request, suffix=''):
+        """ GET all user-specific data, and any applicable feedback """
+        data = self._get_user_state()
+
+        return webob.Response(body=json.dumps(data), content_type='application/json')
+
+    def _validate_do_attempt(self):
+        """
+        Validates if `do_attempt` handler should be executed
+        """
+        if self.mode != self.ASSESSMENT_MODE:
+            raise JsonHandlerError(
+                400,
+                self.i18n_service.gettext("do_attempt handler should only be called for assessment mode")
+            )
+        if not self.attempts_remain:
+            raise JsonHandlerError(
+                409,
+                self.i18n_service.gettext("Max number of attempts reached")
+            )
+
+    def _get_feedback(self):
+        """
+        Builds overall feedback for both standard and assessment modes
+        """
+        answer_correctness = self._answer_correctness()
+        is_correct = answer_correctness == self.SOLUTION_CORRECT
+
+        if self.mode == self.STANDARD_MODE or not self.attempts:
+            feedback_key = 'finish' if is_correct else 'start'
+            return [FeedbackMessage(self.data['feedback'][feedback_key], None)], set()
+
+        required_ids, placed_ids, correct_ids = self._get_item_raw_stats()
+        missing_ids = required_ids - placed_ids
+        misplaced_ids = placed_ids - correct_ids
+
+        feedback_msgs = []
+
+        def _add_msg_if_exists(ids_list, message_template, message_class):
+            """ Adds message to feedback messages if corresponding items list is not empty """
+            if ids_list:
+                message = message_template(len(ids_list), self.i18n_service.ngettext)
+                feedback_msgs.append(FeedbackMessage(message, message_class))
+
+        _add_msg_if_exists(
+            correct_ids, FeedbackMessages.correctly_placed, FeedbackMessages.MessageClasses.CORRECTLY_PLACED
+        )
+        _add_msg_if_exists(misplaced_ids, FeedbackMessages.misplaced, FeedbackMessages.MessageClasses.MISPLACED)
+        _add_msg_if_exists(missing_ids, FeedbackMessages.not_placed, FeedbackMessages.MessageClasses.NOT_PLACED)
+
+        if misplaced_ids and self.attempts_remain:
+            feedback_msgs.append(
+                FeedbackMessage(FeedbackMessages.MISPLACED_ITEMS_RETURNED, None)
+            )
+
+        if self.attempts_remain and (misplaced_ids or missing_ids):
+            problem_feedback_message = self.data['feedback']['start']
+        else:
+            problem_feedback_message = self.data['feedback']['finish']
+
+        problem_feedback_class = self.PROBLEM_FEEDBACK_CLASSES.get(answer_correctness, None)
+        grade_feedback_class = self.GRADE_FEEDBACK_CLASSES.get(answer_correctness, None)
+
+        feedback_msgs.append(FeedbackMessage(problem_feedback_message, problem_feedback_class))
+
+        if not self.attempts_remain:
+            feedback_msgs.append(
+                FeedbackMessage(FeedbackMessages.FINAL_ATTEMPT_TPL.format(score=self.grade), grade_feedback_class)
+            )
+
+        return feedback_msgs, misplaced_ids
+
+    @staticmethod
+    def _present_overall_feedback(feedback_messages):
+        """
+        Transforms feedback messages into format expected by frontend code
+        """
+        return [
+            {"message": msg.message, "message_class": msg.message_class}
+            for msg in feedback_messages
+            if msg.message
+        ]
+
+    def _drop_item_standard(self, item_attempt):
+        """
+        Handles dropping item to a zone in standard mode.
+        """
+        item = self._get_item_definition(item_attempt['val'])
+
+        is_correct = self._is_attempt_correct(item_attempt)  # Student placed item in a correct zone
+        if is_correct:  # In standard mode state is only updated when attempt is correct
+            self.item_state[str(item['id'])] = self._make_state_from_attempt(item_attempt, is_correct)
+
+        self._mark_complete_and_publish_grade()  # must happen before _get_feedback
+        self._publish_item_dropped_event(item_attempt, is_correct)
+
+        item_feedback_key = 'correct' if is_correct else 'incorrect'
+        item_feedback = item['feedback'][item_feedback_key]
+        overall_feedback, __ = self._get_feedback()
+
+        return {
+            'correct': is_correct,
+            'finished': self._is_answer_correct(),
+            'overall_feedback': self._present_overall_feedback(overall_feedback),
+            'feedback': item_feedback
+        }
+
+    def _drop_item_assessment(self, item_attempt):
+        """
+        Handles dropping item into a zone in assessment mode
+        """
+        if not self.attempts_remain:
+            raise JsonHandlerError(409, self.i18n_service.gettext("Max number of attempts reached"))
+
+        item = self._get_item_definition(item_attempt['val'])
+
+        is_correct = self._is_attempt_correct(item_attempt)
+        # State is always updated in assessment mode to store intermediate item positions
+        self.item_state[str(item['id'])] = self._make_state_from_attempt(item_attempt, is_correct)
+
+        self._publish_item_dropped_event(item_attempt, is_correct)
+
+        return {}
+
+    def _validate_drop_item(self, item):
+        """
+        Validates `drop_item` parameters
+        """
+        zone = self._get_zone_by_uid(item['zone'])
         if not zone:
             raise JsonHandlerError(400, "Item zone data invalid.")
 
-        if self._is_finished():
-            overall_feedback = self.data['feedback']['finish']
+    @staticmethod
+    def _make_state_from_attempt(attempt, correct):
+        """
+        Converts "attempt" data coming from browser into "state" entry stored in item_state
+        """
+        return {
+            'zone': attempt['zone'],
+            'correct': correct,
+            'x_percent': attempt['x_percent'],
+            'y_percent': attempt['y_percent'],
+        }
 
-        # don't publish the grade if the student has already completed the problem
-        if not self.completed:
-            if self._is_finished():
-                self.completed = True
-            try:
-                self.runtime.publish(self, 'grade', {
-                    'value': self._get_grade(),
-                    'max_value': self.weight,
-                })
-            except NotImplementedError:
-                # Note, this publish method is unimplemented in Studio runtimes,
-                # so we have to figure that we're running in Studio for now
-                pass
+    def _mark_complete_and_publish_grade(self):
+        """
+        Helper method to update `self.completed` and submit grade event if appropriate conditions met.
+        """
+        # There's no going back from "completed" status to "incomplete"
+        self.completed = self.completed or self._is_answer_correct() or not self.attempts_remain
+        grade = self._get_grade()
+        # ... and from higher grade to lower
+        if grade > self.grade:
+            self.grade = grade
+            self._publish_grade()
+
+    def _publish_grade(self):
+        """
+        Publishes grade
+        """
+        try:
+            self.runtime.publish(self, 'grade', {
+                'value': self.grade,
+                'max_value': self.weight,
+            })
+        except NotImplementedError:
+            # Note, this publish method is unimplemented in Studio runtimes,
+            # so we have to figure that we're running in Studio for now
+            pass
+
+    def _publish_item_dropped_event(self, attempt, is_correct):
+        """
+        Publishes item dropped event.
+        """
+        item = self._get_item_definition(attempt['val'])
+        # attempt should already be validated here - not doing the check for existing zone again
+        zone = self._get_zone_by_uid(attempt['zone'])
 
         self.runtime.publish(self, 'edx.drag_and_drop_v2.item.dropped', {
             'item_id': item['id'],
@@ -344,24 +608,6 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
             'location_id': zone.get("uid"),
             'is_correct': is_correct,
         })
-
-        if self.mode == self.ASSESSMENT_MODE:
-            # In assessment mode we don't send any feedback on drop.
-            result = {}
-        else:
-            result = {
-                'correct': is_correct,
-                'finished': self._is_finished(),
-                'overall_feedback': overall_feedback,
-                'feedback': feedback
-            }
-
-        return result
-
-    @XBlock.json_handler
-    def reset(self, data, suffix=''):
-        self.item_state = {}
-        return self._get_user_state()
 
     def _is_attempt_correct(self, attempt):
         """
@@ -389,35 +635,6 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
                 pass
         return url
 
-    @XBlock.json_handler
-    def expand_static_url(self, url, suffix=''):
-        """ AJAX-accessible handler for expanding URLs to static [image] files """
-        return {'url': self._expand_static_url(url)}
-
-    @property
-    def target_img_expanded_url(self):
-        """ Get the expanded URL to the target image (the image items are dragged onto). """
-        if self.data.get("targetImg"):
-            return self._expand_static_url(self.data["targetImg"])
-        else:
-            return self.default_background_image_url
-
-    @property
-    def target_img_description(self):
-        """ Get the description for the target image (the image items are dragged onto). """
-        return self.data.get("targetImgDescription", "")
-
-    @property
-    def default_background_image_url(self):
-        """ The URL to the default background image, shown when no custom background is used """
-        return self.runtime.local_resource_url(self, "public/img/triangle.png")
-
-    @XBlock.handler
-    def get_user_state(self, request, suffix=''):
-        """ GET all user-specific data, and any applicable feedback """
-        data = self._get_user_state()
-        return webob.Response(body=json.dumps(data), content_type='application/json')
-
     def _get_user_state(self):
         """ Get all user-specific data, and any applicable feedback """
         item_state = self._get_item_state()
@@ -436,24 +653,38 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
                 else:
                     item['zone'] = 'unknown'
 
-        is_finished = self._is_finished()
+            # In assessment mode, if item is placed correctly and than the page is refreshed, "correct"
+            # will spill to the frontend, making item "disabled", thus allowing students to obtain answer by trial
+            # and error + refreshing the page. In order to avoid that, we remove "correct" from an item here
+            if self.mode == self.ASSESSMENT_MODE:
+                del item["correct"]
+
+        overall_feedback_msgs, __ = self._get_feedback()
+        if self.mode == self.STANDARD_MODE:
+            is_finished = self._is_answer_correct()
+        else:
+            is_finished = not self.attempts_remain
+
         return {
             'items': item_state,
             'finished': is_finished,
-            'num_attempts': self.num_attempts,
-            'overall_feedback': self.data['feedback']['finish' if is_finished else 'start'],
+            'attempts': self.attempts,
+            'overall_feedback': self._present_overall_feedback(overall_feedback_msgs)
         }
 
     def _get_item_state(self):
         """
-        Returns the user item state.
+        Returns a copy of the user item state.
         Converts to a dict if data is stored in legacy tuple form.
         """
+
+        # IMPORTANT: this method should always return a COPY of self.item_state - it is called from get_user_state
+        # handler and manipulated there to hide correctness of items placed
         state = {}
 
         for item_id, item in self.item_state.iteritems():
             if isinstance(item, dict):
-                state[item_id] = item
+                state[item_id] = item.copy()  # items are manipulated in _get_user_state, so we protect actual data
             else:
                 state[item_id] = {'top': item[0], 'left': item[1]}
 
@@ -512,17 +743,28 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         Returns a tuple representing the number of correctly-placed items,
         and the total number of items that must be placed on the board (non-decoy items).
         """
-        all_items = self.data['items']
+        required_items, __, correct_items = self._get_item_raw_stats()
+
+        return len(correct_items), len(required_items)
+
+    def _get_item_raw_stats(self):
+        """
+        Returns a 3-tuple containing required, placed and correct items.
+
+        Returns:
+            tuple: (required_items, placed_items, correct_items)
+                * required_items - IDs of items that must be placed on the board
+                * placed_items - IDs of items actually placed on the board
+                * correct_items - IDs of items that were placed correctly
+        """
+        all_items = [str(item['id']) for item in self.data['items']]
         item_state = self._get_item_state()
 
-        required_items = [str(item['id']) for item in all_items if self._get_item_zones(item['id']) != []]
-        placed_items = [item for item in required_items if item in item_state]
-        correct_items = [item for item in placed_items if item_state[item]['correct']]
+        required_items = set(item_id for item_id in all_items if self._get_item_zones(int(item_id)) != [])
+        placed_items = set(item_id for item_id in all_items if item_id in item_state)
+        correct_items = set(item_id for item_id in placed_items if item_state[item_id]['correct'])
 
-        required_count = len(required_items)
-        correct_count = len(correct_items)
-
-        return correct_count, required_count
+        return required_items, placed_items, correct_items
 
     def _get_grade(self):
         """
@@ -531,31 +773,32 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         correct_count, required_count = self._get_item_stats()
         return correct_count / float(required_count) * self.weight
 
-    def _is_finished(self):
+    def _answer_correctness(self):
         """
-        All items are at their correct place and a value has been
-        submitted for each item that expects a value.
+        Checks answer correctness:
+
+        Returns:
+            string: Correct/Incorrect/Partial
+                * Correct: All items are at their correct place.
+                * Partial: Some items are at their correct place.
+                * Incorrect: None items are at their correct place.
         """
         correct_count, required_count = self._get_item_stats()
-        return correct_count == required_count
+        if correct_count == required_count:
+            return self.SOLUTION_CORRECT
+        elif correct_count == 0:
+            return self.SOLUTION_INCORRECT
+        else:
+            return self.SOLUTION_PARTIAL
 
-    @XBlock.json_handler
-    def publish_event(self, data, suffix=''):
-        try:
-            event_type = data.pop('event_type')
-        except KeyError:
-            return {'result': 'error', 'message': 'Missing event_type in JSON data'}
+    def _is_answer_correct(self):
+        """
+        Helper - checks if answer is correct
 
-        self.runtime.publish(self, event_type, data)
-        return {'result': 'success'}
-
-    def _get_unique_id(self):
-        usage_id = self.scope_ids.usage_id
-        try:
-            return usage_id.name
-        except AttributeError:
-            # workaround for xblock workbench
-            return usage_id
+        Returns:
+            bool: True if current answer is correct
+        """
+        return self._answer_correctness() == self.SOLUTION_CORRECT
 
     @staticmethod
     def workbench_scenarios():
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -284,6 +284,57 @@
     border-top: solid 1px #bdbdbd;
 }
 
+.xblock--drag-and-drop .feedback p {
+    margin: 2px 0;
+    padding: 0.5em;
+    border: 2px solid #999;
+}
+
+.xblock--drag-and-drop .feedback p.correct {
+    color: #166e36;
+    border: 2px solid #166e36;
+}
+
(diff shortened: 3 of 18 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=87

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -412,12 +382,120 @@ def default_background_image_url(self):
         """ The URL to the default background image, shown when no custom background is used """
         return self.runtime.local_resource_url(self, "public/img/triangle.png")
 
+    @property
+    def attemps_remain(self):
+        return self.max_attempts is None or self.num_attempts < self.max_attempts
```
mtyaka: We might want to allow both `None` (empty) and `0` values of `max_attempts` to mean "unlimited attempts". It would be consistent with some other problem types where `0` means no limit.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -412,12 +382,120 @@ def default_background_image_url(self):
...
+            'zone': attempt['zone'],
+            'correct': correct,
+            'x_percent': attempt['x_percent'],
+            'y_percent': attempt['y_percent'],
+        }
+
+    def _mark_complete_and_publish_grade(self):
+        # don't publish the grade if the student has already completed the problem
+        if not self.completed:
+            self.completed = self._is_finished()
```
mtyaka: I *think* we will also want to set `self.completed` to `True` when the student runs out of attempts, even though none of their answers was fully correct.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -860,6 +859,44 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+
+        $.ajax({
+            type: 'POST',
+            url: runtime.handlerUrl(element, "do_attempt"),
+            data: '{}'
+        }).done(function(data){
+            state.num_attempts = data.num_attempts;
+            state.overall_feedback = data.feedback;
+            if (data.attempts_remain) {
+                for (var i=0; i < data.misplaced_items.length; i++) {
```
mtyaka: Nit: We usually prefer `data.misplaced_items.forEach(function(misplaced_item_id) { ... })`.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -412,12 +382,120 @@ def default_background_image_url(self):
...
+        if not self.attemps_remain:
+            raise JsonHandlerError(409, _("Max number of attempts reached"))
+
+        item = self._get_item_definition(item_attempt['val'])
+
+        is_correct = self._is_attempt_correct(item_attempt)
+        # State is always updated in assessment mode, to make do_attempt work correctly
+        self.item_state[str(item['id'])] = self._make_state_from_attempt(item_attempt, is_correct)
+
+        self._publish_item_dropped_event(item_attempt, self._is_attempt_correct(item_attempt))
```
mtyaka: `self._is_attempt_correct(item_attempt)` is already invoked a couple of lines above, you can replace the second parameter with `is_correct` variable.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -412,12 +382,120 @@ def default_background_image_url(self):
...
+        }
+
+    def _drop_item_assessment(self, item_attempt):
+        if not self.attemps_remain:
+            raise JsonHandlerError(409, _("Max number of attempts reached"))
+
+        item = self._get_item_definition(item_attempt['val'])
+
+        is_correct = self._is_attempt_correct(item_attempt)
+        # State is always updated in assessment mode, to make do_attempt work correctly
```
mtyaka: Nit: instead of "to make do_attempt work correctly", I would say something like "to store intermediate item positions".

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -294,101 +294,71 @@ def studio_submit(self, submissions, suffix=''):
...
-            'is_correct': is_correct,
-        })
+        __, placed, correct = self._get_item_raw_stats()
+        placed_ids, correct_ids = set(placed), set(correct)
+        misplaced_ids = placed_ids - correct_ids
 
-        if self.mode == self.ASSESSMENT_MODE:
-            # In assessment mode we don't send any feedback on drop.
-            result = {}
+        feedback_msgs = [_('Correctly placed {correct_count} items, misplaced {misplaced_count} items.').format(
```
mtyaka: Can we support pluralization? This returns "Correctly placed 1 items, misplaced 1 items.". `ngettext` might help.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -294,101 +294,71 @@ def studio_submit(self, submissions, suffix=''):
...
+        for item_id in misplaced_ids:
+            del self.item_state[item_id]
+
+        if not self.attemps_remain:
+            feedback_msgs.append(_('Final attempt was used, final score is {score}').format(score=self._get_grade()))
+
+        return {
+            'num_attempts': self.num_attempts,
+            'misplaced_items': list(misplaced_ids),
+            'feedback': '\n'.join(feedback_msgs),
```
mtyaka: The newlines are ignored on the client, we should use `'<br/>'` instead of `'\n'`. Or maybe wrap each individual message in `<p>...<\p>`.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -860,6 +859,44 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+                    var misplaced_item_id = data.misplaced_items[i];
+                    delete state.items[misplaced_item_id]
+                }
+            }
+            else {
```
mtyaka: Nit: we usually don't put a line break between `}` and `else {`.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -294,101 +294,71 @@ def studio_submit(self, submissions, suffix=''):
...
+            'attempts_remain': self.attemps_remain
```
mtyaka: Nit: `attempts_remain` is somewhat redundant since we send `num_attempts` and the client knows the value of `max_attempts`.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -860,6 +859,44 @@ function DragAndDropBlock(runtime, element, configuration) {
         });
     };
 
+    var doAttempt = function(evt) {
```
mtyaka: When you click the button, it should immediately get disabled and optionally show some kind of a spinner while the xhr request is in progress.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -306,23 +306,41 @@ def drop_item(self, item_attempt, suffix=''):
...
+                'Not placed {missing_count} required items.',
+                missing_count
+            ).format(missing_count=missing_count)
+        ]
+
+        if misplaced_ids and self.attemps_remain:
             feedback_msgs.append(_('Misplaced items were returned to item bank.'))
-        else:
+
+        if not misplaced_ids and not missing_ids:
```
mtyaka: I was about to suggest to replace this with `if self._is_finished()` to make it clearer, but then I realized that `_is_finished()` is currently a bit broken in assessment mode, because it does not check decoy items. `_is_finished()` should only return true if all non-decoy items are in their correct positions and all decoy items are in the bank.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -885,17 +883,16 @@ function DragAndDropBlock(runtime, element, configuration) {
...
     var canReset = function() {
         return Object.keys(state.items).length > 0 &&
-            (
-                configuration.mode != DragAndDropBlock.ASSESSMENT_MODE ||
-                canSubmitAttempt()
-            )
+            (configuration.mode != DragAndDropBlock.ASSESSMENT_MODE || attemptsRemain())
+    };
+
+    var attemptsRemain = function() {
```
mtyaka: Nice refactoring.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -885,17 +883,16 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+        return configuration.max_attempts === null || configuration.max_attempts > state.num_attempts;
```
mtyaka: Can we treat zero the same as `null` here, too?

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,10 @@
 # Make '_' a no-op so we can scrape strings
 def _(text):
     return text
+
+
+def ngettext(text_singular, text_plural, n):
+    if n == 1:
+        return text_singular
+    else:
+        return text_plural
```
mtyaka: Nit: missing newline at end of file.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -885,17 +883,16 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+        return Object.keys(state.items).length > 0 && attemptsRemain();
     };
 
     var canReset = function() {
         return Object.keys(state.items).length > 0 &&
-            (
-                configuration.mode != DragAndDropBlock.ASSESSMENT_MODE ||
-                canSubmitAttempt()
-            )
+            (configuration.mode != DragAndDropBlock.ASSESSMENT_MODE || attemptsRemain())
```
mtyaka: Nit: I think we usually prefer `!==` instead of `!=`.

mtyaka:
@e-kolpakov I noticed a problem when refreshing the page after you've dragged some items to the board. The items that are in correct positions are disabled (faded out) - they shouldn't be.

Also I think it would be good if users were required to do some changes before they are allowed to submit the problem again. Currently you can keep clicking the submit button and it will keep submitting the identical solution. I think in some other problem types (such as problem builder MRQs) we require the user to change their answer before they can submit again.
This is probably be out of scope for this task, though.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @mtyaka the problem is slightly deeper, so I've submitted https://tasks.opencraft.com/browse/OC-1829. Won't fix this as part of this PR due to very tight deadline on this one.

drag_and_drop_v2/utils.py
e-kolpakov: @mtyaka Added some more code here in next commit (not yet pushed)

e-kolpakov:
@mtyaka addressed review notes, please review

drag_and_drop_v2/utils.py
```diff
@@ -11,4 +11,18 @@ def ngettext(text_singular, text_plural, n):
     if n == 1:
         return text_singular
     else:
-        return text_plural
\ No newline at end of file
+        return text_plural
+
+
+class FeedbackMessages(object):
+    FINAL_ATTEMPT_TPL = _('Final attempt was used, final score is {score}')
```
mtyaka: @e-kolpakov These strings will get translated too early, see: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html#translating-too-early
One possible way around that would be to change these constants to functions (ie. `FeedbackMessages.final_attempt(score)`, `FeedbackMessages.correctly_placed(correct_count)`, etc).

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -867,6 +872,8 @@ function DragAndDropBlock(runtime, element, configuration) {
 
     var doAttempt = function(evt) {
         evt.preventDefault();
+        state.submit_spinner = true;
```
mtyaka: Nit: can we rename this to `state.submit_xhr_active`?

tests/integration/test_interaction.py
```diff
@@ -384,6 +385,23 @@ def _get_scenario_xml(self):  # pylint: disable=no-self-use
...
+    def _wait_until_enabled(element):
+        wait = WebDriverWait(element, 2)
+        wait.until(lambda e: e.is_displayed() and e.get_attribute('disabled') is None)
+
+    def click_submit(self):
+        submit_button = self._get_submit_button()
+
+        self._wait_until_enabled(submit_button)
+
+        submit_button.click()
```
mtyaka: Nit: the two empty lines are probably not needed

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -139,6 +139,12 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         default=False,
     )
 
+    max_grade = Float(
```
mtyaka: Can we rename this to just `grade`? `max_grade` makes it sound like a setting for maximum possible grade (what we currently use `weight` for). In fact I think CAPA problems use `max_grade` field to mean "maximum possible grade", not "maximum achieved grade".

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -474,12 +480,15 @@ def _mark_complete_and_publish_grade(self):
...
+            if grade > self.max_grade:
+                self.max_grade = grade
+                self._publish_grade(grade)
 
-    def _publish_grade(self):
+    def _publish_grade(self, grade):
         try:
             self.runtime.publish(self, 'grade', {
-                'value': self._get_grade(),
+                'value': grade,
```
mtyaka: Nit: will we ever want to publish a grade different from `self.max_grade`, ie. instead of the `_publish_grade` method taking a `grade` parameter, can it just publish `self.max_grade`?

tests/unit/test_advanced.py
```diff
@@ -272,7 +272,7 @@ def test_do_attempt_post_correct_no_publish_grade(self):
         self._reset_problem()
 
         with mock.patch('workbench.runtime.WorkbenchRuntime.publish', mock.Mock()) as patched_publish:
-            self.call_handler(self.DO_ATTEMPT_HANDLER, data={})
+            res = self.call_handler(self.DO_ATTEMPT_HANDLER, data={})
```
mtyaka: Nit: it doesn't look like we ever use the `res` variable.

tests/unit/test_advanced.py
```diff
@@ -404,17 +427,37 @@ def test_do_attempt_feedback_correct_and_decoy(self):
...
+        # make it a last attempt so we can check feedback
+        self.block.max_attempts = 5
+        self.block.num_attempts = 4
+
+        self._submit_solution({0: self.ZONE_1})  # partial solution, 0.33 score
+        res = self.call_handler(self.DO_ATTEMPT_HANDLER, data={})
+        self.assertEqual(self.block.max_grade, expected_score)
+
+        expected_feedback = FeedbackMessages.FINAL_ATTEMPT_TPL.format(score=expected_score)
+        self.assertIn(expected_feedback, res['feedback'])
```
mtyaka: Nit: no newline at end of file

mtyaka:
@e-kolpakov This problem mentioned above is still present:

> I noticed a problem when refreshing the page after you've dragged some items to the board. The items that are in correct positions are disabled (faded out) - they should NOT be disabled.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -416,6 +377,50 @@ def _get_ngettext(self):
...
+            ngettext(
+                FeedbackMessages.MISPLACED_SINGULAR_TPL,
+                FeedbackMessages.MISPLACED_PLURAL_TPL,
+                misplaced_count
+            ).format(misplaced_count=misplaced_count),
+            ngettext(
+                FeedbackMessages.NOT_PLACED_REQUIRED_SINGULAR_TPL,
+                FeedbackMessages.NOT_PLACED_REQUIRED_PLURAL_TPL,
+                missing_count
+            ).format(missing_count=missing_count)
```
mtyaka: @e-kolpakov We might want to reword this message, or maybe just remove it when `missing_count == 0`, because "Not placed 0 required items" is more confusing than it is helpful.

e-kolpakov:
@mtyaka the problem mentioned above should be fixed in 851b54d

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -374,6 +374,14 @@ def attemps_remain(self):
     def get_user_state(self, request, suffix=''):
         """ GET all user-specific data, and any applicable feedback """
         data = self._get_user_state()
+
+        # In assessment mode, if item is placed correctly and than the page is refreshed, "correct"
+        # will spill to the frontend, making item "disabled", thus allowing students to obtain answer by trial
+        # and error + refreshing the page. In order to avoid that, we remove "correct" from an item here
+        if self.mode == self.ASSESSMENT_MODE:
+            for item_id in data['items']:
+                del data['items'][item_id]["correct"]
```
mtyaka: @e-kolpakov Nice catch! I think we could just move this code into `_get_user_state`, though.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -562,9 +570,12 @@ def _get_user_state(self):
 
     def _get_item_state(self):
         """
-        Returns the user item state.
+        Returns a copy of the user item state.
         Converts to a dict if data is stored in legacy tuple form.
         """
+
+        # IMPORTANT: this method should always return a COPY of self.item_state - it is called from get_user_state
+        # handler and manipulated there to hide correctness of items placed
```
mtyaka: I think we should just move the code that removes `'correct'` entries in assessment mode into this method, as mentioned in a comment above. We can remove this comment then.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -405,12 +366,194 @@ def default_background_image_url(self):
...
+            feedback_msgs.append(
+                ngettext(
+                    FeedbackMessages.NOT_PLACED_REQUIRED_SINGULAR_TPL,
+                    FeedbackMessages.NOT_PLACED_REQUIRED_PLURAL_TPL,
+                    missing_count
+                ).format(missing_count=missing_count)
+            )
+
+        if misplaced_ids and self.attemps_remain:
+            feedback_msgs.append(FeedbackMessages.MISPLACED_ITEMS_RETURNED)
```
mtyaka: The message should be wrapped in `gettext` to get translated.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -405,12 +366,194 @@ def default_background_image_url(self):
...
+
+        if not misplaced_ids and not missing_ids:
+            feedback_msgs.append(self.data['feedback']['finish'])
+
+        if not self.attemps_remain:
+            feedback_msgs.append(
+                FeedbackMessages.FINAL_ATTEMPT_TPL.format(score=self.grade))
```
mtyaka: The message should be wrapped in `gettext` for translation.

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,24 @@
...
+    else:
+        return text_plural
+
+
+class FeedbackMessages(object):
+    FINAL_ATTEMPT_TPL = _('Final attempt was used, highest score is {score}')
+    MISPLACED_ITEMS_RETURNED = _('Misplaced items were returned to item bank.')
+
+    CORRECTLY_PLACED_SINGULAR_TPL = _('Correctly placed {correct_count} item.')
+    CORRECTLY_PLACED_PLURAL_TPL = _('Correctly placed {correct_count} items.')
```
mtyaka: @e-kolpakov Sorry for not catching this earlier, but I think these need to be marked with `ngettext` for gettext utils to extract the strings into `po` files properly. Strings used inside `ngettext` are represented differently from regular `gettext` strings inside `po` files, so wrapping singular and plural strings into `_()` individually will not produce correct entries in `po` files.

I am not sure it is possible to represent strings used inside ngettext as constants due to problems with too early translations: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html#translating-too-early

mtyaka:
@e-kolpakov I now get this error when I refresh the page:

```
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 125, in get_response
    response = middleware_method(request, callback, callback_args, callback_kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/debug_toolbar/middleware.py", line 82, in process_view
    response = panel.process_view(request, view_func, view_args, view_kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/debug_toolbar/panels/profiling.py", line 134, in process_view
    return self.profiler.runcall(view_func, *args, **view_kwargs)
  File "/usr/lib/python2.7/cProfile.py", line 149, in runcall
    return func(*args, **kw)
  File "/edx/app/edxapp/edx-platform/lms/djangoapps/courseware/module_render.py", line 954, in handle_xblock_callback
    return _invoke_xblock_handler(request, course_id, usage_id, handler, suffix, course=course)
  File "/edx/app/edxapp/edx-platform/lms/djangoapps/courseware/module_render.py", line 1060, in _invoke_xblock_handler
    resp = instance.handle(handler, req, suffix)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/xblock/mixins.py", line 89, in handle
    return self.runtime.handle(self, handler_name, request, suffix)
  File "/edx/app/edxapp/edx-platform/common/lib/xmodule/xmodule/x_module.py", line 1335, in handle
    return super(MetricsMixin, self).handle(block, handler_name, request, suffix=suffix)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/xblock/runtime.py", line 1025, in handle
    results = handler(request, suffix)
  File "/edx/xblock-drag-and-drop-v2/drag_and_drop_v2/drag_and_drop_v2.py", line 376, in get_user_state
    data = self._get_user_state()
  File "/edx/xblock-drag-and-drop-v2/drag_and_drop_v2/drag_and_drop_v2.py", line 575, in _get_user_state
    is_finished = self._is_finished()
  File "/edx/xblock-drag-and-drop-v2/drag_and_drop_v2/drag_and_drop_v2.py", line 683, in _is_finished
    correct_count, required_count = self._get_item_stats()
  File "/edx/xblock-drag-and-drop-v2/drag_and_drop_v2/drag_and_drop_v2.py", line 654, in _get_item_stats
    required_items, __, correct_items = self._get_item_raw_stats()
  File "/edx/xblock-drag-and-drop-v2/drag_and_drop_v2/drag_and_drop_v2.py", line 667, in _get_item_raw_stats
    correct_items = [item_id for item_id in placed_items if item_state[item_id]['correct']]
KeyError: 'correct'
```

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @mtyaka No, we can't - this same code is used in get_user_stats that actually needs "correct"

drag_and_drop_v2/utils.py
e-kolpakov: @mtyaka How do you generate those po files for drag and drop?

drag_and_drop_v2/drag_and_drop_v2.py
mtyaka: Ah yes, sorry. I think I had a different branch checked out while looking at the functions.

drag_and_drop_v2/utils.py
mtyaka: @e-kolpakov Currently we just assemble them manually, but the plan is to eventually build some generic xblock gettext-based cli tools that will scan xblock packages and create appropriate po files.

e-kolpakov: @mtyaka Have someone actually tried switching to other locale in devstack?

mtyaka: @e-kolpakov I tried and it didn't work. Check OC-1802 for some more info. Getting translations to work is completely out of scope of this PR, I think we should just mark the strings for extraction correctly, so that when XBlock i18n cli tools are developed and state of XBlock i18n support improved, we will be ready to go.

e-kolpakov: @mtyaka Also, what about [all this stuff](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/drag_and_drop_v2.py#L38-L134)? Looks like it is also translated too early, as strings are not updated even when I switch to other locale in studio: 

image

Note save and cancel buttons have changed captions, while all XBlock fields have not.

e-kolpakov: @mtyaka Created a follow-up task https://tasks.opencraft.com/browse/OC-1831 and reworked so that ngettext are actually translated as late as possible, thorugh XBlock's `i18n` service.

Unfortunately, I was not able to make translations work in devstack (and looks like they aren't working in sandboxes too), so I really can't check if it actually helps. On the other hand, looks like entire XBlock translation thing is broken.

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
```diff
@@ -286,10 +286,6 @@ msgid ""
 msgstr ""
 
 #: templates/html/js_templates.html
-msgid "Zones"
```
e-kolpakov: @mtyaka @itsjeyd @haikuginger `msgfmt` considers multiple occurence of same `msgid` as a fatal error. So, here's the question - have you actually tried switching to a locale? If so, how did you do that,  and did it work for you?

e-kolpakov:
@mtyaka Fixed two outstanding issues (`KeyError: 'correct'` exception) and translations - are we ready to go upstream yet?

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
mtyaka: I tried it once, but it didn't seem to work. I attributed it to lack of full i18n support in xblocks.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -403,31 +413,13 @@ def _get_do_attempt_feedback(self):
...
 
         if misplaced_count > 0:
-            feedback_msgs.append(
-                ngettext(
-                    FeedbackMessages.MISPLACED_SINGULAR_TPL,
-                    FeedbackMessages.MISPLACED_PLURAL_TPL,
-                    misplaced_count
-                ).format(misplaced_count=misplaced_count)
-            )
+            feedback_msgs.append(FeedbackMessages.misplaced(missing_count, ngettext))
```
mtyaka: @e-kolpakov This should be `misplaced_count`, not `missing_count`.

mtyaka:
@e-kolpakov Thanks for all the updates! It's looking really good apart from `missing_count`/`misplace_count` typo.

Now that the PR has been rebased, I noticed one final problem:
After you run out of attempts, the items are faded out and you cannot moved them anymore (which is the expected behavior), but if you reload the page, the items are no longer faded out and you can drag them around (although the drops aren't getting registered).

👍 Once these two issues are fixed.

- [x] I tested:
  - In assesment mode: I tested the new persistence of dropped items, grading, feedback messages.
  - In standard mode: Tested to ensure everything still works correctly.
- [x] I read through the code
- [x] ~~Includes documentation~~ (documentation will be added after the assessment functionality is complete)

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
mtyaka: @e-kolpakov Thanks for cleaning up the translations! How did you do that? Did you use gettext CLI tools to extract the strings? It would be great if you could document the commands you used to do that somewhere.

drag_and_drop_v2/utils.py
bradenmacdonald: JavaScript `gettext()` currently does not support XBlock translation strings, so i18n of our "Keyboard Help" button is blocked on https://openedx.atlassian.net/browse/WL-414

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,44 @@
...
+            'Misplaced {misplaced_count} item.',
+            'Misplaced {misplaced_count} items.',
+            n
+        ).format(misplaced_count=n)
+
+    @staticmethod
+    def not_placed(n, ngettext=ngettext_fallback):
+        return ngettext(
+            'Not placed {missing_count} required item.',
+            'Not placed {missing_count} required items.',
```
bradenmacdonald: Should be "Did not place {missing_count} required items."

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
e-kolpakov: @mtyaka yes, I've used xgettext to collect all the messages, but than manually added new ones to the file to avoid huge diff. And then fed `text.po` to msgfmt to build compiled files. I'll add the steps to README in a separate PR.

e-kolpakov:
Rebased from 1610e51503e00f0944fa9dee97da142ba5ce44a4

drag_and_drop_v2/translations/en/LC_MESSAGES/text.po
e-kolpakov: @mtyaka submitted https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/90 - please review

e-kolpakov:
@cahrens, @staubina, @edx/edx-accessibility, @sstack22 this PR is ready for code, a11y and product review - please schedule the review.

e-kolpakov:
@cptvitamin ^^ - looks like edx/edx-accessibility does not exist is @edx/accessibility the right one?

cptvitamin:
@e-kolpakov sorry about that. I think you have to be edx staff to use that alias.

haikuginger:
Rebased on [`630a8d7`](https://github.com/edx-solutions/xblock-drag-and-drop-v2/commit/630a8d7f519409c6c7237660557823998d9e9c29).

bradenmacdonald:
(Note: I rebased this from 4a27531245fbf875331a2efc76d221be659ded04 to 94b9601a8754fb529358ea37bd432d68488dfaac since #86 merged, and I asked @haikuginger to add a commit that renames the new `num_attempts` field to `attempts`, based on our findings in SOL-1809 that doing so will make the LMS "Reset Student Attempts" work for this XBlock. This is still ready for review.)

cahrens:
When you press "Submit", focus needs to go to the feedback area (currently going to top of problem?).

cahrens:
The feedback area could use some UX love. Also, I got a congrats message when I used up all attempts, even though I still got the answer wrong.

image

cahrens:
Note that there are test failures.

bradenmacdonald:
Sorry about the test failures @cahrens. This repo has flaky bok choy tests and we have struggled to figure out the source of the flakiness despite several investigations. All the tests are passing 95% of the time on this branch, but some flaky tests are causing the build to fail (different one each time). I'm trying to re-run it, and over the longer term, we're going to try CircleCI instead of Travis in the near future to see if that helps.

bradenmacdonald:
@cahrens We don't have a UX design for the feedback area yet; is that a blocker for merging?

I know it's a long shot, but I'd love to get this into tomorrow's release if possible.

cahrens:
@bradenmacdonald I don't think you would want this in production without doing something to the feedback area. Is there a story for that? Also, don't we want a11y review and all before this goes into production? My understanding is that this component will be marked as fully supported.

bradenmacdonald:
>  I got a congrats message when I used up all attempts, even though I still got the answer wrong.

That not a bug but rather a problem with the example on the sandbox; there is currently only one "Final Feedback" field, and whatever text the author enters there is shown to learners when they get 100% or when they run out of attempts regardless of their final score. So the message should be something more generic, such as an overall explanation of the answer, without reference to the user's attempt.

> I don't think you would want this in production without doing something to the feedback area.

Ok, I'll see if there's a quick fix possible there. I don't have a story for that, but I can create one if we're going to need to pull someone to do a proper UX design.

> Don't we want a11y review and all before this goes into production?

Yes, but we have course teams that need to start building courses with it this month and the a11y team has been unable to commit to getting a review done on time. So we may have to publish it as beta / not fully supported for a couple weeks until we can complete [the a11y review, SOL-1978](https://openedx.atlassian.net/browse/SOL-1978).

README.md
```diff
@@ -126,13 +134,14 @@ potentially, overlap the zones below.
...
-either text or an image. You can define custom success and error
-feedback for each item. The feedback text is displayed in a popup
-after the learner drops the item on a zone - the success feedback is
-shown if the item is dropped on a correct zone, while the error
-feedback is shown when dropping the item on an incorrect drop zone.
+In the final step, you define the background and text color for drag items, as
+well as the drag items themselves. A drag item can contain either text or an
+image. You can define custom success and error feedback for each item. In
+standard mode, the feedback text is displayed in a popup after the learner drops
+the item on a zone - the success feedback is shown if the item is dropped on a
```
cahrens: I'm not seeing any success or failure feedback (except whether or not the item moves back to the bank) in standard mode on this problem:

http://studio-dndv2-sandbox8.opencraft.hosting/container/block-v1:Opencraft+DNDv2+v2.1+type@vertical+block@28ae4b11831b42949d8815c86671424b#

Aren't there supposed to be tooltips that show up?

bradenmacdonald:
^ @sstack22 FYI

README.md
bradenmacdonald: No; the buckets unit demo does not have success/failure feedback defined for any of its items:

screen shot 2016-08-11 at 1 46 51 pm

Try the "[Canada Demo](http://studio-dndv2-sandbox8.opencraft.hosting/container/block-v1:Opencraft+DNDv2+v2.1+type@vertical+block@4efd52dbdad6498ba51bb7809aa12ca1)" and drag "British Columbia" onto somewhere wrong.

sstack22:
@bradenmacdonald , @cahrens  - I've chatted with @cptvitamin -  we can release these updates to production as is, so that course teams can begin to create the problems. He'll be able to review the problem next week, and the a11y improvements can be a follow up. Let's maintain the support-level as fully supported, with the understanding that the a11y issues will be addressed.

cahrens:
I can review the code in this PR tomorrow, but most likely not until the afternoon.

bradenmacdonald:
How's this mockup of an improved Feeback UX? (cribbed from capa again, obviously)

screen shot 2016-08-11 at 2 17 29 pm
screen shot 2016-08-11 at 2 18 01 pm

bradenmacdonald:
Or, alternately, we may want an overall "✔ You completed this problem correctly" / "× You did not complete the problem correctly" message (green/red) and then have the 'Correctly placed {correct_count} item.' / 'Misplaced {misplaced_count} item.' / 'Did not place {missing_count} required item.' listed below?

sstack22:
I like your mockup! Just a note that that feedback (# items placed/correct/incorrect) should persist. The score information is great, I would just move it to beneath the display name of the problem (similar to CAPA, and makes that line of text shorter). We are making adjustments to this meta feedback as part of our CAPA work ([TNL-4940](https://openedx.atlassian.net/browse/TNL-4940))

bradenmacdonald:
Just to expand: "Just a note that that feedback (# items placed/correct/incorrect) should persist." means that the feedback currently disappears when the student refreshes the page, but that should not be the case.

sstack22:
Thanks for clarifying @bradenmacdonald! Any additional feedback is OK to disappear - the correct/incorrect and score need to persist.

For partial credit we use the font awesome asterisk icon I believe, same green as the green check. I can get more details if helpful, having a hard time tracking it down.

So for the mockup you included, I would use the partial credit icon, and have 1 line of text.

e-kolpakov:
@sstack22 

> The score information is great, I would just move it to beneath the display name of the problem

I've decided to not implement this right now, as there are a couple of important questions:

* Should grade always be displayed there (like CAPA) or "appear" when it is available?
* Should we show highest achieved grade, or result of most recent submission? Or both (somehow)?

I could just implemented it as I see fit, but that's a good way to do the work twice, and unfortunately we don't have time for this right now.

Also, it is not that trivial to implement, so I would suggest stop overloading this PR with more features (and diverging from [discovery document](https://docs.google.com/document/d/1rFliZo1vlohJ7cde7KfP5bgEk0QqcvT1VNnUUa9_vjE/edit#heading=h.17dp8vu) and address this as a follow up PR, when we have more time.

CC @bradenmacdonald

e-kolpakov:
@bradenmacdonald @sstack22 @cahrens sandbox updated with the latest code

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -512,17 +740,22 @@ def _get_item_stats(self):
...
+        required_items, __, correct_items = self._get_item_raw_stats()
 
-        required_items = [str(item['id']) for item in all_items if self._get_item_zones(item['id']) != []]
-        placed_items = [item for item in required_items if item in item_state]
-        correct_items = [item for item in placed_items if item_state[item]['correct']]
+        return len(correct_items), len(required_items)
+
+    def _get_item_raw_stats(self):
+        """
+        Returns a 3-tuple containing required, placed and correct items.
```
staubina: Nit: Update docstring to include the Returns and descriptions as described here: 

http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python_guidelines.html#docstrings

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -531,31 +764,27 @@ def _get_grade(self):
         correct_count, required_count = self._get_item_stats()
         return correct_count / float(required_count) * self.weight
 
-    def _is_finished(self):
+    def _answer_correctness(self):
         """
-        All items are at their correct place and a value has been
-        submitted for each item that expects a value.
+        Checks answer correctness:
```
staubina: same as last comment on docstrings

tests/integration/test_interaction.py
```diff
@@ -562,6 +582,81 @@ def test_submit_button_shown(self):
 
         self.assertEqual(submit_button.get_attribute('disabled'), None)
 
+    def test_misplaced_items_returned_to_bank(self):
```
staubina: Nit: Add comments describing the test goals for clarity

e-kolpakov: @staubina Ok, but I would like to point out that most of the tests does not have docstrings and test name is intended to be self-explanatory.

cahrens:
The feedback area looks better-- thank you. At some point, it would be good to sync up with the feedback style being developed for capa (see https://github.com/edx/edx-platform/pull/13149 and https://openedx.atlassian.net/browse/TNL-4881, and take care to look at the most recent screenshots).

I also wonder about focus not going to the feedback area, but see that you have an aria-live region on the feedback section. So I think it's fine to wait and see what the a11y review produces.

I'm going to review the code now.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -127,7 +143,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         default={},
     )
 
-    num_attempts = Integer(
+    attempts = Integer(
```
cahrens: This has never been released previously, correct? Just want to make sure there is no data migration issue.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -294,107 +322,81 @@ def studio_submit(self, submissions, suffix=''):
...
-            try:
-                self.runtime.publish(self, 'grade', {
-                    'value': self._get_grade(),
-                    'max_value': self.weight,
-                })
-            except NotImplementedError:
-                # Note, this publish method is unimplemented in Studio runtimes,
-                # so we have to figure that we're running in Studio for now
-                pass
+        Raises 400 error in standard mode.
```
cahrens: Nit: might also be worth mentioning that it raises a 409 error if max attempts are exhausted.

e-kolpakov:
@cahrens

> I also wonder about focus not going to the feedback area

We've had that [discussion already](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/86#issuecomment-236660358) - I've got an impression the idea was to wait on a11y to weigh in on this, so I haven't modified this in this PR (despite that line being marked as changed - I've just rearranged the code around it)

tests/integration/test_interaction.py
staubina: @e-kolpakov I would not block the PR for this, it is more a preference than anything.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -412,12 +414,224 @@ def default_background_image_url(self):
...
+
+        item = self._get_item_definition(item_attempt['val'])
+
+        is_correct = self._is_attempt_correct(item_attempt)
+        # State is always updated in assessment mode to store intermediate item positions
+        self.item_state[str(item['id'])] = self._make_state_from_attempt(item_attempt, is_correct)
+
+        self._publish_item_dropped_event(item_attempt, is_correct)
+
+        return {}
```
cahrens: Why does this method return an empty dict, while _drop_item_standard returns the correct and feedback data? They are both called by drop_item- what is the contract for what that method should return?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -412,12 +414,224 @@ def default_background_image_url(self):
...
+        return attempt['zone'] in correct_zones
+
+    def _expand_static_url(self, url):
+        """
+        This is required to make URLs like '/static/dnd-test-image.png' work (note: that is the
+        only portable URL format for static files that works across export/import and reruns).
+        This method is unfortunately a bit hackish since XBlock does not provide a low-level API
+        for this.
+        """
+        if hasattr(self.runtime, 'replace_urls'):
```
cahrens: This method is pretty complicated. Hopefully there is test coverage for it (haven't gotten to the tests yet).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -436,24 +650,38 @@ def _get_user_state(self):
                 else:
                     item['zone'] = 'unknown'
 
-        is_finished = self._is_finished()
+            # In assessment mode, if item is placed correctly and than the page is refreshed, "correct"
```
cahrens: Very helpful comment.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @cahrens In standard mode, when item is dropped into a zone, learner get feedback (both per-item and overall) immediately.

In assessment mode, when an item is dropped into a zone, learner get no feedback at all. Instead feedback is displayed after an attempt is submitted, which is handled by `do_attempt`.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -531,31 +770,29 @@ def _get_grade(self):
...
+            return self.SOLUTION_PARTIAL
 
-    def _get_unique_id(self):
-        usage_id = self.scope_ids.usage_id
-        try:
-            return usage_id.name
-        except AttributeError:
-            # workaround for xblock workbench
-            return usage_id
+    def _is_finished(self):
```
cahrens: I actually found it confusing above (line 663) that a problem is not considered "finished" in assessment mode unless all attempts are exhausted, even if the answer is correct. If the desire is truly to have these different concepts of "finished", it would be helpful to rename this method-- _is_finished_standard_mode or something similar.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @cahrens (1) yes, there is; both unit and integration. (2) it is not modified at all - I've just moved it from [there](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/87/files/5b837e7e1ccde62cc51c8c1805fe056af5b287de#diff-81cf2d60a06d8f43d0df802d9aad9417L373)

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -284,6 +284,57 @@
...
+    color: #b20610;
+    border: 2px solid #b20610;
+}
+
+
+/* Font Awesome icons have different width - margin-right tries to compensate it */
+.xblock--drag-and-drop .feedback p:before {
+    content: "\f129";
+    font-family: FontAwesome;
+    margin-right: 0.7em;
```
cahrens: Have you tested that this works for RTL languages?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -227,12 +227,21 @@ function DragAndDropTemplates(configuration) {
...
-        var feedback_display = ctx.feedback_html ? 'block' : 'none';
         var properties = { attributes: { 'aria-live': 'polite' } };
+        var messages = ctx.overall_feedback_messages || [];
+        var feedback_display = messages.length > 0 ? 'block' : 'none';
+        var feedback_messages = messages.map(function(message) {
+            var selector = "p.message";
+            if (message.message_class) {
+                selector += "."+message.message_class;
+            }
+            return h(selector, {innerHTML: message.message}, []);
```
cahrens: Are the messages getting escaped? We have to protect against possibly nefarious translations as well (and I'm not yet sure where the messages are ultimately coming from).

e-kolpakov: @cahrens No, and they should be, because those messages are coming from DnDv2 author and may contain plain HTML. This is a feature.

drag_and_drop_v2/public/css/drag_and_drop.css
e-kolpakov: @cahrens no, but I would assume it won't work as I intended. Can we address that as part of a11y follow up?

drag_and_drop_v2/public/js/drag_and_drop.js
cahrens: Do you mean that they should not be escaped? @rraposa do we have any best practices around features where course staff might want to specify HTML?

e-kolpakov: @cahrens The basic idea is that author can use HTML markup in feedback messages, and it should not be escaped (yes, I've missed "not" in previous comment). It does not need to work exactly like this, but this is how it worked previously - I haven't changed the approach, only the code to implement it.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @cahrens As far as I can tell - no, it has never been released.

drag_and_drop_v2/public/css/drag_and_drop.css
cahrens: Can you create a story for a UX review and testing of RTL? Really RTL is something you should be thinking about (and testing) when you first write the CSS, but I understand the desire to get something out quickly for MIT to start working with.

drag_and_drop_v2/public/js/drag_and_drop.js
cahrens: Note also that some of these message are not expected to have HTML-- for example, the messages in FeedbackMessages. It is worth spending some time thinking about what can be escaped; this is an area where we've been notoriously sloppy and open to vulnerabilities. And since this code is in a separate repo, it doesn't benefit from the safe linting checks in edx-platform.

tests/integration/test_interaction.py
```diff
@@ -562,6 +582,90 @@ def test_submit_button_shown(self):
...
+        for item_id in misplaced_items:
+            self.assert_reverted_item(item_id)
+
+    def test_max_attempts_reached_submit_and_reset_disabled(self):
+        """
+        Test "Submit" and "Reset" buttons are disabled when no more attempts remaining
+        """
+        self.place_item(0, TOP_ZONE_ID)
+
+        for _ in xrange(self.MAX_ATTEMPTS):
```
cahrens: It would be nice to verify that the reset button is not disabled until MAX_ATTEMPTS is reached (I assume it would not be possible to click on a disabled submit button).

tests/unit/test_advanced.py
```diff
@@ -150,22 +161,220 @@ def test_do_attempt_final(self):
...
+
+    def _submit_solution(self, solution):
+        for item_id, zone_id in solution.iteritems():
+            data = self._make_submission(item_id, zone_id)
+            self.call_handler(self.DROP_ITEM_HANDLER, data)
+
+    def _submit_complete_solution(self):
+        self._submit_solution(self.CORRECT_SOLUTION)
+
+    def _submit_partial_solution(self):  # pylint: disable=no-self-use
```
cahrens: Why is this here (with NotImplementedError)?

cahrens:
I am done reviewing. I have a couple suggestions above, but nothing that I would consider blocking (for the purposes of getting something out that MIT can start working with).

But in addition to the a11y review that will be done next week, there also needs to be a ticket for UX review and testing/bug-fixing of RTL mode-- please include a link to that ticket.

With those caveats, 👍

drag_and_drop_v2/drag_and_drop_v2.py
bradenmacdonald: Correct - this field was introduced in the previous PR, #86, and has never been released.

drag_and_drop_v2/public/js/drag_and_drop.js
robrap: @cahrens: As you noted, we were terribly sloppy in the past and basically everything allowed HTML (and thus was vulnerable to XSS).  Many people used this as a feature in certain areas.  Now, we simply need to be more intentional and determine whether HTML is an acceptable input in certain places.  Documentation should probably be updated to reflect the intentional decision where HTML is accepted.

Additionally, where HTML is accepted, I guess we could further determine if limited HTML is accepted or any HTML.  @nedbat?

tests/unit/test_advanced.py
e-kolpakov: @cahrens This class is intended to contain assessment mode tests that does not depend on concrete items/zones configuration. For some tests, we need a partial solution to be submitted, but don't actually care about what items were submitted or not.

That was a motivation. From technical perspective, I want `AssessmentModeFixture` be an abstract class, while keeping as much common tests as possible. In that case, it allows to introduce more data-tied tests easier. So far we've got only one assessment test suite: `TestDragAndDropAssessmentData`, but with `AssessmentModeFixture` it would be easier to add more (similar to `TestDragAndDropHtmlData` - `TestDragAndDropPlainData` - `TestOldDataFormat`).

To make it more "simmetrical", I've made `_submit_correct_solution` raise `NotImplementedError` as well.

e-kolpakov:
@cahrens rebased from 5269326e1f5e98515459a40483e45719c3459cb2

e-kolpakov:
@sstack22 @mtyaka ^^

e-kolpakov:
@cahrens created https://openedx.atlassian.net/browse/SOL-2030 for UX follow up. I've logged a ticket in Opnecraft internal tracker for RTL fixes, I just want acceptance criteria and wording to be reviewed and/or polished by @bradenmacdonald first.

e-kolpakov:
@sstack22 I believe this PR just waits for your approval now - could you please review it today (no pressure, but we're significantly overdue here) and ping me so I could push it forward? Thank you!

pdesjardins:
@bradenmacdonald, for documentation planning, is it possible that this will merge into production  in this week's release (August 15th)?

e-kolpakov:
@pdesjardins Looks like Braden is not available yet, so I'll try to answer to the best of my knowledge.

As far as I know, the plan is to include this changeset into today's release. This could happen if @sstack22 approves the PR today; today = 3-4 hours from now for me, and probably something like 8-10 hours for Braden (obviously, he knows better). If it happens, we'll merge this PR and push a version bump to release candidate branch.

sstack22:
@e-kolpakov, @bradenmacdonald  - 👍 on my end. This is great work! I'll comment on the follow-up UX ticket to iron out those future adjustments.

e-kolpakov:
@sstack22 Thank you - merging!

PR #88, opened 2016-08-04
Clean up studio edit template

mtyaka:
**Description**:

This PR cleans up the Studio edit template and associated JS code to:

- Use `field.display_name` values for field labels instead of hardcoding/duplicating the names in the template.
- Directly display help text under associated input field instead of `title` and `<span class="sr">`.
- Replace usage of HTML `id` attributes with `class`-es. In cases where that is not possible, ensure that the `id` attributes are globally unique.
- Use `<button>` elements instead of `<a>` elements for buttons.
- Replace input placeholders with help texts visible to everyone.

**Screenshots**:

Before:

screen shot 2016-08-04 at 17 53 48

After:

screen shot 2016-08-04 at 17 52 52

**JIRA ticket**: https://openedx.atlassian.net/browse/SOL-1990

**Dependencies**: None

**Discussions**: https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/77#discussion_r71034788

**Sandbox URLs** (installed version 9845083078d129009933cd8c8106b27c14e17242):

- LMS: http://dndv2-sandbox5.opencraft.hosting/
- Studio: http://studio-dndv2-sandbox5.opencraft.hosting/

You are probably mostly interested in the Studio edit functionality. Some DnDv2 blocks can be edited at: http://studio-dndv2-sandbox5.opencraft.hosting/container/block-v1:OpenCraftX+OC1+1+type@vertical+block@e3c62b19090948b8ad89d3386bff4435

**Testing Instructions**:

This PR does not implement any new functionality, so there is no specific feature to test.
The important thing is to ensure that the studio edit functionality continues to look and function correctly.

**Reviewers**:

- [x] OpenCraft: @bradenmacdonald
- [ ] ~~a11y: @edx/edx-accessibility~~ Scheduled combined review for all DnDv2.1 stories: [SOL-1978](https://openedx.atlassian.net/browse/SOL-1978)
- [x] Product: @sstack22

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -255,19 +255,10 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         # connect 'for' and 'aria-describedby' attributes to the associated elements.
         id_suffix = uuid.uuid4().hex[:16]
         js_templates = js_templates.replace('{{id_suffix}}', id_suffix)
-        help_texts = {
-            field_name: self.ugettext(field.help)
-            for field_name, field in self.fields.viewitems() if hasattr(field, "help")
-        }
-        field_values = {
-            field_name: field.values
-            for field_name, field in self.fields.viewitems() if hasattr(field, "values")
-        }
         context = {
             'js_templates': js_templates,
-            'help_texts': help_texts,
-            'field_values': field_values,
             'id_suffix': id_suffix,
+            'fields': self.fields,
             'self': self,
             'data': urllib.quote(json.dumps(self.data)),
         }
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -115,14 +115,14 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
     item_background_color = String(
         display_name=_("Item background color"),
-        help=_("The background color of draggable items in the problem (e.g. 'blue' or '#0000ff')."),
+        help=_("The background color of draggable items in the problem (example: 'blue' or '#0000ff')."),
         scope=Scope.settings,
         default="",
     )
 
     item_text_color = String(
         display_name=_("Item text color"),
-        help=_("Text color to use for draggable items (e.g. 'white' or '#ffffff')."),
+        help=_("Text color to use for draggable items (example: 'white' or '#ffffff')."),
         scope=Scope.settings,
         default="",
     )
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -106,7 +106,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     )
 
     weight = Float(
-        display_name=_("Weight"),
+        display_name=_("Maximum score"),
         help=_("The maximum score the learner can receive for the problem"),
         scope=Scope.settings,
         default=1,
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
 #
 """ Drag and Drop v2 XBlock """
+
 # Imports ###########################################################
 
 import copy
 import json
 import urllib
+import uuid
 import webob
 
 from xblock.core import XBlock
@@ -247,6 +249,12 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
 
         js_templates = loader.load_unicode('/templates/html/js_templates.html')
+        # A short random string to append to HTML element ID attributes to avoid multiple instances
+        # of the DnDv2 block on the same page sharing the same ID values.
+        # We avoid using ID attributes in preference to classes, but sometimes we still need IDs to
+        # connect 'for' and 'aria-describedby' attributes to the associated elements.
+        id_suffix = uuid.uuid4().hex[:16]
+        js_templates = js_templates.replace('{{id_suffix}}', id_suffix)
         help_texts = {
             field_name: self.ugettext(field.help)
             for field_name, field in self.fields.viewitems() if hasattr(field, "help")
@@ -259,6 +267,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
             'js_templates': js_templates,
             'help_texts': help_texts,
             'field_values': field_values,
+            'id_suffix': id_suffix,
             'self': self,
             'data': urllib.quote(json.dumps(self.data)),
         }
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -7,7 +7,6 @@
 import copy
 import json
 import urllib
-import uuid
 import webob
 
 from xblock.core import XBlock
@@ -95,7 +94,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
     question_text = String(
         display_name=_("Problem text"),
-        help=_("The description of the problem or instructions shown to the learner"),
+        help=_("The description of the problem or instructions shown to the learner."),
         scope=Scope.settings,
         default="",
     )
@@ -109,21 +108,21 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
     weight = Float(
         display_name=_("Maximum score"),
-        help=_("The maximum score the learner can receive for the problem"),
+        help=_("The maximum score the learner can receive for the problem."),
         scope=Scope.settings,
         default=1,
     )
 
     item_background_color = String(
         display_name=_("Item background color"),
-        help=_("The background color of draggable items in the problem."),
+        help=_("The background color of draggable items in the problem (e.g. 'blue' or '#0000ff')."),
         scope=Scope.settings,
         default="",
     )
 
     item_text_color = String(
         display_name=_("Item text color"),
-        help=_("Text color to use for draggable items."),
+        help=_("Text color to use for draggable items (e.g. 'white' or '#ffffff')."),
         scope=Scope.settings,
         default="",
     )
@@ -249,11 +248,12 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
 
         js_templates = loader.load_unicode('/templates/html/js_templates.html')
-        # A short random string to append to HTML element ID attributes to avoid multiple instances
-        # of the DnDv2 block on the same page sharing the same ID values.
+        # Get a 'html_id' string that is unique for this block.
+        # We append it to HTML element ID attributes to ensure multiple instances of the DnDv2 block
+        # on the same page don't share the same ID value.
         # We avoid using ID attributes in preference to classes, but sometimes we still need IDs to
         # connect 'for' and 'aria-describedby' attributes to the associated elements.
-        id_suffix = uuid.uuid4().hex[:16]
+        id_suffix = self.location.html_id()
         js_templates = js_templates.replace('{{id_suffix}}', id_suffix)
         context = {
             'js_templates': js_templates,
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -253,7 +253,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         # on the same page don't share the same ID value.
         # We avoid using ID attributes in preference to classes, but sometimes we still need IDs to
         # connect 'for' and 'aria-describedby' attributes to the associated elements.
-        id_suffix = self.location.html_id()
+        id_suffix = self.location.html_id()  # pylint: disable=no-member
         js_templates = js_templates.replace('{{id_suffix}}', id_suffix)
         context = {
             'js_templates': js_templates,
--- a/drag_and_drop_v2/public/css/drag_and_drop_edit.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop_edit.css
@@ -68,15 +68,19 @@
   clear: both;
 }
 
-.xblock--drag-and-drop--editor .tab h3,
-.xblock--drag-and-drop--editor .tab .h3 {
+.xblock--drag-and-drop--editor .tab h3 {
+    font-size: 18px;
+}
+
+.xblock--drag-and-drop--editor .tab h4,
+.xblock--drag-and-drop--editor .tab .h4 {
+    display: block;
+    font-size: 16px;
     margin: 20px 0 8px 0;
 }
 
-.xblock--drag-and-drop--editor .tab .h3 {
-    display: block;
-    font: inherit;
-    font-size: 100%;
+.xblock--drag-and-drop--editor .tab .items-form .item .row:first-of-type .h4 {
+    margin-top: 0;
 }
 
 .xblock--drag-and-drop--editor .tab-header,
@@ -98,18 +102,35 @@
     margin: 10px 0 0 0;
 }
 
-.xblock--drag-and-drop--editor .target-image-form input,
-.xblock--drag-and-drop--editor .target-image-form textarea {
+.xblock--drag-and-drop--editor .target-image-form input {
     width: 50%;
+}
+.xblock--drag-and-drop--editor .target-image-form textarea {
+    width: 97%;
 }
 
 .xblock--drag-and-drop--editor .target-image-form textarea {
     display: block;
 }
 
+.xblock--drag-and-drop--editor input,
+.xblock--drag-and-drop--editor textarea {
+    box-sizing: border-box;
+    font-size: 14px;
+    background: #fff;
+    box-shadow: none;
+    padding: 6px 8px;
+    border: 1px solid #b2b2b2;
+    color: #4c4c4c;
+}
+
 .xblock--drag-and-drop--editor label > span {
     display: inline-block;
     margin-bottom: 0.25em;
+}
+
+.xblock--drag-and-drop--editor label.checkbox-label {
+    font-size: 14px;
 }
 
 /* Main Tab */
@@ -151,29 +172,36 @@
   max-width: 100%;
 }
 
+.xblock--drag-and-drop--editor .zones-form .zone-row {
+    background-color: #b1d9f1;
+    padding: 1rem;
+    margin-bottom: 2rem;
+}
+
 .xblock--drag-and-drop--editor .zones-form .zone-row label {
     display: block;
 }
 
 .xblock--drag-and-drop--editor .zones-form .zone-row label > span {
     display: inline-block;
-    width: 6em;
+    font-size: 14px;
+    min-width: 8rem;
 }
 
 .xblock--drag-and-drop--editor .zones-form .zone-row label > input {
-    width: 60%;
+    width: 66%;
     margin: 0 0 5px;
     line-height: 2.664rem;  /* .title gets line-height from a Studio rule that does not apply to .description;
                                here we make sure that both input fields get the same value for line-height */
 }
 
-.xblock--drag-and-drop--editor .zones-form .zone-row .alignment {
-    margin-bottom: 15px;
+.xblock--drag-and-drop--editor .zones-form .zone-row .layout {
+    margin: 2rem 0;
 }
 
 .xblock--drag-and-drop--editor .zones-form .zone-row .layout label {
     display: inline-block;
-    width: 40%;
+    width: 45%;
 }
 
 .xblock--drag-and-drop--editor .zones-form .zone-row .layout .zone-size,
@@ -183,14 +211,19 @@
     line-height: inherit;
 }
 
+.xblock--drag-and-drop--editor .zones-form .zone-row .alignment {
+    margin-bottom: 15px;
+}
+
+
 .xblock--drag-and-drop--editor .feedback-form textarea {
     width: 99%;
     height: 128px;
 }
 
 .xblock--drag-and-drop--editor .form-help {
-    margin-top: 5px;
-    font-size: small;
+    margin: 0.5rem 0 1rem;
+    font-size: 12px;
 }
 
 .xblock--drag-and-drop--editor .item-styles-form,
@@ -199,14 +232,9 @@
 }
 
 .xblock--drag-and-drop--editor .items-form .item {
-    background-color: #8fcaec;
-    padding: 10px 0 1px;
-    margin: 15px 0;
-}
-
-.xblock--drag-and-drop--editor .items-form label,
-.xblock--drag-and-drop--editor .items-form .h3 {
-    margin: 0 1%;
+    background-color: #b1d9f1;
+    padding: 2rem;
+    margin-bottom: 2rem;
 }
 
 .xblock--drag-and-drop--editor .items-form select {
@@ -220,7 +248,6 @@
 .xblock--drag-and-drop--editor .items-form .item-text,
 .xblock--drag-and-drop--editor .items-form .item-image-url {
     width: 50%;
-    margin: 0 1%;
 }
 
 .xblock--drag-and-drop--editor .items-form .item-width {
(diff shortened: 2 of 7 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=88

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -222,6 +242,7 @@
     display: none;
 }
 .xblock--drag-and-drop--editor .items-form .row.advanced-link {
+    cursor: pointer;
```
bradenmacdonald: This is because the `<a>` element has no `href=""` attribute, right? Is it better to do this or use `href="#"` + `preventDefault()` ?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -227,6 +234,7 @@ def studio_view(self, context):
             'js_templates': js_templates,
             'help_texts': help_texts,
             'field_values': field_values,
```
bradenmacdonald: It seems inconsistent to pass the help text and field values as context vars, but to use `self.fields.foo.display_name` to get the display name. Maybe we should just pass `fields` as a context variable and then use `fields.foo.display_name`, `fields.food.help`, and `fields.foo.values`? That would be more consistent and efficient.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -117,20 +149,24 @@
...
-                    <label class="h3" for="item-background-color">{% trans "Background color" %}</label>
-                    <input id="item-background-color"
-                           placeholder="e.g. blue or #0000ff"
-                           value="{{ self.item_background_color}}"
-                           aria-describedby="item-background-color-description">
-                    <div id="item-background-color-description" class="item-styles-form-help">
+                    <label class="h3">
+                        <span>{% trans self.fields.item_background_color.display_name %}</span>
+                        <input class="item-background-color"
+                               placeholder="e.g. blue or #0000ff"
```
bradenmacdonald: Needs i18n. `blue` is a CSS keyword so shouldn't be translated, but "e.g." is English-specific. So this should be marked for i18n as

```
{% blocktrans with example1="blue" example2="#0000ff" %}
e.g. {{example1}} or {{example2}}
{% endblocktrans %}
```

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -117,20 +149,24 @@
...
-                    <label class="h3" for="item-text-color">{% trans "Text color" %}</label>
-                    <input id="item-text-color"
-                           placeholder="e.g. white or #ffffff"
-                           value="{{ self.item_text_color}}"
-                           aria-describedby="item-text-color-description">
-                    <div id="item-text-color-description" class="item-styles-form-help">
+                    <label class="h3">
+                        <span>{% trans self.fields.item_text_color.display_name %}</span>
+                        <input class="item-text-color"
+                               placeholder="e.g. white or #ffffff"
```
bradenmacdonald: Also needs i18n

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -9,80 +9,84 @@
...
-        <div class="layout">
-            <label for="zone-{{index}}-width">{{i18n "width"}}</label>
+        <label>
+            <span>{{i18n "Text"}}</span>
             <input type="text"
-                   id="zone-{{index}}-width"
-                   class="size width"
-                   value="{{ zone.width }}" />
-            <label for="zone-{{index}}-height">{{i18n "height"}}</label>
+                   class="title"
```
bradenmacdonald: This class name (`title`) is too generic; Studio already defines a `[class*="view-"] .modal-window .title` rule which is affecting this. I'd suggest `zone-title` instead.

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -9,80 +9,84 @@
...
-        <input type="text"
-               id="zone-{{index}}-description"
-               class="description"
-               value="{{ zone.description }}"
-               placeholder="{{i18n 'Describe this zone to non-visual users'}}"
-               required />
-        <div class="layout">
-            <label for="zone-{{index}}-width">{{i18n "width"}}</label>
+        <label>
+            <span>{{i18n "Text"}}</span>
```
bradenmacdonald: Maybe we should change this to "Title" ?

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -9,80 +9,84 @@
...
+                   value="{{ zone.title }}"
+                   required />
+        </label>
+        <label>
+            <span>{{i18n "Description"}}</span>
             <input type="text"
-                   id="zone-{{index}}-height"
-                   class="size height"
-                   value="{{ zone.height }}" />
+                   class="description"
```
bradenmacdonald: This class name is also too generic and likely to be affected by external CSS rules.

bradenmacdonald:
@mtyaka Just a few minor notes/questions - thanks for this!

drag_and_drop_v2/drag_and_drop_v2.py
mtyaka: Good idea, passing `fields` only makes it more consistent; done.

drag_and_drop_v2/public/css/drag_and_drop_edit.css
mtyaka: Yes, that's right. I removed `href="#"` and added this CSS rule. I believe this is considered to be a better approach than `href="#"`, which has always felt like a hack to me.

mtyaka:
@bradenmacdonald Thanks for the review! Your comments have been addressed.

bradenmacdonald:
👍  Nice improvement, @mtyaka.

- [x] I tested this:
   * Made changes to the dnd components on the Sandbox, using the updated Studio editor.
   * I used `?preview-lang=eo` on the sandbox to test some i18n, though the edx-platform is currently only able to translate strings that already exist in the platform and does not yet always load xblock-specific strings.
   * I used chrome's element inspector to copy the rendered outerHTML for each tab of the Studio editor and search for `id="..."` attributes that had non-unique values.
- [x] I read through the code
- [x] Includes documentation: no readme changes required. This PR adds brief documentation where it belongs - inline where the user is making changes.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -215,18 +216,16 @@ def studio_view(self, context):
...
-        }
-        field_values = {
-            field_name: field.values
-            for field_name, field in self.fields.viewitems() if hasattr(field, "values")
-        }
+        # A short random string to append to HTML element ID attributes to avoid multiple instances
+        # of the DnDv2 block on the same page sharing the same ID values.
+        # We avoid using ID attributes in preference to classes, but sometimes we still need IDs to
+        # connect 'for' and 'aria-describedby' attributes to the associated elements.
+        id_suffix = uuid.uuid4().hex[:16]
```
cahrens: Another way of creating unique IDs is to incorporate the XBlock id. That's what we do for capa and ORA.

cahrens:
I would suggest a doc review of the field help text; it's pretty inconsistent in formatting and style. Who is the edX doc person working on Drag and Drop?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -215,18 +216,16 @@ def studio_view(self, context):
...
+        js_templates = js_templates.replace('{{id_suffix}}', id_suffix)
         context = {
```
cahrens: Why are you directly replacing id_suffix on line 224 in js_templates (I'm not sure how those are used) and also passing id_suffix in the context (which is what I would expect)?

bradenmacdonald:
^ @pdesjardins is doing docs for drag and drop

drag_and_drop_v2/public/css/drag_and_drop_edit.css
cahrens: Instead of using an a-tag, you should use a button. A-tag elements should take you someplace, whereas buttons do actions. Buttons give you the correct a11y behavior, whereas you have to manually create those behaviors with other elements using aria properties.

Tabbing through the Studio editing UI, I don't see any way to even get to these "Remove" links with the keyboard.

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -254,14 +275,15 @@
 }
 
 .xblock--drag-and-drop--editor .remove-zone {
+    cursor: pointer;
```
cahrens: Same note here-- don't use a-tags.

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -254,14 +275,15 @@
...
     float: right;
     margin-top: 2px;
     margin-right: 16px;
 }
 
 .xblock--drag-and-drop--editor .remove-item {
+    cursor: pointer;
```
cahrens: Same note here-- don't use a-tags.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -67,15 +89,21 @@
...
                 <form class="target-image-form">
-                    <label class="h3" for="background-url">{% trans "Background URL" %}</label>
-                    <input id="background-url"
+                    <label class="h3" for="background-url-{{id_suffix}}">
+                        <span>{% trans "Background URL" %}</span>
+                    </label>
+                    <input class="background-url"
+                           id="background-url-{{id_suffix}}"
                            type="text"
                            placeholder="{% trans 'For example, http://example.com/background.png or /static/background.png' %}">
```
cahrens: Placeholders should be avoided-- they are too low in contrast, and they don't provide a very good user experience because the text disappears when you start typing. It's better to just add a visible (to all) description.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -87,13 +115,17 @@
             <section class="tab-content">
                 <form class="display-labels-form">
                     <h3>{% trans "Zone labels" %}</h3>
-                    <label for="display-labels">{% trans "Display label names on the image" %}:</label>
-                    <input name="display-labels" id="display-labels" type="checkbox" />
+                    <label>
+                        <span>{% trans "Display label names on the image" %}:</span>
```
cahrens: Include the colon in the text to be translated-- colons won't make sense in all languages. Or is it more natural to have the checkbox before the text, so there is no need for a colon?

cahrens:
I imagine these font sizes (12?) are too small. Hopefully an a11y review can weigh in.

image

cahrens:
There are color contrast errors in this "blue" section.

image

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -117,21 +149,25 @@
...
-                    <input id="item-background-color"
-                           placeholder="e.g. blue or #0000ff"
-                           value="{{ self.item_background_color}}"
-                           aria-describedby="item-background-color-description">
-                    <div id="item-background-color-description" class="item-styles-form-help">
-                      {{ help_texts.item_background_color }}
+                    <label class="h3">
+                        <span>{% trans fields.item_background_color.display_name %}</span>
+                        <input class="item-background-color"
+                               placeholder="{% blocktrans with example1='blue' example2='#0000ff' %}e.g. {{example1}} or {{example2}}{% endblocktrans %}"
```
cahrens: Again, avoid placeholders.

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -9,80 +9,84 @@
...
+        </label>
+        <label>
+            <span>{{i18n "Description"}}</span>
             <input type="text"
-                   id="zone-{{index}}-height"
-                   class="size height"
-                   value="{{ zone.height }}" />
+                   class="zone-description"
+                   value="{{ zone.description }}"
+                   placeholder="{{i18n 'Describe this zone to non-visual users'}}"
```
cahrens: Avoid placeholders.

cahrens:
There are some good improvements here! However, there is still more work to do to make the Studio UI accessible.

I'm sure a full a11y review will find things I have missed, but for now, try accessing everything in the UI just with just the keyboard-- there are several elements that one cannot currently navigate to (specifically, there are many problematic a-tags that are acting as buttons).

staubina:
I will wait to review in depth until updates are made per @cahrens comments

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -87,13 +115,17 @@
...
+                        <span>{% trans "Display label names on the image" %}:</span>
+                        <input name="display-labels" class="display-labels" type="checkbox" />
+                    </label>
                 </form>
                 <form class="display-borders-form">
                     <h3>{% trans "Zone borders" %}</h3>
-                    <label for="display-borders">{% trans "Display zone borders on the image" %}:</label>
-                    <input name="display-borders" id="display-borders" type="checkbox" />
+                    <label>
+                        <span>{% trans "Display zone borders on the image" %}:</span>
```
staubina: Another color take a  look at @cahrens comment above

drag_and_drop_v2/templates/html/js_templates.html
```diff
@@ -126,32 +130,38 @@
...
         </div>
         <div class="row advanced-link">
-            <a href="#">{{i18n "Show advanced settings" }}</a>
+            <a>{{i18n "Show advanced settings" }}</a>
         </div>
         <div class="row advanced">
             <label>
-                {{i18n "Preferred width as a percentage of the background image width (or blank for automatic width):"}}
+                <span>
+                  {{i18n "Preferred width as a percentage of the background image width (or blank for automatic width):"}}
```
staubina: Should we include this colon in the translation?

staubina:
There seems to be a lack of consistency in the font sizes throughout the code. Is this something that could be updated here or in a later PR?

screen shot 2016-08-09 at 8 27 33 am

drag_and_drop_v2/public/css/drag_and_drop_edit.css
mtyaka: @cahrens Thanks, thanks makes sense. I replaced `<a>` elements with `<button>`s.

drag_and_drop_v2/drag_and_drop_v2.py
mtyaka: @cahrens Thanks for the hint. I replaced `uuid.uuid4()` with `block.location.html_id()`. I was not aware of the `html_id()` function before.

drag_and_drop_v2/drag_and_drop_v2.py
mtyaka: @cahrens js_templates contains Handlebars templates. Replacing `id_suffix` with raw string substitutions is a bit of a hack, however I think it's still better than passing the `id_suffix` variable to JavaScript and making sure it is present in the handlebars context every time when rendering a handlebars template.

We still have to pass the `id_suffix` variable in the context, because we use it inside django templates, too.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
mtyaka: @cahrens I think having the checkbox before the text is more natural, so I swapped them and removed the colon.

drag_and_drop_v2/templates/html/js_templates.html
mtyaka: @staubina I changed the way this field is displayed in the last commit to make it more consistent with other fields. The new layout does not need the colon, so I removed it.

mtyaka:
Thanks for the review @cahrens and @staubina!

I pushed an updated that addresses your comments, including making font sizes more consistent. I updated the sandbox with the latest code, please take another look.

Here are some screenshots:

screen shot 2016-08-10 at 13 44 35

screen shot 2016-08-10 at 13 45 06

screen shot 2016-08-10 at 13 45 34

staubina:
👍

cahrens:
This is looking a lot better! Make sure that when a button is clicked, the focus goes someplace sensible. For instance, when clicking "Add an Item", focus should move to the newly created Item.

I also find it odd that when focus is on a button (both "Add an Item" and the Delete buttons), they actually look disabled.

image

cahrens:
Perhaps beyond the scope of this PR, but note that the Continue and Cancel buttons at the bottom of the dialog continue to be a-links. I noticed it because I hit Space on Continue (a button is triggered by both Return and Space), and I was confused because the page scrolled.

cahrens:
Have you tested for RTL? You can do that (on your sandbox) by adding ?preview-lang=rtl to the end of the URL. I see some minor layout issues.

If you'd like a review of the CSS, I can ask the FED person on my team to review. I'm not nearly as good with CSS, and also I'm not sure what sort of standards we try to maintain for code outside of edx-platform (that isn't using the pattern library).

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -208,22 +256,27 @@
 
 .xblock--drag-and-drop--editor .items-form textarea {
     width: 97%;
-    margin: 0 1%;
 }
 
 .xblock--drag-and-drop--editor .items-form .zone-checkbox-label {
     text-align: left;
```
cahrens: Things like this are likely problematic for RTL.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -86,21 +117,29 @@
...
                 </form>
             </section>
             <section class="tab-content">
+                <h4>{% trans "Zone definitions" %}</h4>
                 <div class="zone-editor">
                     <div class="controls">
                         <form class="zones-form"></form>
-                        <a href="#" class="add-zone add-element"><div class="icon add"></div>{% trans "Add a zone" %}</a>
+                        <button class="btn add-zone add-element">
+                            <span class="icon add"></span>
```
cahrens: I'm not sure what type of icon this is, but for Font Awesome icons we always specify aria-hidden (with either visible text like this or screen reader text).

See example HTML here--

mtyaka:
Thanks @cahrens, I addressed most of your comments:

- Focus goes to the new item when clicking the "Add Zone" or "Add Item" button.
- Fix `:hover` and `:focus` styles so that buttons don't look disabled.
- Fixed several RTL layout issues.
- Added `aria-hidden="true"` to icons.

> Perhaps beyond the scope of this PR, but note that the Continue and Cancel buttons at the bottom of the dialog continue to be a-tags.

That button HTML code was copied from the Studio (CAPA problems currently also use `<a>` tags) so that we can use default Studio styles and avoid having to copy/imitate the look using our own CSS. I tried switching to button elements, but that broke the styles, so I would prefer to wait until `<a>` are replaced with `<buttons>` in the Studio, after which we can simply copy the new HTML structure.

> If you'd like a review of the CSS, I can ask the FED person on my team to review.

A review of the CSS would be great, but I think it would be best if it was done outside of this PR (reviewing all CSS, not just the changes in this PR).

cahrens:
I understand about the Continue/Cancel buttons. When Studio was written we incorrectly used a-tags everywhere, and that is something that will be cleaned up later this year.

However, when I tab to the Continue button and press Enter, focus is remaining in the button bar-- I think this is new behavior since yesterday. After Continue, keyboard focus needs to move to the top of the new settings panel. As it is, it is very easy to move keyboard focus outside of the dialog entirely (while tabbing to try to figure out where focus is), and difficult to figure out how to get back.

Also a nit about adding a little padding around this close button:

image

mtyaka:
@cahrens 

> However, when I tab to the Continue button and press Enter, focus is remaining in the button bar-- I think this is new behavior since yesterday. After Continue, keyboard focus needs to move to the top of the new settings panel. As it is, it is very easy to move keyboard focus outside of the dialog entirely (while tabbing to try to figure out where focus is), and difficult to figure out how to get back.

The behavior was not new, but I should have realized and fixed it after your previous comment, sorry I missed that. It's fixed in the latest update.

> Also a nit about adding a little padding around this close button.

Last commit adds a bit of extra padding on small screens.

I updated the sandbox with latest changes.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -90,22 +90,22 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
...
+        display_name=_("Maximum score"),
+        help=_("The maximum score the learner can receive for the problem."),
         scope=Scope.settings,
         default=1,
     )
 
     item_background_color = String(
         display_name=_("Item background color"),
-        help=_("The background color of draggable items in the problem."),
+        help=_("The background color of draggable items in the problem (e.g. 'blue' or '#0000ff')."),
```
staubina: Should we use "example" instead of e.g.? I am not sure how well e.g. translates.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -90,22 +90,22 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
...
         scope=Scope.settings,
         default="",
     )
 
     item_text_color = String(
         display_name=_("Item text color"),
-        help=_("Text color to use for draggable items."),
+        help=_("Text color to use for draggable items (e.g. 'white' or '#ffffff')."),
```
staubina: Same as last comment on the use of e.g.

staubina:
The rest looks good to me for the context of the story.  👍  Pending @cahrens comments being resolved.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -137,6 +140,7 @@ function DragAndDropEditBlock(runtime, element, params) {
                         $fbkTab.addClass('hidden');
                         $zoneTab.removeClass('hidden');
                         self.scrollToTop();
+                        $zoneTab.find('input:first').select();
 
                         $(this).one('click', function loadThirdTab(e) {
```
cahrens: I believe there are preventDefault calls you can remove in this file (and perhaps others) now that the a-tags have been converted to buttons.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -67,6 +67,9 @@ function DragAndDropEditBlock(runtime, element, params) {
 
                     // Hide settings that are specific to assessment mode
                     _fn.build.$el.feedback.form.find('.problem-mode').trigger('change');
+
+                    // Set focus on first input field.
+                    $element.find('input:first').select();
```
cahrens: @cptvitamin can weigh in on this during the a11y review. I do wonder which is better-- moving focus to the first input field vs. moving it to the top of the visible text above the field (with tabindex = -1). I suppose it depends on if the text is linked to the input field. For example, on the second "tab", the section "Zones" is not linked to the label/input field and could be missed by a screen reader.

But as I said, that level of detail can wait until the a11y review.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -137,7 +140,7 @@ function DragAndDropEditBlock(runtime, element, params) {
                         $fbkTab.addClass('hidden');
                         $zoneTab.removeClass('hidden');
                         self.scrollToTop();
-                        $zoneTab.find('input:first').focus();
+                        $zoneTab.find('input:first').select();
```
cahrens: I also don't know which is better-- focus vs. select.

cahrens:
👍 It would be good to remove the unneeded event.preventDefault() calls, but I don't need to take a look at this again.

itsjeyd:
Rebased https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/88/commits/0af3fc39303aa6bbaf2dc357948b24252e0e9f70 onto master.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
itsjeyd: @cahrens Thanks for pointing that out. I removed the `preventDefault` calls for the a-tags that are now buttons. (Taking over for @mtyaka since he is away this week.)

itsjeyd:
@staubina @cahrens Thanks again for the review. Comments addressed (aside from the one about moving focus that you said could be handled as part of the a11y review, @cahrens).

@cptvitamin @clrux I've added this story to the a11y review ticket for DnDv2.1 ([SOL-1978](https://openedx.atlassian.net/browse/SOL-1978)) so you can have a look at it together with the other DnDv2.1 features. (My bandwidth for working on this PR is limited this week, and since it builds on some DnDv2.1 stories that were already part of SOL-1978, it makes sense to add this PR to the combined review.)

@sstack22 This is ready for you, please have a look.

sstack22:
Looks great, 👍  on my end. Two small nits: capitalization is a bit inconsistent (Final Feedback vs. Item description, as an example).  I would also suggest removing the Note at the top of the Studio Editor - I'm concerned about course teams being prompted to delete problems. As this is outside the scope of this PR, I'm fine with this being addressed separately.

itsjeyd:
@sstack22 Thanks for the review! Merging the PR now.

Re: your comments about inconsistent capitalization and removing the "Note", I'll see with @bradenmacdonald whether/how we can incorporate those into the work that's left for the DnDv2.1 epic.

PR #92, opened 2016-08-09
Prevent overlapping drop placements

e-kolpakov:
**Description:** This PR updates item placement logic:

* Zone alignment "none" (aka no alignment) is removed, dropped items always use the "aligned" layout option rather than being placed where the learner dropped them.
* "Max items per zone" setting added to allow limiting max number of items dropped into a single zone - course authors can use this setting to provide some kind of a hint, or to prevent visual overflow when too many items are dropped into single zone.

**JIRA:** [SOL-1979](https://openedx.atlassian.net/browse/SOL-1979)

**Discussions:** [SOL-1979](https://openedx.atlassian.net/browse/SOL-1979) and https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/83#issuecomment-236017205

**Sandboxes:**

* LMS: http://dndv2-sandbox9.opencraft.hosting/
* Studio: http://studio-dndv2-sandbox9.opencraft.hosting/

**Note:** at the moment #86 was merged, but #87 was not, and this PR does not include changes from #87; as a result, DnDv2 on sandbox show "Submit" button, but it does not do anything - this is an expected behavior until #87 is merged

**Testing instructions:**

Items alignment:
1. Open new or existing DnDv2 block in studio, open editor, go to "Zones" tab. *Expected:* zone "Alignment" dropdown does not have "none" option; default is "center"
2. Open DnDv2 XBlock in LMS.
3. Make sure that items dropped into a zone are aligned according to zone alignment configuration.
4. Existing zones with "none" alignment should be treated as having "center"alignment; this should be the case even if an XBlock was not updated in Studio after this PR is applied.

Max items per zone:
1. Open new or existing DnDv2 block in studio, open editor, go to "Items" tab. *Expected:* "Max items per zone" is displayed.
2. Setting "max items per zone" to ~~zero or~~ empty string has no further effect (make sure you can put as much items into a single dropzone as you wish in LMS)
3. Setting "max items per zone" to zero or negative value ~~should cause validation errors~~ will result in value being set to empty string
4. ~~Setting "max items per zone" to positive value will validate item placement. If any zone has more "correct" items than "max items per zone" a validation error will be shown, and XBlock will not be updated.~~
5. After setting "max items per zone" to positive value ~~and passing the validation~~, publish the unit and open it in LMS. *Expected:* can only put up to "max items per zone" items into a single zone.
6. *Note:* existing answers are not updated, so it is possible to have more than "max items per zone" items dropped into a single zone if "max items" setting changed after the answer was submitted.

**Screenshots:**

Studio "Zones" tab:

image

Studio "Items" tab:

image

LMS - item alignment:

image

Attempting to exceed max items per zone:

image

**Reviewers:**
- [x] @itsjeyd 
- [ ] ~~a11y: @cptvitamin~~  a11y review will be performed later, for all changes in v2.1: [SOL-1978](https://openedx.atlassian.net/browse/SOL-1978)
- [x] Product: @sstack22

**Checklist:**
- [x] Internal review
- [x] Upstream review
- [x] Squash commits

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The editor is fully guided. Features include:
 * custom zone labels
 * ability to show or hide zone borders
 * custom text and background colors for items
-* optional auto-alignment for items (left, right, center)
+* auto-alignment for items: left, right, center
 * image items
 * decoy items that don't have a zone
 * feedback popups for both correct and incorrect attempts
@@ -122,15 +122,13 @@ whether or not to display borders outlining the zones. It is possible
 to define an arbitrary number of drop zones as long as their labels
 are unique.
 
-Additionally, you can specify the alignment for items once they are dropped in
-the zone.  No alignment is the default, and causes items to stay where the
-learner drops them.  Left alignment causes dropped items to be placed from left
+You can specify the alignment for items once they are dropped in
+the zone. Centered alignment is the default, and places items from top to bottom
+along the center of the zone. Left alignment causes dropped items to be placed from left
 to right across the zone. Right alignment causes the items to be placed from
-right to left across the zone. Center alignment places items from top to bottom
-along the center of the zone. If left, right, or center alignment is chosen,
-items dropped in a zone will not overlap, but if the zone is not made large
-enough for all its items, they will overflow the bottom of the zone, and
-potentially, overlap the zones below.
+right to left across the zone. Items dropped in a zone will not overlap,
+but if the zone is not made large enough for all its items, they will overflow the bottom
+of the zone, and potentially overlap the zones below.
 
 ![Drag item edit](/doc/img/edit-view-items.png)
 
@@ -150,6 +148,9 @@ You can leave all of the checkboxes unchecked in order to create a
 
 You can define an arbitrary number of drag items, each of which may
 be attached to any number of zones.
+
+"Maximum items per Zone" setting controls how many items can be dropped into a
+single zone, allowing some degree of control over items overlapping zones below.
 
 Scoring
 -------
--- a/drag_and_drop_v2/default_data.py
+++ b/drag_and_drop_v2/default_data.py
@@ -38,6 +38,7 @@ DEFAULT_DATA = {
             "y": 30,
             "width": 196,
             "height": 178,
+            "align": "center"
         },
         {
             "uid": MIDDLE_ZONE_ID,
@@ -47,6 +48,7 @@ DEFAULT_DATA = {
             "y": 210,
             "width": 340,
             "height": 138,
+            "align": "center"
         },
         {
             "uid": BOTTOM_ZONE_ID,
@@ -56,6 +58,7 @@ DEFAULT_DATA = {
             "y": 350,
             "width": 485,
             "height": 135,
+            "align": "center"
         }
     ],
     "items": [
@@ -65,7 +68,9 @@ DEFAULT_DATA = {
                 "incorrect": ITEM_INCORRECT_FEEDBACK,
                 "correct": ITEM_CORRECT_FEEDBACK.format(zone=TOP_ZONE_TITLE)
             },
-            "zones": [TOP_ZONE_ID],
+            "zones": [
+                TOP_ZONE_ID
+            ],
             "imageURL": "",
             "id": 0,
         },
@@ -75,7 +80,9 @@ DEFAULT_DATA = {
                 "incorrect": ITEM_INCORRECT_FEEDBACK,
                 "correct": ITEM_CORRECT_FEEDBACK.format(zone=MIDDLE_ZONE_TITLE)
             },
-            "zones": [MIDDLE_ZONE_ID],
+            "zones": [
+                MIDDLE_ZONE_ID
+            ],
             "imageURL": "",
             "id": 1,
         },
@@ -85,7 +92,9 @@ DEFAULT_DATA = {
                 "incorrect": ITEM_INCORRECT_FEEDBACK,
                 "correct": ITEM_CORRECT_FEEDBACK.format(zone=BOTTOM_ZONE_TITLE)
             },
-            "zones": [BOTTOM_ZONE_ID],
+            "zones": [
+                BOTTOM_ZONE_ID
+            ],
             "imageURL": "",
             "id": 2,
         },
@@ -95,7 +104,11 @@ DEFAULT_DATA = {
                 "incorrect": "",
                 "correct": ITEM_ANY_ZONE_FEEDBACK
             },
-            "zones": [TOP_ZONE_ID, BOTTOM_ZONE_ID, MIDDLE_ZONE_ID],
+            "zones": [
+                TOP_ZONE_ID,
+                BOTTOM_ZONE_ID,
+                MIDDLE_ZONE_ID
+            ],
             "imageURL": "",
             "id": 3
         },
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -6,6 +6,7 @@
 
 import copy
 import json
+import logging
 import urllib
 import webob
 
@@ -16,16 +17,17 @@ from xblock.fragment import Fragment
 from xblockutils.resources import ResourceLoader
 from xblockutils.settings import XBlockWithSettingsMixin, ThemableXBlockMixin
 
-from .utils import _, DummyTranslationService, FeedbackMessage, FeedbackMessages, ItemStats
+from .utils import _, DummyTranslationService, FeedbackMessage, FeedbackMessages, ItemStats, StateMigration, Constants
 from .default_data import DEFAULT_DATA
 
 
 # Globals ###########################################################
 
 loader = ResourceLoader(__name__)
-
+logger = logging.getLogger(__name__)
 
 # Classes ###########################################################
+
 
 @XBlock.wants('settings')
 @XBlock.needs('i18n')
@@ -33,8 +35,6 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     """
     XBlock that implements a friendly Drag-and-Drop problem
     """
-    STANDARD_MODE = "standard"
-    ASSESSMENT_MODE = "assessment"
 
     SOLUTION_CORRECT = "correct"
     SOLUTION_PARTIAL = "partial"
@@ -69,10 +69,10 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         ),
         scope=Scope.settings,
         values=[
-            {"display_name": _("Standard"), "value": STANDARD_MODE},
-            {"display_name": _("Assessment"), "value": ASSESSMENT_MODE},
+            {"display_name": _("Standard"), "value": Constants.STANDARD_MODE},
+            {"display_name": _("Assessment"), "value": Constants.ASSESSMENT_MODE},
         ],
-        default=STANDARD_MODE
+        default=Constants.STANDARD_MODE
     )
 
     max_attempts = Integer(
@@ -127,6 +127,13 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         default="",
     )
 
+    max_items_per_zone = Integer(
+        display_name=_("Maximum items per zone"),
+        help=_("This setting limits the number of items that can be dropped into a single zone."),
+        scope=Scope.settings,
+        default=None
+    )
+
     data = Dict(
         display_name=_("Problem data"),
         help=_(
@@ -157,7 +164,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     )
 
     grade = Float(
-        help=_("Keeps maximum achieved score by student"),
+        help=_("Keeps maximum score achieved by student"),
         scope=Scope.user_state,
         default=0
     )
@@ -223,8 +230,9 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
         return {
             "mode": self.mode,
+            "zones": self.zones,
             "max_attempts": self.max_attempts,
-            "zones": self._get_zones(),
+            "max_items_per_zone": self.max_items_per_zone,
             # SDK doesn't supply url_name.
             "url_name": getattr(self, 'url_name', ''),
             "display_zone_labels": self.data.get('displayLabels', False),
@@ -286,7 +294,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         items = self.data.get('items', [])
 
         for item in items:
-            zones = self._get_item_zones(item['id'])
+            zones = self.get_item_zones(item['id'])
             # Note that we appear to be mutating the state of the XBlock here, but because
             # the change won't be committed, we're actually just affecting the data that
             # we're going to send to the client, not what's saved in the backing store.
@@ -315,11 +323,45 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         self.weight = float(submissions['weight'])
         self.item_background_color = submissions['item_background_color']
         self.item_text_color = submissions['item_text_color']
+        self.max_items_per_zone = self._get_max_items_per_zone(submissions)
         self.data = submissions['data']
 
         return {
             'result': 'success',
         }
+
+    @staticmethod
+    def _get_max_items_per_zone(submissions):
+        """
+        Parses Max items per zone value coming from editor.
+
+        Returns:
+            * None if invalid value is passed (i.e. not an integer)
+            * None if value is parsed into zero or negative integer
+            * Positive integer otherwise.
+
+        Examples:
+            * _get_max_items_per_zone(None) -> None
+            * _get_max_items_per_zone('string') -> None
+            * _get_max_items_per_zone('-1') -> None
+            * _get_max_items_per_zone(-1) -> None
+            * _get_max_items_per_zone('0') -> None
+            * _get_max_items_per_zone('') -> None
+            * _get_max_items_per_zone('42') -> 42
+            * _get_max_items_per_zone(42) -> 42
+        """
+        raw_max_items_per_zone = submissions.get('max_items_per_zone', None)
+
+        # Entries that aren't numbers should be treated as null. We assume that if we can
+        # turn it into an int, a number was submitted.
+        try:
+            max_attempts = int(raw_max_items_per_zone)
+            if max_attempts > 0:
+                return max_attempts
+            else:
+                return None
+        except (ValueError, TypeError):
+            return None
 
     @XBlock.json_handler
     def drop_item(self, item_attempt, suffix=''):
@@ -328,9 +370,9 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
         self._validate_drop_item(item_attempt)
 
-        if self.mode == self.ASSESSMENT_MODE:
+        if self.mode == Constants.ASSESSMENT_MODE:
             return self._drop_item_assessment(item_attempt)
-        elif self.mode == self.STANDARD_MODE:
+        elif self.mode == Constants.STANDARD_MODE:
             return self._drop_item_standard(item_attempt)
         else:
             raise JsonHandlerError(
@@ -434,7 +476,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
         Validates if `do_attempt` handler should be executed
         """
-        if self.mode != self.ASSESSMENT_MODE:
+        if self.mode != Constants.ASSESSMENT_MODE:
             raise JsonHandlerError(
                 400,
                 self.i18n_service.gettext("do_attempt handler should only be called for assessment mode")
@@ -452,7 +494,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         answer_correctness = self._answer_correctness()
         is_correct = answer_correctness == self.SOLUTION_CORRECT
 
-        if self.mode == self.STANDARD_MODE or not self.attempts:
+        if self.mode == Constants.STANDARD_MODE or not self.attempts:
             feedback_key = 'finish' if is_correct else 'start'
             return [FeedbackMessage(self.data['feedback'][feedback_key], None)], set()
 
@@ -563,9 +605,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
         return {
             'zone': attempt['zone'],
-            'correct': correct,
-            'x_percent': attempt['x_percent'],
-            'y_percent': attempt['y_percent'],
+            'correct': correct
         }
 
     def _mark_complete_and_publish_grade(self):
@@ -613,7 +653,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
         Check if the item was placed correctly.
         """
-        correct_zones = self._get_item_zones(attempt['val'])
+        correct_zones = self.get_item_zones(attempt['val'])
         return attempt['zone'] in correct_zones
 
     def _expand_static_url(self, url):
@@ -640,12 +680,12 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         item_state = self._get_item_state()
         # In assessment mode, we do not want to leak the correctness info for individual items to the frontend,
         # so we remove "correct" from all items when in assessment mode.
-        if self.mode == self.ASSESSMENT_MODE:
+        if self.mode == Constants.ASSESSMENT_MODE:
             for item in item_state.values():
                 del item["correct"]
 
         overall_feedback_msgs, __ = self._get_feedback()
-        if self.mode == self.STANDARD_MODE:
+        if self.mode == Constants.STANDARD_MODE:
             is_finished = self._is_answer_correct()
         else:
             is_finished = not self.attempts_remain
@@ -666,33 +706,10 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         # IMPORTANT: this method should always return a COPY of self.item_state - it is called from get_user_state
         # handler and the data it returns is manipulated there to hide correctness of items placed.
         state = {}
+        migrator = StateMigration(self)
 
-        for item_id, raw_item in self.item_state.iteritems():
-            if isinstance(raw_item, dict):
-                # Items are manipulated in _get_user_state, so we protect actual data.
-                item = copy.deepcopy(raw_item)
-            else:
-                item = {'top': raw_item[0], 'left': raw_item[1]}
-            # If information about zone is missing
-            # (because problem was completed before a11y enhancements were implemented),
-            # deduce zone in which item is placed from definition:
-            if item.get('zone') is None:
-                valid_zones = self._get_item_zones(int(item_id))
-                if valid_zones:
-                    # If we get to this point, then the item was placed prior to support for
-                    # multiple correct zones being added. As a result, it can only be correct
-                    # on a single zone, and so we can trust that the item was placed on the
-                    # zone with index 0.
-                    item['zone'] = valid_zones[0]
-                else:
-                    item['zone'] = 'unknown'
-            # If correctness information is missing
-            # (because problem was completed before assessment mode was implemented),
-            # assume the item is in correct zone (in standard mode, only items placed
-            # into correct zone are stored in item state).
-            if item.get('correct') is None:
-                item['correct'] = True
-            state[item_id] = item
+        for item_id, item in self.item_state.iteritems():
+            state[item_id] = migrator.apply_item_state_migrations(item_id, item)
 
         return state
 
@@ -702,7 +719,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
         return next(i for i in self.data['items'] if i['id'] == item_id)
 
-    def _get_item_zones(self, item_id):
+    def get_item_zones(self, item_id):
         """
         Returns a list of the zones that are valid options for the item.
 
@@ -720,27 +737,20 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         else:
             return []
 
-    def _get_zones(self):
+    @property
+    def zones(self):
         """
         Get drop zone data, defined by the author.
         """
         # Convert zone data from old to new format if necessary
-        zones = []
-        for zone in self.data.get('zones', []):
-            zone = zone.copy()
-            if "uid" not in zone:
-                zone["uid"] = zone.get("title")  # Older versions used title as the zone UID
-            # Remove old, now-unused zone attributes, if present:
-            zone.pop("id", None)
-            zone.pop("index", None)
-            zones.append(zone)
-        return zones
+        migrator = StateMigration(self)
+        return [migrator.apply_zone_migrations(zone) for zone in self.data.get('zones', [])]
 
     def _get_zone_by_uid(self, uid):
         """
         Given a zone UID, return that zone, or None.
         """
-        for zone in self._get_zones():
+        for zone in self.zones:
             if zone["uid"] == uid:
                 return zone
 
@@ -772,7 +782,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         item_state = self._get_item_state()
 
         all_items = set(str(item['id']) for item in self.data['items'])
-        required = set(item_id for item_id in all_items if self._get_item_zones(int(item_id)) != [])
+        required = set(item_id for item_id in all_items if self.get_item_zones(int(item_id)) != [])
         placed = set(item_id for item_id in all_items if item_id in item_state)
         correctly_placed = set(item_id for item_id in placed if item_state[item_id]['correct'])
         decoy = all_items - required
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -84,23 +84,11 @@ function DragAndDropTemplates(configuration) {
             style['outline-color'] = item.color;
         }
         if (item.is_placed) {
-            if (item.zone_align === 'none') {
-                // This is not an "aligned" zone, so the item gets positioned where the learner dropped it.
-                style.left = item.x_percent + "%";
-                style.top = item.y_percent + "%";
-                if (item.widthPercent) {  // This item has an author-defined explicit width
-                    style.width = item.widthPercent + "%";
-                    style.maxWidth = item.widthPercent + "%"; // default maxWidth is ~30%
-                }
-            } else {
-                // This is an "aligned" zone, so the item position within the zone is calculated by the browser.
-                // Make up for the fact we're in a wrapper container by calculating percentage differences.
-                var maxWidth = (item.widthPercent || 30) / 100;
-                var widthPercent = zone.width_percent / 100;
-                style.maxWidth = ((1 / (widthPercent / maxWidth)) * 100) + '%';
-                if (item.widthPercent) {
-                    style.width = style.maxWidth;
-                }
+            var maxWidth = (item.widthPercent || 30) / 100;
+            var widthPercent = zone.width_percent / 100;
+            style.maxWidth = ((1 / (widthPercent / maxWidth)) * 100) + '%';
+            if (item.widthPercent) {
+                style.width = style.maxWidth;
             }
             // Finally, if the item is using automatic sizing and contains an image, we
             // always prefer the natural width of the image (subject to the max-width):
@@ -192,7 +180,6 @@ function DragAndDropTemplates(configuration) {
     var zoneTemplate = function(zone, ctx) {
         var className = ctx.display_zone_labels ? 'zone-name' : 'zone-name sr';
         var selector = ctx.display_zone_borders ? 'div.zone.zone-with-borders' : 'div.zone';
-
         // If zone is aligned, mark its item alignment
         // and render its placed items as children
         var item_wrapper = 'div.item-wrapper';
@@ -389,8 +376,6 @@ function DragAndDropTemplates(configuration) {
         var is_item_placed = function(i) { return i.is_placed; };
         var items_placed = $.grep(ctx.items, is_item_placed);
         var items_in_bank = $.grep(ctx.items, is_item_placed, true);
-        var is_item_placed_unaligned = function(i) { return i.zone_align === 'none'; };
-        var items_placed_unaligned = $.grep(items_placed, is_item_placed_unaligned);
         var item_bank_properties = {};
         if (ctx.item_bank_focusable) {
             item_bank_properties.attributes = {
@@ -421,7 +406,6 @@ function DragAndDropTemplates(configuration) {
                                 h('img.target-img', {src: ctx.target_img_src, alt: ctx.target_img_description}),
                             ]
                         ),
-                        renderCollection(itemTemplate, items_placed_unaligned, ctx),
                         renderCollection(zoneTemplate, ctx.zones, ctx)
                     ]),
                 ]),
@@ -463,6 +447,8 @@ function DragAndDropBlock(runtime, element, configuration) {
     // Event string size limit.
     var MAX_LENGTH = 255;
 
+    var DEFAULT_ZONE_ALIGN = 'center';
+
     // Keyboard accessibility
     var ESC = 27;
     var RET = 13;
@@ -490,7 +476,7 @@ function DragAndDropBlock(runtime, element, configuration) {
             });
             state = stateResult[0]; // stateResult is an array of [data, statusText, jqXHR]
             migrateConfiguration(bgImg.width);
-            migrateState(bgImg.width, bgImg.height);
+            migrateState();
             markItemZoneAlign();
             bgImgNaturalWidth = bgImg.width;
 
@@ -755,40 +741,40 @@ function DragAndDropBlock(runtime, element, configuration) {
 
     var placeItem = function($zone, $item) {
         var item_id;
-        var $anchor;
         if ($item !== undefined) {
             item_id = $item.data('value');
-            // Element was placed using the mouse,
-            // so use relevant properties of *item* when calculating new position below.
-            $anchor = $item;
         } else {
             item_id = $selectedItem.data('value');
-            // Element was placed using the keyboard,
-            // so use relevant properties of *zone* when calculating new position below.
-            $anchor = $zone;
         }
+
         var zone = String($zone.data('uid'));
         var zone_align = $zone.data('zone_align');
-        var $target_img = $root.find('.target-img');
 
-        // Calculate the position of the item to place relative to the image.
-        var x_pos = $anchor.offset().left + ($anchor.outerWidth()/2) - $target_img.offset().left;
-        var y_pos = $anchor.offset().top + ($anchor.outerHeight()/2) - $target_img.offset().top;
-        var x_pos_percent = x_pos / $target_img.width() * 100;
-        var y_pos_percent = y_pos / $target_img.height() * 100;
+        var items_in_zone_count = countItemsInZone(zone, [item_id.toString()]);
+        if (configuration.max_items_per_zone && configuration.max_items_per_zone <= items_in_zone_count) {
+            state.last_action_correct = false;
+            state.feedback = gettext("You cannot add any more items to this zone.");
+            applyState();
+            return;
+        }
 
         state.items[item_id] = {
             zone: zone,
             zone_align: zone_align,
-            x_percent: x_pos_percent,
-            y_percent: y_pos_percent,
             submitting_location: true,
         };
         // Wrap in setTimeout to let the droppable event finish.
         setTimeout(function() {
             applyState();
-            submitLocation(item_id, zone, x_pos_percent, y_pos_percent);
+            submitLocation(item_id, zone);
         }, 0);
+    };
+
+    var countItemsInZone = function(zone, exclude_ids) {
+        var ids_to_exclude = exclude_ids ? exclude_ids : [];
+        return Object.keys(state.items).filter(function(item_id) {
+            return state.items[item_id].zone === zone && $.inArray(item_id, ids_to_exclude) === -1;
+        }).length;
     };
 
     var initDroppable = function() {
@@ -805,6 +791,7 @@ function DragAndDropBlock(runtime, element, configuration) {
                         releaseItem($selectedItem);
                     } else if (isActionKey(evt)) {
                         evt.preventDefault();
+                        evt.stopPropagation();
                         state.keyboard_placement_mode = false;
                         releaseItem($selectedItem);
                         if ($zone.is('.item-bank')) {
@@ -946,16 +933,14 @@ function DragAndDropBlock(runtime, element, configuration) {
         });
     };
 
-    var submitLocation = function(item_id, zone, x_percent, y_percent) {
+    var submitLocation = function(item_id, zone) {
         if (!zone) {
             return;
         }
         var url = runtime.handlerUrl(element, 'drop_item');
         var data = {
             val: item_id,
-            zone: zone,
-            x_percent: x_percent,
-            y_percent: y_percent,
+            zone: zone
         };
 
         $.post(url, JSON.stringify(data), 'json')
@@ -1107,8 +1092,6 @@ function DragAndDropBlock(runtime, element, configuration) {
             if (item_user_state) {
                 itemProperties.zone = item_user_state.zone;
                 itemProperties.zone_align = item_user_state.zone_align;
-                itemProperties.x_percent = item_user_state.x_percent;
-                itemProperties.y_percent = item_user_state.y_percent;
             }
             if (configuration.item_background_color) {
                 itemProperties.background_color = configuration.item_background_color;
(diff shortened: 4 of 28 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=92

drag_and_drop_v2/default_data.py
```diff
@@ -46,6 +47,8 @@
       "y": 210,
       "width": 340,
       "height": 138,
+      "align": "center"
+
```
itsjeyd: @e-kolpakov Nit: Remove extra blank line

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -280,12 +299,60 @@ def studio_submit(self, submissions, suffix=''):
...
+        # Possible ValueError should be catched in _validate_submissions
+        self.max_items_per_zone = self._get_max_items(submissions)
         self.data = submissions['data']
 
         return {
             'result': 'success',
         }
 
+    @staticmethod
+    def _get_max_items(submission):
```
itsjeyd: @e-kolpakov For consistency, can we rename `submission` to `submissions` here? When I first read this I found myself double-checking if `studio_submit` was passing the `submissions` it receives unaltered, or if it was passing only a subset of the `submissions` data.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -280,12 +299,60 @@ def studio_submit(self, submissions, suffix=''):
...
+        """
+        Parses Max items per zone value coming from editor.
+
+        Does not catch ValueError intentionally to allow validation to capture this problem
+        """
+        raw_max_submissions = submission.get('max_items_per_zone', None)
```
itsjeyd: @e-kolpakov It would make sense to rename this to `raw_max_items` I think.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -280,12 +299,60 @@ def studio_submit(self, submissions, suffix=''):
...
+
+    def _validate_submissions(self, submissions):
+        """
+        Validates that submissions passed are correct
+        """
+        errors = []
+
+        items = submissions['data']['items']
+
+        max_submissions = 0
```
itsjeyd: @e-kolpakov Similar note here: `max_items` or `max_items_per_zone` would probably be a little more accurate.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -280,12 +299,60 @@ def studio_submit(self, submissions, suffix=''):
...
+        try:
+            max_submissions = self._get_max_items(submissions)
+        except ValueError as exc:
+            errors.append(_('Failed to parse "Max items per zone" - please check settings'))
+            logger.exception(exc)
+
+        if max_submissions:
+            counter = Counter()
+            for item in items:
+                zones = item.get('zones', [item.get('zone')])
```
itsjeyd: @e-kolpakov Add an inline comment about backward compatibility here, perhaps?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -444,10 +509,7 @@ def _get_item_state(self):
         state = {}
 
         for item_id, item in self.item_state.iteritems():
-            if isinstance(item, dict):
-                state[item_id] = item
-            else:
-                state[item_id] = {'top': item[0], 'left': item[1]}
+            state[item_id] = StateMigration.apply_item_state_migrations(item)
```
itsjeyd: @e-kolpakov Thanks for factoring the migration logic out to utils.py! Makes the code much more readable +1

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -186,16 +174,13 @@ function DragAndDropTemplates(configuration) {
...
         // If zone is aligned, mark its item alignment
         // and render its placed items as children
         var item_wrapper = 'div.item-wrapper';
-        var items_in_zone = [];
-        if (zone.align !== 'none') {
-            item_wrapper += '.item-align.item-align-' + zone.align;
-            var is_item_in_zone = function(i) { return i.is_placed && (i.zone === zone.uid); };
-            items_in_zone = $.grep(ctx.items, is_item_in_zone);
-        }
+        item_wrapper += '.item-align.item-align-' + zone.align;
```
itsjeyd: @e-kolpakov A couple nits:

* Maybe define `item_wrapper` in a single line
* Update the comment above to reflect the fact that all zones use alignment now

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -657,43 +641,48 @@ function DragAndDropBlock(runtime, element, configuration) {
...
         var zone_align = $zone.data('zone_align');
-        var $target_img = $root.find('.target-img');
 
-        // Calculate the position of the item to place relative to the image.
-        var x_pos = $anchor.offset().left + ($anchor.outerWidth()/2) - $target_img.offset().left;
-        var y_pos = $anchor.offset().top + ($anchor.outerHeight()/2) - $target_img.offset().top;
-        var x_pos_percent = x_pos / $target_img.width() * 100;
-        var y_pos_percent = y_pos / $target_img.height() * 100;
+        var items_in_zone_count = countItemsInZone(zone);
+        if (configuration.max_items_per_zone && configuration.max_items_per_zone <= items_in_zone_count) {
```
itsjeyd: @e-kolpakov If `max_items_per_zone` is equal to `items_in_zone_count`, that should probably be OK? In other words, should this code only reject the placement of an item if `items_in_zone_count > max_items_per_zone`?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -657,43 +641,48 @@ function DragAndDropBlock(runtime, element, configuration) {
...
             applyState();
-            submitLocation(item_id, zone, x_pos_percent, y_pos_percent);
+            submitLocation(item_id, zone);
         }, 0);
     };
 
+    var countItemsInZone = function(zone) {
+        var count = 0;
+        for (var key in state.items) {
+            if (state.items.hasOwnProperty(key)) {
```
itsjeyd: @e-kolpakov Why is this check necessary? (genuine question)

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1024,32 +1010,8 @@ function DragAndDropBlock(runtime, element, configuration) {
...
-                var left_px = parseFloat(item.left) - 220; // 220 px for the items container that used to be on the left
-                var top_px = parseFloat(item.top);
-                item.x_percent = (left_px + width/2) / bg_image_width * 100;
-                item.y_percent = (top_px + height/2) / bg_image_height * 100;
-                delete item.left;
-                delete item.top;
-                delete item.absolute;
-            }
-        });
+    var migrateState = function() {
```
itsjeyd: @e-kolpakov Leftover stub

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -514,18 +514,27 @@ function DragAndDropEditBlock(runtime, element, params) {
...
                             'item_text_color': $element.find('#item-text-color').val(),
+                            'max_items_per_zone': $element.find("#max-items-per-zone").val(),
                             'data': _fn.data,
                         };
 
                         $('.xblock-editor-error-message', element).html();
                         $('.xblock-editor-error-message', element).css('display', 'none');
                         var handlerUrl = runtime.handlerUrl(element, 'studio_submit');
+
+                        $('.xblock-editor-error-message', element).css('display', 'none');
```
itsjeyd: @e-kolpakov Duplicate of L522. Could L521 be removed?

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,135 @@
 # Make '_' a no-op so we can scrape strings
 def _(text):
     return text
+
+
+class Constants(object):
+    ALLOWED_ZONE_ALIGNMENT = ['left', 'right', 'center']
```
itsjeyd: @e-kolpakov Nit: `ALLOWED_ZONE_ALIGNMENTS` (since it's a list)

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,135 @@
...
+
+        return cls._apply_migration(item_state, migrations)
+
+    @classmethod
+    def _zone_v1_to_v2(cls, zone):
+        """
+        Migrates zone data from v1.0 format to v2.0 format.
+
+        Changes:
+        * v1 used zone "title" as UID, while v2 have "uid" property
```
itsjeyd: @e-kolpakov `while v2 zones have a "uid" property`

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,135 @@
...
+            zone["uid"] = zone.get("title")
+        zone.pop("id", None)
+        zone.pop("index", None)
+
+        return zone
+
+    @classmethod
+    def _zone_v2_to_v2p1(cls, zone):
+        """
+        Migrates zone data from v2.0 to v2.5
```
itsjeyd: @e-kolpakov I think this should be `v2.0` to `v.2.1`. Same comment applies to the docstring of `_item_state_v1p5_to_v2p1` below.

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,135 @@
...
+        else:
+            return {'top': item[0], 'left': item[1]}
+
+    @classmethod
+    def _item_state_v1p5_to_v2(cls, item):
+        """
+        Migrates item_state from v1.5 to v2.0
+
+        Changes:
+        * Item placement attributes switched from absolute (left-top) to relative (x_percent-y_pecrent) units
```
itsjeyd: @e-kolpakov Typo: `pecrent`

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,135 @@
...
+
+        In: {'zone': 'Zone", 'correct': True, 'top': '100px', 'left': '120px'}
+        Out: {'zone': 'Zone", 'correct': True, 'top': '100px', 'left': '120px'}
+        """
+        # Conversion can't be made as parent dimensions are unknown to python - converted in JS
+        # Since 2.5 JS this conversion became unnecesary, so it was removed from JS code
```
itsjeyd: @e-kolpakov `v2.1`?

drag_and_drop_v2/utils.py
```diff
@@ -5,3 +5,135 @@
...
+        return item
+
+    @classmethod
+    def _item_state_v1p5_to_v2p1(cls, item):
+        """
+        Migrates item_state from v1.5 to v2.5
+
+        Changes:
+        * Removed "none" zone alignment - remove old "absolute" placement attributes
```
itsjeyd: @e-kolpakov Maybe turn this into two items (putting note about "absolute" placement first to match the examples below):

```
* Removed old "absolute" placement attributes
* Removed "none" zone alignment, making "x_percent" and "y_percent" attributes obsolete
```

tests/integration/test_interaction.py
```diff
@@ -817,28 +821,9 @@ def _assert_zone_align_item(self, item_id, zone_id, align, action_key=None):
...
-
-        # Non-aligned items are absolute positioned, with top/bottom set to px
-        self.assertEquals(self._get_style(target_item_selector, 'position'), 'absolute')
-        self.assertRegexpMatches(self._get_style(target_item_selector, 'left'), r'^\d+(\.\d+)?px$')
-        self.assertRegexpMatches(self._get_style(target_item_selector, 'top'), r'^\d+(\.\d+)?px$')
-
     @data(
-        ([3, 4, 5], "Zone Invalid Align", "start"),
+        ([0, 1, 2], "Zone No Align", "center"),
+        ([3, 4, 5], "Zone Invalid Align", "center"),
```
itsjeyd: @e-kolpakov Why is `"center"` an invalid alignment?

tests/unit/data/old/config_out.json
```diff
@@ -20,15 +21,17 @@
           "y": "200",
           "width": 200,
           "height": 100,
-          "uid": "Zone 1"
+          "uid": "Zone 1",
+          "align": "center"
```
itsjeyd: @e-kolpakov Since this is supposed to represent v1 data, I'm wondering if we should remove the `"align"` setting here, and modify the migration code to add the setting if it's not present.

(The same goes for the `"uid"` setting, but I know you didn't add that.)

tests/unit/test_advanced.py
```diff
@@ -57,11 +57,16 @@ def initial_feedback(cls):
         return cls.expected_configuration()["initial_feedback"]
 
     def test_get_configuration(self):
-        self.assertEqual(self.expected_configuration(), self.block.get_configuration())
+        try:
+            self.assertEqual(self.expected_configuration(), self.block.get_configuration())
+        except AssertionError:
+            print self.expected_configuration()
+            print self.block.get_configuration()
+            raise
```
itsjeyd: @e-kolpakov This can be useful, but maybe it's a leftover from testing on your local machine? :)

tests/unit/test_basics.py
```diff
@@ -15,6 +15,29 @@ def setUp(self):
         self.block = make_block()
         self.patch_workbench()
 
+    def _make_submission(self, modify_submitssion=None):
```
itsjeyd: @e-kolpakov Typo: `submitssion`

tests/unit/test_basics.py
```diff
@@ -107,33 +131,101 @@ def assert_user_state_empty():
...
         self.assertEqual(self.block.question_text, "Problem Drag & Drop")
         self.assertEqual(self.block.show_question_header, False)
         self.assertEqual(self.block.item_background_color, "cornflowerblue")
         self.assertEqual(self.block.item_text_color, "coral")
         self.assertEqual(self.block.weight, 5)
-        self.assertEqual(self.block.data, {'foo': 1})
+        self.assertEqual(self.block.max_items_per_zone, None)
+        self.assertEqual(self.block.data, {'foo': 1, 'items': []})
+
+    def test_studio_submit2(self):
```
itsjeyd: @e-kolpakov Call this `test_studio_submit_assessment` or something like that?

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @itsjeyd no, this code works before the item is placed, so if we allow it to continue when `items_in_zone_count == max_items_per_zone` it will place one more item and we will exceed the limit.

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @itsjeyd `state.items` is an object, and iterating over object this way yields all "own" properties and all "inherited" properties (properties from prototype), mostly functions, liek `toString` and so on.

I believe the approach we used here was to use `Object.keys(object).forEach(...)` - I just have missed that and written it old-style :)

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @itsjeyd no, it was not empty before. I thought it might make sense to keep it to give a hint to future developers (self?) that we can apply some migrations in JS.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
e-kolpakov: @itsjeyd yes, nice catch - L521 and L522 totally filtered out from my attention; actually, we don't need this two lines added.

tests/integration/test_interaction.py
e-kolpakov: @itsjeyd "Zone Invalid Align" refers to a zone that have `"align": "invalid"` in configuration. In this case, we expect items placed into that zone to be centered.

tests/unit/data/old/config_out.json
e-kolpakov: @itsjeyd nope, the thing is, this particular test checks how DnDv2 parses old data format (aka v1) and how it responds to various fields/data members missing/having different names.

`config_out` files represent DnDv2 response to get_configuration handler as it is passed to frontend, so we expect it to have a most recent format (in this case v2.1)

tests/unit/test_advanced.py
e-kolpakov: @itsjeyd both - it is a leftover, but it is a useful leftover, because diffs are very hard to read, so most of the time it is simpler if you have those dicts printed.

CC @mtyaka @bradenmacdonald what do you think about this?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -708,6 +690,7 @@ function DragAndDropBlock(runtime, element, configuration) {
                         releaseItem($selectedItem);
                     } else if (isActionKey(evt)) {
                         evt.preventDefault();
+                        evt.stopPropagation();
```
e-kolpakov: @itsjeyd just to make sure - without this `stopPropagation` dropping an item using keyboard closed error popup before user could see it, resulting in item being returned to item bank without any explanation why.

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -161,7 +161,7 @@ function DragAndDropEditBlock(runtime, element, params) {
                             $(this).addClass('hidden');
                             $('.save-button', element).parent()
                                 .removeClass('hidden')
-                                .one('click', function submitForm(e) {
+                                .on('click', function submitForm(e) {
```
e-kolpakov: @itsjeyd FYI important one-letter change: with `one` if submission failed due to validation error, `Save` button stop working, so you've got to cancel and go all over again.

tests/unit/test_advanced.py
mtyaka: This is very useful while debugging a problem, but I think I'd prefer to take it out and manually add it back in while debugging a failing test.

bradenmacdonald: The problem is that committed code should almost never contain "print", unless you're writing some kind of interactive shell script. So if you leave it in, you should use this syntax:

```python
expected = self.expected_configuration()
actual = self.block.get_configuration()
self.assertEqual(
    expected,
    actual,
    "Unexpected configuration - Expected:\n{}\nObserved:\n{}".format(expected, actual)
)
```

p.s. You should get out of the habit of writing `print "foo"` since it's not python3 compatible :-p

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -657,43 +640,42 @@ function DragAndDropBlock(runtime, element, configuration) {
...
 
-        // Calculate the position of the item to place relative to the image.
-        var x_pos = $anchor.offset().left + ($anchor.outerWidth()/2) - $target_img.offset().left;
-        var y_pos = $anchor.offset().top + ($anchor.outerHeight()/2) - $target_img.offset().top;
-        var x_pos_percent = x_pos / $target_img.width() * 100;
-        var y_pos_percent = y_pos / $target_img.height() * 100;
+        var items_in_zone_count = countItemsInZone(zone);
+        if (configuration.max_items_per_zone && configuration.max_items_per_zone <= items_in_zone_count) {
+            state.last_action_correct = false;
+            state.feedback = gettext("This zone already contains maximum allowed number of items or more");
```
bradenmacdonald: Better phrasing: "You cannot add any more items to this zone."

bradenmacdonald:
In this screenshot, the left-aligned items retain a nice small automatic size, but the centered ones grow to fill their zone width - is there a way to use the "left" sizing behavior in both cases?

image

Also, on the sandbox I wasn't able to see the warning in red when dragging too many items onto a zone. 

Finally, can you please enable the XBlock view on the sandbox so we can test how this works on small devices / in the mobile app?

http://studio-dndv2-sandbox9.opencraft.hosting/xblock/block-v1:OpenCraft+DnDv2+demos+type@vertical+block@28ae4b11831b42949d8815c86671424b should work but seems to need `FEATURES["ENABLE_XBLOCK_VIEW_ENDPOINT"]` set to True.

drag_and_drop_v2/public/js/drag_and_drop.js
itsjeyd: @e-kolpakov I see, that makes sense. Thanks!

tests/unit/data/old/config_out.json
itsjeyd: @e-kolpakov Ah, right -- I thought I was commenting on one of the `data.json` files :) Sorry for the noise.

drag_and_drop_v2/public/js/drag_and_drop.js
itsjeyd: @e-kolpakov OK, that's a bit unusual but I get where you're coming from. If you're not going to remove the method, could you update the docstring above to mention why it's there (and remove the part about image size)?

drag_and_drop_v2/public/js/drag_and_drop.js
itsjeyd: @e-kolpakov Thanks for the explanation, I had wondered about that briefly but somehow forgot to comment about it - you read my mind :)

drag_and_drop_v2/public/js/drag_and_drop_edit.js
itsjeyd: @e-kolpakov OK. But now that I'm looking at the surrounding code I'm wondering what the purpose of [this code](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/c1941da6629d38d41b5b2cc8b162b323dc5f80d6/drag_and_drop_v2/public/js/drag_and_drop_edit.js#L168-L171) is. It looks like it was supposed to work around the behavior you're describing. If that's the case, we should probably remove it -- the `.on` solution seems cleaner anyway.

tests/unit/test_basics.py
```diff
@@ -107,33 +134,115 @@ def assert_user_state_empty():
...
+        res = self.call_handler('studio_submit', self._make_submission(modify_submission))
+        self.assertEqual(res['result'], 'failure')
+        self.assertEqual(len(res['messages']), 1)
+        self.assertIn('should be positive', res['messages'][0])
+
+    def test_studio_submit_max_items_validation(self):
+        def submission_success(submission):
+            submission['max_items_per_zone'] = 1
+            submission['data']['items'] = [
+                {'zones': ['1'], 'title': 'item 1'}, {'zones': ['2'], 'title': 'item 2'}
```
itsjeyd: @e-kolpakov For consistency with the other submission modification methods, should zones be called `'Zone 1'` and `'Zone 2'` here as well?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -280,12 +299,71 @@ def studio_submit(self, submissions, suffix=''):
...
+                errors.extend(self._validate_item_zones(items, max_items_per_zone))
+
+        return errors
+
+    @staticmethod
+    def _validate_item_zones(items, max_items_per_zone):
+        errors = []
+        counter = Counter()
+        for item in items:
+            for zone in item.get('zones', []):
```
itsjeyd: @e-kolpakov This doesn't need special logic for backward compatibility because it operates on item data from `submissions` passed to `studio_submit`, right? If it were operating on existing instance data, the following scenario might cause "item zone validation" to be skipped:

A course author sets `max_items_per_zone` to a positive integer on an old DnDv2 instance with items that still use `item["zone"]` instead of `item["zones"]` and then tries to save it for the first time. Since validation happens before any of the updated data gets saved, `item.get('zones', [])` would return `[]` for each item. As a result, the item count for each zone would be `0`, and the DnDv2 instance would pass validation even if there were more than `max_items_per_zone` items associated with any of the zones.

e-kolpakov: @itsjeyd I don't think an old DnDv2 still using `item['zone']` will have this validation code - not like we're separately deploying backend and frontend here, so when this validation is added, JS will use `item["zones"]` - [it is already in master](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/630a8d7f519409c6c7237660557823998d9e9c29/drag_and_drop_v2/public/js/drag_and_drop_edit.js#L485-L487), while this validation is not.

e-kolpakov:
Rebased from https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92/commits/c1941da6629d38d41b5b2cc8b162b323dc5f80d6

e-kolpakov:
@itsjeyd test failures are from flaky tests again - could you please take a look at this PR again?

itsjeyd:
@e-kolpakov Testing on the sandbox right now :) What version/commit is the sandbox currently using?

itsjeyd:
@e-kolpakov Tested on the sandbox and everything works as advertised. The only thing that needs a couple of small adjustments is the "Maximum items per zone" setting:

* Adding a help text to the Studio editor would be nice. Something like "Upper limit for the number of items that can be associated with, and placed in, a zone. Leave unset if you don't want to constrain the number of items per zone."
* It would be good to have only a single value represent "unlimited items per zone". To be consistent with CAPA problems and other fields that set limits for certain things (e.g. the `max_attempts` field), that should be the empty string/None (and the `min` value for the input field should be `1`). That would open up the possibility for introducing "decoy" zones that don't take any items; but as long as we don't support that feature, manually entering `0` should cause a validation error, just like negative values do.

---

I also looked at your recent updates to the code, they're looking good. The only thing left to address is [this comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92/files#r74302172).

---

Finally, one question about the test instructions: Which unit on the sandbox should I use to verify this item:

> 4. Existing zones with "none" alignment should be treated as having "center"alignment; this should be the case even if an XBlock was not updated in Studio after this PR is applied.

e-kolpakov:
@itsjeyd Addressed your comments, but I don't have time to update the sandbox right now (read: until Monday). So, if you to test this before that - could you please update the sandbox? :) Thank you!

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -287,12 +306,76 @@ def studio_submit(self, submissions, suffix=''):
...
+    def _get_max_items_per_zone(submissions):
+        """
+        Parses Max items per zone value coming from editor.
+
+        Does not catch ValueError intentionally to allow validation to capture this problem
+        """
+        raw_max_items_per_zone = submissions.get('max_items_per_zone', None)
+
+        # tricky condition here: we want None and empty string considered "no value", but integer 0 as "value"
+        # Thus, can't use just `not raw_max_items_per_zone` - integer 0 will be converted to None
```
itsjeyd: @e-kolpakov Did you mean "integer 0 would be converted to False" here?

tests/unit/test_basics.py
```diff
@@ -110,33 +137,127 @@ def assert_user_state_empty():
...
+                {'zones': ['Zone 1'], 'title': 'item 1'}, {'zones': ['Zone 1'], 'title': 'item 2'}
+            ]
+
+        res = self.call_handler('studio_submit', self._make_submission(modify_submission))
+        self.assertEqual(res['result'], 'failure')
+        self.assertEqual(len(res['messages']), 1)
+        self.assertIn("Zone 1", res['messages'][0])
+
+    def test_studio_submit_max_items_validation_failure_multiple_items(self):
+        def submission_failure2(submission):
```
itsjeyd: @e-kolpakov This should be called `modify_submission` to be consistent with the other functions that change the default submission. There's also a couple of functions that serve the same purpose and are called `update_submission` (in L157 and L185 above); it would make sense to rename those as well.

itsjeyd:
@e-kolpakov I updated the sandbox (you won't need to update it again after addressing the last remaining comments since they don't concern the application code) and verified the changes to the "Maximum items per zone" setting. You didn't make the help text for that field visible in the UI, but I'm OK with skipping that for now/leaving the decision to upstream.

The sandbox is now at https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92/commits/0270b0b035d88a350b814ddc6006f3186b3067ad.

So this is *almost* ready to go -- I'll just need to verify this item:

> Existing zones with "none" alignment should be treated as having "center"alignment; this should be the case even if an XBlock was not updated in Studio after this PR is applied.

Which unit on the sandbox should I use to do that? Is there any particular DnDv2 instance that you didn't modify and that should still have "none" alignment stored for it?

e-kolpakov:
@itsjeyd Thank you for updating the sandbox! For testing existing answers - I believe sandbox 3 should be free and not updated to this version, so you could use it to post answers, deploy this PR and check this feature.

On the other hand it might be esaier to test this in devstack :)

itsjeyd:
@e-kolpakov Verified that last item on devstack; it works as expected. The only thing I noticed was that for problems in assessment mode, decoy items got returned to the bank after updating to the code in this branch:

1. Add default DnDv2 problem to unit and set mode to "assessment".
2. Drag items to board in the LMS, including decoy item. Result: All items on board, none in item bank.
3. Switch from `master` to `ekolpakov/non-overlapping-placement`.
4. Reload problem in LMS. Result: Normal items on board (correctly centered in their zones), decoy item back in item bank.

If the changes from this PR are going to go live on edx.org together with the main implementation of assessment mode, I don't expect this to be a problem. But if the changes from this PR land at a later date, the behavior described above should probably be fixed now.

---

One more thing: Please update the README (shouldn't say that item alignment is optional, and shouldn't mention "No alignment is the default", etc.)

itsjeyd:
:+1: once last remaining comments have been addressed ([1](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#discussion_r74623925), [2](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#discussion_r74624004), [3](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#issuecomment-239749633)) and build is green.

- [x] I tested this: Features work as advertised in description
- [x] I read through the code
- [ ] Includes documentation -- **not yet** (will update this comment once that's done)

e-kolpakov:
@itsjeyd 

> The only thing I noticed was that for problems in assessment mode, decoy items got returned to the bank after updating to the code in this branch

That's because this branch is not based on #87, so decoy items are not saved when dropped onto a zone. Also, when I switched to this branch the decoy item stayed where they were - it was returned to bank only after I reset the problem and tried dropping it again.

So, I don't think it will be a problem - this PR have some value on it's own (as it affects standard mode), but since it is not based on #87 assessment mode is still non-functional here. So it shouldn't be deployed without #87.

e-kolpakov:
@itsjeyd rebased from f18278a9e87bd700c57f9a547d279403f2341352

itsjeyd:
@e-kolpakov Thanks for the documentation update, I adjusted my earlier [comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#issuecomment-239750103).

e-kolpakov:
@cahrens, @staubina, @cptvitamin, @sstack22 this PR is ready for your review. 

CI build is red, but only [flaky test failures ](https://travis-ci.org/edx-solutions/xblock-drag-and-drop-v2/builds/152385035) on the last build.

README.md
```diff
@@ -114,12 +114,11 @@ whether or not to display borders outlining the zones. It is possible
...
-Additionally, you can specify the alignment for items once they are dropped in
-the zone.  No alignment is the default, and causes items to stay where the
-learner drops them.  Left alignment causes dropped items to be placed from left
+You can specify the alignment for items once they are dropped in
+the zone.  Centered alignment is the default, and places items from top to bottom
+along the center of the zone. Left alignment causes dropped items to be placed from left
 to right across the zone. Right alignment causes the items to be placed from
-right to left across the zone. Center alignment places items from top to bottom
-along the center of the zone. If left, right, or center alignment is chosen,
+right to left across the zone. If left, right, or center alignment is chosen,
```
cahrens: Reword "**If** left, right, or center...". One of those has to be chosen now.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -110,6 +113,13 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         default="",
     )
 
+    max_items_per_zone = Integer(
```
cahrens: It feels odd that there is a single value for all zones in the problem. One can imagine that some zones are much larger than other zones. But product has decided that this is a valuable option to have?

cahrens:
Display of errors (in Studio) needs UX improvements. Also keyboard focus needs to move to the error notification area.

image

drag_and_drop_v2/public/js/drag_and_drop_edit.js
```diff
@@ -524,8 +521,13 @@ function DragAndDropEditBlock(runtime, element, params) {
                             if (response.result === 'success') {
                                 window.location.reload(false);
                             } else {
-                                $('.xblock-editor-error-message', element)
-                                    .html(gettext('Error: ') + response.message);
+                                var errorMsg = $("<p></p>").text(gettext('Error: '));
+                                $('.xblock-editor-error-message', element).append(errorMsg);
```
cahrens: Note that previous error messages are never cleared out (save with error, save again with same or different error).

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -133,6 +133,9 @@
                     <div id="item-text-color-description" class="item-styles-form-help">
                       {{ help_texts.item_text_color }}
                     </div>
+                    <label class="h3" for="max-items-per-zone" title="{{ help_texts.max_items_per_zone }}">{% trans "Maximum items per zone" %}</label>
+                    <input name="max-items-per-zone" id="max-items-per-zone" type="number" min="1" step="1" value="{{ self.max_items_per_zone|unlocalize }}" />
```
cahrens: Nest input inside of label so you don't need to use ids.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @cahrens I've talked [about that]https://openedx.atlassian.net/browse/SOL-1979?focusedCommentId=206512&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-206512) and there were no objections to implementing this approach. So I would assume product is fine with single setting for all zones. 

@sstack22 - any comments? Just FYI, I'll be away starting Friday and I would really love to land all my PRs (or at least this and #93) before that; implementing per zone max items would make it problematic :)

e-kolpakov:
@cahrens Unfortunately I'm no UX expert, so the best I can do to improve error UX is this:

image

e-kolpakov:
@cahrens, @staubina, @cptvitamin, @sstack22 Same here as on [other PR](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/93#issuecomment-240344038) - I would love to finish it before my vacation, so I would be very grateful if you could help me with it.

@cptvitamin this can be a part of [SOL-1978](https://openedx.atlassian.net/browse/SOL-1978), so I believe we can go without you reviewing/approving this PR right now.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -116,6 +116,7 @@
                 <h3>{% trans "Items" %}</h3>
             </header>
             <section class="tab-content">
+                <div class="xblock-editor-error-messages" aria-live="polite" tabindex="0"></div>
```
cahrens: tabindex should be -1, not 0

cahrens:
Inconsistent spacing between label and input. See how other items have some padding between label and input field.

image

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
cahrens: Doesn't this div need to be present on all the tabs? Right now it is just within "Items".

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -133,6 +134,12 @@
                     <div id="item-text-color-description" class="item-styles-form-help">
                       {{ help_texts.item_text_color }}
                     </div>
+                    <label class="h3" title="{{ help_texts.max_items_per_zone }}">
+                        {% trans "Maximum items per zone" %}
+                        <br/>
```
cahrens: Why br? The styling should be one with CSS.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -133,6 +134,12 @@
...
+                        <input name="max-items-per-zone" id="max-items-per-zone" type="number" min="1" step="1" value="{{ self.max_items_per_zone|unlocalize }}" />
```
cahrens: It is better to use classes instead of IDs.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -133,6 +134,12 @@
...
+                    </label>
+                    <span class="sr">{{ help_texts.max_items_per_zone }}</span>
```
cahrens: Don't think you need to duplicate the title text in a screen reader span (which isn't linked to the input anyway).

cahrens:
@e-kolpakov I don't have time to review the migration and test code (much of which appeared to just move around, making it difficult for me to tell what is new). I am done reviewing the rest of the PR.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
e-kolpakov: @cahrens Fixed aria-live and tabindex.

As for showing this div on all tabs - it only shows error feedback from backend, and this tab is the only one actually sending the feedback. Your suggestion totally makes sense, but it is not necessary now, so I would keep it where it is needed only.

To be honest, I don't like the editor much, so we might want to completely overhaul it (both design and UX) later. But that's my point of view :)

cahrens: That feels very icky-- only showing it on one tab because that's where it is currently used. The JavaScript method that you call to show the notification is generic, correct? Isn't there a general place where it can be shown (perhaps above the "Items" label)? It felt weird anyway that the error was injected into that location anyway. It's great if the error is inline with the control that caused it, but that isn't the case; it was just put into a place that felt a bit arbitrary anyway.

e-kolpakov: @cahrens No, it's not generic - those method collects data from all the "tabs" and sends it; each individual tab does not send anything. The validation on previous tabs are client-side only (btw, I don't like it as well - it is relatively easy to workaround that).

Putting the error next to control is a sensible suggestion, and actually I evaluated that option first; unfortunately, it would require somewhat involving overhaul of editor, so, taking into account deadlines, my vacation and other things I've decided that current solution would be the best I could reliably deliver. 

To explain this decision a bit further, in this particular case those errors are caused by groups of checkboxes that are part of "item" block; there are always more than one such "item" blocks; and so far they are indistinguishable from each other. So, to implement that suggestion I would have to assign some identifiers to the fields, or at least "item" blocks + make sure backend validation returns errors associated with those identifiers somehow.

Since this task is not about improving author experience, I decided to stop right there and follow a simpler approach.

Also, I really love how me being proactive about some possible problems brings just more and more feedback, requesting my little improvement not even mentioned in the task description/acceptance criteria (no one thought about that) to be absolutely bulletproof :)

Also, I've just finished a11y course training, and one of the recommendations was to put errors summary *in the beginning* of the form, so I'll put it there - in the beginning of the form. Visually it wouldn't make any difference though.

Should we bring someone from UX team maybe?

cahrens:
@e-kolpakov If the errors are collected across all tabs, why are they shown on one specific tab? Because it is the last one displayed? But then the errors may not be related to "Items" at all, so placing the div within "Items" doesn't make sense to me.

e-kolpakov:
@cahrens 

> If the errors are collected across all tabs, why are they shown on one specific tab? Because it is the last one displayed? But then the errors may not be related to "Items" at all, so placing the div within "Items" doesn't make sense to me.

Proper solution to this problem will require a significant rework of the editor - for example there are no way to go back to previous tab.

Also, only errors from "Items" tab are validated on server, so unless more validation checks are added, error messages will only contain errors from this tab.

bradenmacdonald:
@e-kolpakov There is already an existing UX for displaying errors built in to studio, which we should use rather than making something idiosyncratic. I believe that all you need to do to show an error message is:

```javascript
runtime.notify('error', {title: gettext("Error title here"), message: gettext("Error message here")});
```

You can use this for client-side validation or you can follow [this pattern](https://github.com/edx/xblock-utils/blob/24f20b8/xblockutils/public/studio_edit.js#L116-L139) to use it for server-side validation.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
bradenmacdonald: I definitely want to avoid re-working the XBlock editor substantially for this PR if we can help it.

If the UX for this particular validation is a problem, I'd rather leave out this warning. Besides, I think there's a bug in the current logic: if I have 10 items and two zones and I limit each zone to 5 items but allow each item to be placed in any zone, then it will raise an error, though that's a perfectly valid problem. Validating this in a way that accounts for such problems is likely going to be overly complex.

e-kolpakov:
@bradenmacdonald That's a questionable improvement:

image

* It is not very well suited to showing multiple error messages, like what we might have here.
* It is much farther from the controls that caused the error than current solution
* It does not address most of the feedback I've got:
** Only shown on "Items" tab
** Does not highlight controls caused the error in any way.

So, looks like removing the validation is a best course of action, especially taking into account [this comment](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#discussion_r75341535) (good catch btw).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -306,6 +317,14 @@ def studio_submit(self, submissions, suffix=''):
         """
         Handles studio save.
         """
+        errors = self._validate_submissions(submissions)
```
cahrens: I thought the validation was being removed.

haikuginger:
@cahrens, validation should now be more thoroughly removed.

drag_and_drop_v2/utils.py
```diff
@@ -78,3 +78,139 @@ def not_placed(number, ngettext=ngettext_fallback):
...
+        Out: {'zone': 'Zone", 'correct': True, 'top': '100px', 'left': '120px'}
+        """
+        # Conversion can't be made as parent dimensions are unknown to python - converted in JS
+        # Since 2.1 JS this conversion became unnecesary, so it was removed from JS code
+        return item
+
+    @classmethod
+    def _item_state_v1p5_to_v2p1(cls, item):
+        """
+        Migrates item_state from v1.5 to v2.5
```
itsjeyd: @haikuginger Following up on https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#discussion-diff-74210508, this should be `v2.0` to `v.2.1`.

README.md
```diff
@@ -122,15 +122,13 @@ whether or not to display borders outlining the zones. It is possible
...
+along the center of the zone. Left alignment causes dropped items to be placed from left
 to right across the zone. Right alignment causes the items to be placed from
-right to left across the zone. Center alignment places items from top to bottom
-along the center of the zone. If left, right, or center alignment is chosen,
-items dropped in a zone will not overlap, but if the zone is not made large
-enough for all its items, they will overflow the bottom of the zone, and
-potentially, overlap the zones below.
+right to left across the zone. Items dropped in a zone will not overlap, 
+but if the zone is not made large enough for all its items, they will overflow the bottom 
+of the zone, and potentially, overlap the zones below.
```
itsjeyd: @haikuginger Remove comma after "potentially".

README.md
```diff
@@ -143,6 +141,9 @@ correct zone, while the error feedback is shown when dropping the item on an
 incorrect drop zone.  In assessment mode, the success and error feedback texts
 are not used.
 
+"Maximum items per Zone" setting controls how much items can be dropped into a
```
itsjeyd: @haikuginger "how many items"

Also, it would make sense to move this paragraph to the end of this section: The paragraphs above and below talk about per-item settings, while this one describes a setting that applies to all items.

drag_and_drop_v2/public/css/drag_and_drop_edit.css
```diff
@@ -371,3 +371,12 @@
...
 }
+
+.xblock--drag-and-drop--editor .xblock-editor-error-messages {
+    color: darkred;
+}
+
+.xblock--drag-and-drop--editor .xblock-editor-error-messages ul li {
+    list-style-type: disc;
+    margin-left: 20px;
+}
```
itsjeyd: @haikuginger The `xblock-editor-error-messages` class doesn't seem to be used anywhere; I think you can remove the rules targeting it.

itsjeyd:
@haikuginger I made another pass over the PR to double-check whether all comments had been addressed. There's a few things that slipped through or should be confirmed after rebasing:

* Inconsistent spacing between label and input ("Maximum item per zone setting"), see https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#issuecomment-240456637. AFAICT, this hasn't been addressed in any of the updates that were pushed after Christina posted the comment.
* Help text for "Maximum items per zone setting" should be displayed in Studio, see https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#issuecomment-239389310. This was deferred to the story for improving help texts ([OC-1778](https://tasks.opencraft.com/browse/OC-1778)) in a follow-up comment (https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#issuecomment-239605793), but the PR for that story (#88) was merged in the meantime, so we'll need to take care of it in this context.
* Decoy items should stay on board after reload, see https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#issuecomment-239749633. As per https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92#issuecomment-239792271, this should work once after rebasing, which will pull in changes from #87.

I will do a final round of manual testing when you're done rebasing and addressing these items, and the sandbox is up-to-date.

haikuginger:
Rebased from 123235a9431f43f23b65188e44ef5c8969ff3e82; comments addressed. @itsjeyd, I'm reprovisioning the sandbox right now.

haikuginger:
@itsjeyd, actually, hold off a bit - there's some stuff that automerge didn't catch.

haikuginger:
@itsjeyd, sorry for the delay. This is set up in the sandbox and ready for review.

drag_and_drop_v2/utils.py
```diff
@@ -78,3 +78,139 @@ def not_placed(number, ngettext=ngettext_fallback):
...
+
+        In: {'zone': 'Zone", 'correct': True, 'top': '100px', 'left': '120px', 'absolute': true}
+        Out: {'zone': 'Zone", 'correct': True}
+
+        In: {'zone': 'Zone", 'correct': True, 'x_percent': '90%', 'y_percent': '20%'}
+        Out: {'zone': 'Zone", 'correct': True}
+        """
+        attributes_to_remove = ['x_percent', 'y_percent', 'left', 'top', 'absolute']
+        for attribute in attributes_to_remove:
+            del[attribute]
```
itsjeyd: @haikuginger This should be `del item[attribute]`, otherwise `item` will be returned unmodified.

mtyaka:
@itsjeyd @haikuginger I just noticed the state migration related changes in this PR - #98 is somewhat related.

haikuginger:
Well spotted, @itsjeyd. I've switched that to a `pop` to avoid any possible KeyErrors.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -127,6 +129,13 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         default="",
     )
 
+    max_items_per_zone = Integer(
+        display_name=_("Maximum items per zone"),
+        help=_("This setting limits the number of items that can be dropped into a single zone"),
```
itsjeyd: @haikuginger Please add a period to make this consistent with help messages for other fields, and update the translation files.

itsjeyd:
@haikuginger I tested on the sandbox, and most everything is working as expected. The only issue I found was:

* Can't save zone alignment: When opening a DnDv2 instance in Studio after changing the alignment of a zone and submitting the changes, the value is back to "center".

I also had a look at the commits you've pushed so far. They look good (although the commit message of https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92/commits/074a8571c69a7515a80d25f1a18249971036827e is fairly cryptic ;))

haikuginger:
@itsjeyd, those changes are deployed. We seem to be down to pretty minor stuff, so if it's all right, I'll wait to update the sandbox until we've got a 👍.

itsjeyd:
Rebased from https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/92/commits/a58f72d3835e6825cdaf523effa6f4f525f3fb12.

itsjeyd:
@haikuginger Thanks for the updates, they fix the problem. In my capacity as sprint firefighter I took the liberty to push this forward some more: I rebased the branch, cleaned up the test code post-rebase, deployed the updated code to the sandbox, and tested everything again according to the PR description (including behavior for legacy zones with "none" alignment -- this I verified in my local dev env, since it requires working with different branches).

This is ready for upstream review now. :+1: 

- [x] I tested this: Features work as advertised in description
- [x] I read through the code
- [x] Includes documentation

cahrens:
@itsjeyd Is this PR ready for TNL re-review?

itsjeyd:
Hi @cahrens, yes, this PR is ready for TNL re-review. Thanks!

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -315,12 +325,34 @@ def studio_submit(self, submissions, suffix=''):
         self.weight = float(submissions['weight'])
         self.item_background_color = submissions['item_background_color']
         self.item_text_color = submissions['item_text_color']
+        # Possible ValueError should be catched in _validate_submissions
```
cahrens: This comment is no longer relevant, correct?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -315,12 +325,34 @@ def studio_submit(self, submissions, suffix=''):
...
         return {
             'result': 'success',
         }
 
+    @staticmethod
+    def _get_max_items_per_zone(submissions):
+        """
+        Parses Max items per zone value coming from editor.
+
+        Does not catch ValueError intentionally to allow validation to capture this problem
```
cahrens: Also not sure what you want to do about this.

cahrens:
Nit: the following scenario feels a bit odd:

1) Set maximum items per zone to 2.
2) Drag 2 items to the zone.
3) Pick up one of the items and release it (back into the same zone).
4) You get an error message about not adding more items to the zone.

I wouldn't expect an error message because I'm not adding new items to the zone, just "moving them around".

e-kolpakov:
@cahrens Notes about comments addressed, showing error message when dropping an item to the same zone is fixed, sandbox is updated

sstack22:
I've finished my review, just one small comment:

- I believe the help text for alignment needs to be updated: "Default is no alignment (items stay exactly where the user drops them)." This is no longer true, correct?

cahrens:
👍 on review feedback commits. I still have not reviewed the migration and test code, but I trust your internal reviews. 

Please do see @sstack22's comment above though about documentation.

e-kolpakov:
@cahrens @sstack22 rebased from 4beaba2 and fixed help text for zone alignment. Updating now.

sstack22:
@e-kolpakov - great, confirmed the fix! 👍

e-kolpakov:
Squashed commits - previous revision fabb8f608b3cae52948b3195bb7a9c9b19b6033b

PR #93, opened 2016-08-11
Assessment mode: display item feedback

e-kolpakov:
**Note:** ~~This PR is based on #87, only relevant commit is the last one.~~

**Description:** This PR allows DnDv2 to show item feedback messages in assessment mode.

**JIRA:** [SOL-1980](https://openedx.atlassian.net/browse/SOL-1980)

**Discussions:** [SOL-1980](https://openedx.atlassian.net/browse/SOL-1980)

**Dependencies:** ~~#87~~

**Sandboxes:**

* LMS: http://dndv2-sandbox8.opencraft.hosting/
* Studio: http://studio-dndv2-sandbox8.opencraft.hosting

**Testing instructions:**
1. Configure DnDv2 to have unique "incorrect" item feedback messages. Set to assessment mode.
2. In LMS, put some items into *wrong* zones.
3. Check answer (click "Submit")
4. Feedback popup should appear showing incorrect feedback messages for every misplaced item.
5. ~~If only one item is misplaced, only feedback message is shown; if multiple items are misplaced,~~ popup contains a note about ~~multiple~~ item(s) being misplaced, and feedback messages are shown in form of bullet list.

**Screenshots:**

image

**Reviewers:**
- [x] internal: @itsjeyd 
- [x] internal: @haikuginger
- [x] TNL: @cahrens 
- [x] TNL: @staubina
- [ ] ~~a11y: @cptvitamin~~ a11y review will be performed later, for all changes in v2.1: [SOL-1978](https://openedx.atlassian.net/browse/SOL-1978)
- [x] Product: @sstack22

**Checklist:**
- [x] Internal review
- [x] Upstream review
- [x] Squash commits

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,8 +138,8 @@ image. You can define custom success and error feedback for each item. In
 standard mode, the feedback text is displayed in a popup after the learner drops
 the item on a zone - the success feedback is shown if the item is dropped on a
 correct zone, while the error feedback is shown when dropping the item on an
-incorrect drop zone.  In assessment mode, the success and error feedback texts
-are not used.
+incorrect drop zone.  In assessment mode, the success feedback texts
+are not used, while error feedback texts are shown when learner submits a solution.
 
 You can select any number of zones for an item to belong to using
 the checkboxes; all zones defined in the previous step are available.
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -246,7 +246,6 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
             "target_img_description": self.target_img_description,
             "item_background_color": self.item_background_color or None,
             "item_text_color": self.item_text_color or None,
-            "initial_feedback": self.data['feedback']['start'],
             # final feedback (data.feedback.finish) is not included - it may give away answers.
         }
 
@@ -392,17 +391,29 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         self._validate_do_attempt()
 
         self.attempts += 1
-        self._mark_complete_and_publish_grade()  # must happen before _get_feedback
+        # pylint: disable=fixme
+        # TODO: Refactor this method to "freeze" item_state and pass it to methods that need access to it.
+        # These implicit dependencies between methods exist because most of them use `item_state` or other
+        # fields, either as an "input" (i.e. read value) or as output (i.e. set value) or both. As a result,
+        # incorrect order of invocation causes issues:
+        self._mark_complete_and_publish_grade()  # must happen before _get_feedback - sets grade
+        correct = self._is_answer_correct()  # must happen before manipulating item_state - reads item_state
 
-        overall_feedback_msgs, misplaced_ids = self._get_feedback()
+        overall_feedback_msgs, misplaced_ids = self._get_feedback(include_item_feedback=True)
 
+        misplaced_items = []
         for item_id in misplaced_ids:
             del self.item_state[item_id]
+            misplaced_items.append(self._get_item_definition(int(item_id)))
+
+        feedback_msgs = [FeedbackMessage(item['feedback']['incorrect'], None) for item in misplaced_items]
 
         return {
+            'correct': correct,
             'attempts': self.attempts,
             'misplaced_items': list(misplaced_ids),
-            'overall_feedback': self._present_overall_feedback(overall_feedback_msgs)
+            'feedback': self._present_feedback(feedback_msgs),
+            'overall_feedback': self._present_feedback(overall_feedback_msgs)
         }
 
     @XBlock.json_handler
@@ -487,7 +498,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
                 self.i18n_service.gettext("Max number of attempts reached")
             )
 
-    def _get_feedback(self):
+    def _get_feedback(self, include_item_feedback=False):
         """
         Builds overall feedback for both standard and assessment modes
         """
@@ -510,16 +521,14 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
                 message = message_template(len(ids_list), self.i18n_service.ngettext)
                 feedback_msgs.append(FeedbackMessage(message, message_class))
 
-        _add_msg_if_exists(
-            items.correctly_placed, FeedbackMessages.correctly_placed, FeedbackMessages.MessageClasses.CORRECTLY_PLACED
-        )
-        _add_msg_if_exists(misplaced_ids, FeedbackMessages.misplaced, FeedbackMessages.MessageClasses.MISPLACED)
-        _add_msg_if_exists(missing_ids, FeedbackMessages.not_placed, FeedbackMessages.MessageClasses.NOT_PLACED)
-
-        if misplaced_ids and self.attempts_remain:
-            feedback_msgs.append(
-                FeedbackMessage(FeedbackMessages.MISPLACED_ITEMS_RETURNED, None)
+        if self.item_state or include_item_feedback:
+            _add_msg_if_exists(
+                items.correctly_placed,
+                FeedbackMessages.correctly_placed,
+                FeedbackMessages.MessageClasses.CORRECTLY_PLACED
             )
+            _add_msg_if_exists(misplaced_ids, FeedbackMessages.misplaced, FeedbackMessages.MessageClasses.MISPLACED)
+            _add_msg_if_exists(missing_ids, FeedbackMessages.not_placed, FeedbackMessages.MessageClasses.NOT_PLACED)
 
         if self.attempts_remain and (misplaced_ids or missing_ids):
             problem_feedback_message = self.data['feedback']['start']
@@ -539,7 +548,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         return feedback_msgs, misplaced_ids
 
     @staticmethod
-    def _present_overall_feedback(feedback_messages):
+    def _present_feedback(feedback_messages):
         """
         Transforms feedback messages into format expected by frontend code
         """
@@ -563,14 +572,14 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         self._publish_item_dropped_event(item_attempt, is_correct)
 
         item_feedback_key = 'correct' if is_correct else 'incorrect'
-        item_feedback = item['feedback'][item_feedback_key]
+        item_feedback = FeedbackMessage(item['feedback'][item_feedback_key], None)
         overall_feedback, __ = self._get_feedback()
 
         return {
             'correct': is_correct,
             'finished': self._is_answer_correct(),
-            'overall_feedback': self._present_overall_feedback(overall_feedback),
-            'feedback': item_feedback
+            'overall_feedback': self._present_feedback(overall_feedback),
+            'feedback': self._present_feedback([item_feedback])
         }
 
     def _drop_item_assessment(self, item_attempt):
@@ -612,6 +621,18 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
         Helper method to update `self.completed` and submit grade event if appropriate conditions met.
         """
+        # pylint: disable=fixme
+        # TODO: (arguable) split this method into "clean" functions (with no side effects and implicit state)
+        # This method implicitly depends on self.item_state (via _is_answer_correct and _get_grade)
+        # and also updates self.grade if some conditions are met. As a result this method implies some order of
+        # invocation:
+        # * it should be called after learner-caused updates to self.item_state is applied
+        # * it should be called before self.item_state cleanup is applied (i.e. returning misplaced items to item bank)
+        # * it should be called before any method that depends on self.grade (i.e. self._get_feedback)
+
+        # Splitting it into a "clean" functions will allow to capture this implicit invocation order in caller method
+        # and help avoid bugs caused by invocation order violation in future.
+
         # There's no going back from "completed" status to "incomplete"
         self.completed = self.completed or self._is_answer_correct() or not self.attempts_remain
         grade = self._get_grade()
@@ -694,7 +715,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
             'items': item_state,
             'finished': is_finished,
             'attempts': self.attempts,
-            'overall_feedback': self._present_overall_feedback(overall_feedback_msgs)
+            'overall_feedback': self._present_feedback(overall_feedback_msgs)
         }
 
     def _get_item_state(self):
@@ -794,8 +815,8 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
         Returns the student's grade for this block.
         """
-        correct_count, required_count = self._get_item_stats()
-        return correct_count / float(required_count) * self.weight
+        correct_count, total_count = self._get_item_stats()
+        return correct_count / float(total_count) * self.weight
 
     def _answer_correctness(self):
         """
@@ -807,8 +828,8 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
                 * Partial: Some items are at their correct place.
                 * Incorrect: None items are at their correct place.
         """
-        correct_count, required_count = self._get_item_stats()
-        if correct_count == required_count:
+        correct_count, total_count = self._get_item_stats()
+        if correct_count == total_count:
             return self.SOLUTION_CORRECT
         elif correct_count == 0:
             return self.SOLUTION_INCORRECT
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -367,9 +367,73 @@ function DragAndDropTemplates(configuration) {
         )
     };
 
+    var itemFeedbackPopupTemplate = function(ctx) {
+        var popupSelector = 'div.popup.item-feedback-popup.popup-wrapper';
+        var msgs = ctx.feedback_messages || [];
+        var have_messages = msgs.length > 0;
+        var popup_content;
+
+        var close_button_describedby_id = "close-popup-"+configuration.url_name;
+
+        if (msgs.length > 0 && !ctx.last_action_correct) {
+            popupSelector += '.popup-incorrect';
+        }
+
+        if (ctx.mode == DragAndDropBlock.ASSESSMENT_MODE) {
+            var content_items = [
+                (!ctx.last_action_correct) ? h("p", {}, gettext("Some of your answers were not correct.")) : null,
+                h("p", {}, gettext("Hints:")),
+                h("ul", {}, msgs.map(function(message) {
+                    return h("li", {innerHTML: message.message});
+                }))
+            ];
+            popup_content = h("div.popup-content", {}, have_messages ? content_items : []);
+        } else {
+            popup_content = h("div.popup-content", {}, msgs.map(function(message) {
+                return h("p", {innerHTML: message.message});
+            }))
+        }
+
+        return h(
+            popupSelector,
+            {
+                style: {display: have_messages ? 'block' : 'none'},
+                attributes: {
+                    "tabindex": "-1",
+                    'aria-live': 'polite',
+                    'aria-atomic': 'true',
+                    'aria-relevant': 'additions',
+                }
+            },
+            [
+                h(
+                    'button.unbutton.close-feedback-popup-button',
+                    {},
+                    [
+                        h(
+                            'span.sr',
+                            {
+                                innerHTML: gettext("Close item feedback popup")
+                            }
+                        ),
+                        h(
+                            'span.icon.fa.fa-times-circle',
+                            {
+                                attributes: {
+                                    'aria-hidden': true
+                                }
+                            }
+                        )
+                    ]
+                ),
+                popup_content
+            ]
+        )
+    };
+
     var mainTemplate = function(ctx) {
-        var problemTitle = ctx.show_title ? h('h2.problem-title', {innerHTML: ctx.title_html}) : null;
-        var problemHeader = ctx.show_problem_header ? h('h3.title1', gettext('Problem')) : null;
+        var problemTitle = ctx.show_title ? h('h3.problem-title', {innerHTML: ctx.title_html}) : null;
+        var problemHeader = ctx.show_problem_header ? h('h4.title1', gettext('Problem')) : null;
 
         // Render only items_in_bank and items_placed_unaligned here;
         // items placed in aligned zones will be rendered by zoneTemplate.
@@ -401,7 +465,7 @@ function DragAndDropTemplates(configuration) {
                     h('div.target',
                         {},
                         [
-                            popupTemplate(ctx),
+                            itemFeedbackPopupTemplate(ctx),
                             h('div.target-img-wrapper', [
                                 h('img.target-img', {src: ctx.target_img_src, alt: ctx.target_img_description}),
                             ]
@@ -427,6 +491,11 @@ function DragAndDropBlock(runtime, element, configuration) {
 
     DragAndDropBlock.STANDARD_MODE = 'standard';
     DragAndDropBlock.ASSESSMENT_MODE = 'assessment';
+
+    var Selector = {
+        popup_box: '.popup',
+        close_button: '.popup .close-feedback-popup-button'
+    };
 
     var renderView = DragAndDropTemplates(configuration);
 
@@ -482,7 +551,7 @@ function DragAndDropBlock(runtime, element, configuration) {
 
             // Set up event handlers:
 
-            $(document).on('keydown mousedown touchstart', closePopup);
+            $element.on('click', '.item-feedback-popup .close-feedback-popup-button', closePopupEventHandler);
             $element.on('click', '.submit-answer-button', doAttempt);
             $element.on('click', '.keyboard-help-button', showKeyboardHelp);
             $element.on('keydown', '.keyboard-help-button', function(evt) {
@@ -646,12 +715,23 @@ function DragAndDropBlock(runtime, element, configuration) {
     /**
      * Update the DOM to reflect 'state'.
      */
-    var applyState = function() {
+    var applyState = function(keepDraggableInit) {
+        sendFeedbackPopupEvents();
+        updateDOM();
+        if (!keepDraggableInit) {
+            destroyDraggable();
+            if (!state.finished) {
+                initDraggable();
+            }
+        }
+    };
+
+    var sendFeedbackPopupEvents = function() {
         // Has the feedback popup been closed?
         if (state.closing) {
             var data = {
                 event_type: 'edx.drag_and_drop_v2.feedback.closed',
-                content: previousFeedback || state.feedback,
+                content: concatenateFeedback(previousFeedback || state.feedback),
                 manually: state.manually_closed,
             };
             truncateField(data, 'content');
@@ -663,17 +743,15 @@ function DragAndDropBlock(runtime, element, configuration) {
         if (state.feedback) {
             var data = {
                 event_type: 'edx.drag_and_drop_v2.feedback.opened',
-                content: state.feedback,
+                content: concatenateFeedback(state.feedback),
             };
             truncateField(data, 'content');
             publishEvent(data);
         }
+    };
 
-        updateDOM();
-        destroyDraggable();
-        if (!state.finished) {
-            initDraggable();
-        }
+    var concatenateFeedback = function (feedback_msgs_list) {
+        return feedback_msgs_list.map(function(message) { return message.message; }).join('\n');
     };
 
     var updateDOM = function(state) {
@@ -739,6 +817,15 @@ function DragAndDropBlock(runtime, element, configuration) {
         $root.find('.item-bank .option').first().focus();
     };
 
+    var focusItemFeedbackPopup = function() {
+        var popup = $root.find('.item-feedback-popup');
+        if (popup.length && popup.is(":visible")) {
+            popup.focus();
+            return true;
+        }
+        return false;
+    };
+
     var placeItem = function($zone, $item) {
         var item_id;
         if ($item !== undefined) {
@@ -753,7 +840,7 @@ function DragAndDropBlock(runtime, element, configuration) {
         var items_in_zone_count = countItemsInZone(zone, [item_id.toString()]);
         if (configuration.max_items_per_zone && configuration.max_items_per_zone <= items_in_zone_count) {
             state.last_action_correct = false;
-            state.feedback = gettext("You cannot add any more items to this zone.");
+            state.feedback = [{message: gettext("You cannot add any more items to this zone."), message_class: null}];
             applyState();
             return;
         }
@@ -763,6 +850,7 @@ function DragAndDropBlock(runtime, element, configuration) {
             zone_align: zone_align,
             submitting_location: true,
         };
+
         // Wrap in setTimeout to let the droppable event finish.
         setTimeout(function() {
             applyState();
@@ -895,13 +983,16 @@ function DragAndDropBlock(runtime, element, configuration) {
     var grabItem = function($item, interaction_type) {
         var item_id = $item.data('value');
         setGrabbedState(item_id, true, interaction_type);
-        updateDOM();
+        closePopup(false);
+        // applyState(true) skips destroying and initializing draggable
+        applyState(true);
     };
 
     var releaseItem = function($item) {
         var item_id = $item.data('value');
         setGrabbedState(item_id, false);
-        updateDOM();
+        // applyState(true) skips destroying and initializing draggable
+        applyState(true);
     };
 
     var setGrabbedState = function(item_id, grabbed, interaction_type) {
@@ -967,31 +1058,31 @@ function DragAndDropBlock(runtime, element, configuration) {
             });
     };
 
-    var closePopup = function(evt) {
+    var closePopupEventHandler = function(evt) {
         if (!state.feedback) {
             return;
         }
 
         var target = $(evt.target);
-        var popup_box = '.xblock--drag-and-drop .popup';
-        var close_button = '.xblock--drag-and-drop .popup .close';
 
-        if (target.is(popup_box)) {
+        if (target.is(Selector.popup_box)) {
             return;
         }
-        if (target.parents(popup_box).length && !target.is(close_button)) {
+        if (target.parents(Selector.popup_box).length && !target.parent().is(Selector.close_button) && !target.is(Selector.close_button)) {
             return;
         }
 
-        state.closing = true;
-        previousFeedback = state.feedback;
-        if (target.is(close_button)) {
-            state.manually_closed = true;
-        } else {
-            state.manually_closed = false;
-        }
-
+        closePopup(target.is(Selector.close_button) || target.parent().is(Selector.close_button));
         applyState();
+    };
+
+    var closePopup = function(manually_closed) {
+        // do not apply state here - callers are responsible to call it when other appropriate state changes are applied
+        if ($root.find(Selector.popup_box).is(":visible")) {
+            state.closing = true;
+            previousFeedback = state.feedback;
+            state.manually_closed = manually_closed;
+        }
     };
 
     var resetProblem = function(evt) {
@@ -1018,7 +1109,10 @@ function DragAndDropBlock(runtime, element, configuration) {
             data: '{}'
         }).done(function(data){
             state.attempts = data.attempts;
+            state.feedback = data.feedback;
             state.overall_feedback = data.overall_feedback;
+            state.last_action_correct = data.correct;
+
             if (attemptsRemain()) {
                 data.misplaced_items.forEach(function(misplaced_item_id) {
                     delete state.items[misplaced_item_id]
@@ -1026,15 +1120,15 @@ function DragAndDropBlock(runtime, element, configuration) {
             } else {
                 state.finished = true;
             }
-            focusFirstDraggable();
         }).always(function() {
             state.submit_spinner = false;
             applyState();
+            focusItemFeedbackPopup() || focusFirstDraggable();
         });
     };
 
     var canSubmitAttempt = function() {
-        return Object.keys(state.items).length > 0 && attemptsRemain();
+        return Object.keys(state.items).length > 0 && attemptsRemain() && !submittingLocation();
     };
 
     var canReset = function() {
@@ -1056,6 +1150,15 @@ function DragAndDropBlock(runtime, element, configuration) {
     var attemptsRemain = function() {
         return !configuration.max_attempts || configuration.max_attempts > state.attempts;
     };
+
+    var submittingLocation = function() {
+        var result = false;
+        Object.keys(state.items).forEach(function(item_id) {
+            var item = state.items[item_id];
+            result = result || item.submitting_location;
+        });
+        return result;
+    }
 
     var render = function() {
         var items = configuration.items.map(function(item) {
@@ -1114,7 +1217,6 @@ function DragAndDropBlock(runtime, element, configuration) {
             show_title: configuration.show_title,
             mode: configuration.mode,
             max_attempts: configuration.max_attempts,
-            attempts: state.attempts,
             problem_html: configuration.problem_text,
             show_problem_header: configuration.show_problem_header,
             show_submit_answer: configuration.mode == DragAndDropBlock.ASSESSMENT_MODE,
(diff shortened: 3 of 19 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=93&repo=edx-solutions%2Fxblock-drag-and-drop-v2

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -221,7 +221,6 @@ def items_without_answers():
             "target_img_description": self.target_img_description,
             "item_background_color": self.item_background_color or None,
             "item_text_color": self.item_text_color or None,
-            "initial_feedback": self.data['feedback']['start'],
```
itsjeyd: @e-kolpakov I probably missed some of the more recent discussions about "Introductory Feedback" in assessment mode, but don't we at least need to pass this to the client for problems in "standard" mode?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -642,7 +653,10 @@ def _get_item_definition(self, item_id):
         """
         Returns definition (settings) for item identified by `item_id`.
         """
-        return next(i for i in self.data['items'] if i['id'] == item_id)
+        try:
+            return next(i for i in self.data['items'] if i['id'] == item_id)
+        except StopIteration:
+            raise Exception(self.data['items'], item_id)
```
itsjeyd: @e-kolpakov Under what conditions do we expect this exception to fire? It might be worth adding comment about that here.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -227,12 +227,17 @@ function DragAndDropTemplates(configuration) {
     };
 
     var feedbackTemplate = function(ctx) {
-        var feedback_display = ctx.feedback_html ? 'block' : 'none';
         var properties = { attributes: { 'aria-live': 'polite' } };
+        var messages = ctx.overall_feedback_messages || [];
+        var feedback_display = messages.length > 0 ? 'block' : 'none';
+        var feedback_messages = messages.map(function(message) {
+            return h("p.message", {innerHTML: message}, []);
```
itsjeyd: @e-kolpakov Nit: I don't think this needs to pass `[]` to `h` to indicate that the element shouldn't have any children. Not passing a third argument has the same effect.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -312,13 +317,42 @@ function DragAndDropTemplates(configuration) {
         )
     };
 
-    var mainTemplate = function(ctx) {
-        var problemTitle = ctx.show_title ? h('h2.problem-title', {innerHTML: ctx.title_html}) : null;
-        var problemHeader = ctx.show_problem_header ? h('h3.title1', gettext('Problem')) : null;
+    var itemFeedbackPopupTemplate = function(ctx) {
         var popupSelector = 'div.popup';
-        if (ctx.popup_html && !ctx.last_action_correct) {
+        var msgs = (ctx.feedback_messages || []).filter(function(msg) { return (msg) ? true : false; });
```
itsjeyd: @e-kolpakov IIUC, empty messages are also getting filtered out on the server (via `_suppress_empty_messages`). If that's the case you can remove the `.filter` portion from this line I think.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -312,13 +317,42 @@ function DragAndDropTemplates(configuration) {
...
+
+        if (msgs.length > 1) {
+            popup_content = h("div.popup-content", {}, [
+                (!ctx.last_action_correct) ? h("p", {}, gettext("Some of your answers were not correct")) : null,
+                h("p", {}, gettext("Hints:")),
+                h("ul", {}, msgs.map(function(message) {
+                    return h("li", {innerHTML: message}, []);
+                }))
+            ])
+        } else {
```
itsjeyd: @e-kolpakov Did we discuss with upstream whether feedback behavior should be different if only a single item was misplaced?

(Personally, I'd prefer feedback to behave consistently, regardless of the number of items that got misplaced. If the popup doesn't include a summary-style message like the one we're using for multiple misplaced items, it feels a bit like a "delayed hint" that would have been more appropriate right when the offending item was placed on the board.)

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -973,6 +1002,17 @@ function DragAndDropBlock(runtime, element, configuration) {
         });
     };
 
+    var wrapMessage = function(element, message) {
+        return "<"+element+">"+message+"</"+element+">";
```
itsjeyd: @e-kolpakov Nit: Add whitespace around operators

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -973,6 +1002,17 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+    };
+
+    var concatFeedbackMessages = function(messages, wrapper_element, prefix) {
```
itsjeyd: @e-kolpakov This doesn't seem to be used anywhere at the moment.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -984,7 +1024,13 @@ function DragAndDropBlock(runtime, element, configuration) {
             data: '{}'
         }).done(function(data){
             state.attempts = data.attempts;
-            state.overall_feedback = data.feedback;
+            state.feedback = data.feedback;
+            state.overall_feedback = data.overall_feedback;
+
+            if (data.feedback) {
+                state.last_action_correct = false;
```
itsjeyd: @e-kolpakov I'm not sure if `last_action_correct` is the most appropriate name in this context. The last action that the user performed was to submit an answer; we might decide that that doesn't count, but the last "drop item" action before that might still have been correct. Rename to `all_items_correct`, perhaps?

tests/integration/test_interaction.py
```diff
@@ -657,6 +656,23 @@ def test_do_attempt_feedback_is_updated(self):
         expected_feedback = "\n".join(feedback_lines)
         self.assertEqual(self._get_feedback().text, expected_feedback)
 
+    def test_multiple_misplaced_item_feedback(self):
```
itsjeyd: @e-kolpakov Could you rename this test to something like `test_per_item_feedback`, to distinguish it from other tests that deal with "global" feedback displayed on submit?

Also, it would probably be worth extending this test to cover scenarios where

* no item was misplaced
* a single item was misplaced

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @itsjeyd as far as I can tell "initial_feedback" is not used anywhere in JS.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @itsjeyd debug leftover

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @itsjeyd no, we didn't. This would be relatively simple to address, so I would suggest to keep this as is now, and draw upstream reviewers attention to this feature - than if they think it should behave differently, I'll replace it.

tests/integration/test_interaction.py
e-kolpakov: @itsjeyd I believe those two scenarios are covered by `parameterized_something_...` tests.

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @itsjeyd Having thought about this, I believe this is exactly right name for it. `last_action_correct` is used in `itemFeedbackPopupTemplate` to choose popup class - in that case we're showing a popup for last action; checking answer in assessment mode and dropping an item in standard mode.

e-kolpakov:
@itsjeyd Just once [no tests failed](https://travis-ci.org/edx-solutions/xblock-drag-and-drop-v2/builds/151797454), but I didn't notice pep8 errors :(

e-kolpakov:
@itsjeyd Sandbox updated

e-kolpakov:
@itsjeyd I might have updated this sandbox with the version intended for other PR (sandbox9) - if you're to test it while I'm away, could you please reinstall DnDv2 on the sandbox? Thank you!

haikuginger:
The sandbox looks to be correctly provisioned at this point.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -320,28 +320,30 @@ function DragAndDropTemplates(configuration) {
...
         var popup_content;
 
         if (msgs.length > 0 && !ctx.last_action_correct) {
             popupSelector += '.popup-incorrect';
         }
 
-        if (msgs.length > 1) {
-            popup_content = h("div.popup-content", {}, [
+        if (ctx.mode == DragAndDropBlock.ASSESSMENT_MODE) {
+            var content_items = [
```
itsjeyd: @e-kolpakov `message_items` would be a little more to the point I think.

Thanks for making the behavior consistent!

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -320,28 +320,30 @@ function DragAndDropTemplates(configuration) {
...
                 h("p", {}, gettext("Hints:")),
                 h("ul", {}, msgs.map(function(message) {
                     return h("li", {innerHTML: message}, []);
                 }))
-            ])
+            ];
+            popup_content = h("div.popup-content", {}, have_messages ? content_items : []);
         } else {
             popup_content = h("div.popup-content", {}, msgs.map(function(message) {
                 return h("p", {innerHTML: message}, []);
```
itsjeyd: @e-kolpakov Nit: Remove last argument (`[]`) from call to `h`. Same comment applies to L335 above.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1003,7 +1003,7 @@ function DragAndDropBlock(runtime, element, configuration) {
     };
 
     var wrapMessage = function(element, message) {
-        return "<"+element+">"+message+"</"+element+">";
+        return "<" + element + ">" + message + "</" + element + ">";
     };
 
     var concatFeedbackMessages = function(messages, wrapper_element, prefix) {
```
itsjeyd: @e-kolpakov I still think you can remove this method. When you do, `wrapMessage` will probably become obsolete as well.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -312,13 +317,42 @@ function DragAndDropTemplates(configuration) {
...
+        var msgs = (ctx.feedback_messages || []).filter(function(msg) { return (msg) ? true : false; });
+        var popup_content;
+
+        if (msgs.length > 0 && !ctx.last_action_correct) {
             popupSelector += '.popup-incorrect';
         }
+
+        if (msgs.length > 1) {
+            popup_content = h("div.popup-content", {}, [
+                (!ctx.last_action_correct) ? h("p", {}, gettext("Some of your answers were not correct")) : null,
```
itsjeyd: @e-kolpakov As part of applying finishinging touches, it would be nice if the main message ("Some of you answers ...") ended with a period/full stop.

itsjeyd:
Thanks @haikuginger!

@e-kolpakov I tested the feature on the sandbox, and it's working well. My only comment is that it would be nice to adjust the padding of the feedback popup so that long messages don't get a chance to run all the way to the right border:

long-hint

It would look better if the padding on the right (approximately) matched the padding on the left (space between left edge of popup and bullet).

---

Other than that, the documentation will need to be updated. It currently says:

> In assessment mode, the success and error feedback texts are not used.

... which is not true anymore as of this PR :)

---

:+1: once remaining comments have been addressed and build is green.

- [x] I tested this: Feature works as advertised in PR description
- [x] I read through the code
- [ ] Includes documentation -- **not yet** (will update this comment once that's done)

e-kolpakov:
@itsjeyd I've addressed your notes and am updating sandbox right now - if you'd like another look at this this is the right time; I'll ping upstream reviewers in about half an hour.

README.md
```diff
@@ -141,7 +141,7 @@ standard mode, the feedback text is displayed in a popup after the learner drops
 the item on a zone - the success feedback is shown if the item is dropped on a
 correct zone, while the error feedback is shown when dropping the item on an
 incorrect drop zone.  In assessment mode, the success feedback texts
-are not used, while error feedback texts are shown when learner submits a solution.
+are not used, while error feedback texts are shown when learner submits a solution.3
```
itsjeyd: @e-kolpakov I think this got added accidentally?

itsjeyd:
@e-kolpakov The popup looks good! But now I'm seeing this when submitting an answer on the sandbox (instead of proper feedback):

feedback-boxes

Probably a regression caused by the rebase?

e-kolpakov:
@itsjeyd squashed from 2b1ca2d4f0f11a4728acd769401afd26805a6316

e-kolpakov:
@itsjeyd Issue with feedback is now fixed - I've changed how feedback messages are sent to frontend (from plain string to `{"message": ..., "message_class": ...}`) and python updates were not picked yet. Now it works ok.

e-kolpakov:
@cahrens, @staubina, @cptvitamin, @sstack22 this PR is ready for your review.

CI build is read, but [only flaky test failures are on the last build](https://travis-ci.org/edx-solutions/xblock-drag-and-drop-v2/builds/152393376), failed tests pass locally (almost always).

pdesjardins:
@e-kolpakov, will error feedback be optional in assessment mode? I can imagine course teams wanting to suppress error feedback for a drag and drop problem that is part of an exam.

e-kolpakov:
@pdesjardins both success and error feedback messages are optional, in standard and assessment mode.

pdesjardins:
@e-kolpakov, thanks for clarifying. It looks like an empty string in the error feedback message suppresses it. Also, the feedback isn't immediate, it occurs only after the learner submits.

e-kolpakov:
@pdesjardins Yes, exactly - empty string suppresses the feedback and in assessment mode it is only shown after submitting an attempt.

e-kolpakov:
@cahrens, @staubina, @cptvitamin, @sstack22 just FYI, I'll be away starting Friday, so your's "today" is the last day I can meaningfully address any comments. I would really love to go on vacation knowing I've finished all my tasks - please help me achieve that. :)

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -353,14 +352,18 @@ def do_attempt(self, data, suffix=''):
         self._mark_complete_and_publish_grade()  # must happen before _get_feedback
 
         overall_feedback_msgs, misplaced_ids = self._get_feedback()
+        misplaced_items = [self._get_item_definition(int(item_id)) for item_id in misplaced_ids]
```
staubina: Is the int() required around item_id?

Also could you combine this and the loop on line 358 into one loop?
You are looping through misplaced_items 2 times.

staubina:
I know accessibility will do a review, and I am still looking in the code for this, but is focus shifted to the feedback when it is shown on the drag and drop?  Also does it contain the necessary elements for a screen reader to read it?

staubina:
Please correct the build and test issues. 

Overall the changes look good to me.  Some comments and questions mentioned above.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @staubina yes, as far as I know, `self.item_state` stores items' IDs as strings, while `self.data['items']` uses integers. I don't think this is a good time to modify it so both `self.data['items']` and `self.item_state` use ints, as those will affect existing DnDv2 instances.

e-kolpakov:
@staubina Item feedback popup is a child of a "region" with `aria-live: polite` attribute, so it is read to screen readers. As of focusing the popup - I would wait for accessibility team to weigh in: there are two feedback areas: "item" feedback (the popup) and overall feedback (below the problem) - thus it is not actually clear which one of those should receive focus. So far, after submitting an answer in assessment mode, focus returns to item bank (this was introduced in #87).

Also, we might want to use some `aria-role` on the popup; `alert` seems like a good one, but it bears a little to "alarming" meaning - the feedback is not time-sensitive and actually is not an alert - just an additional information for the learner. So, again, since I'm not an expert, I would wait for a11y team advice on this.

e-kolpakov:
@staubina as of test failures - these are known flaky tests; they pass locally 95% of time (and 100% of time when run one by one) + we have done a couple of fruitless attempts to fix them last couple of weeks. Since this DnDv2.1 project is on a tight deadline we kind of made a deal with our conscience and do not require the green build for a while.

I don't have actual info about deadlines and remaining work, so we might want to switch back to green build policy, but it will require some time to debug and address the problems. @bradenmacdonald are there any considerations about when we stop merging PRs with red builds?

cahrens:
@staubina thanks for taking a pass through this. I will try to look today as well @e-kolpakov. I was unexpectedly out yesterday, and like you, I am on vacation starting on Friday.

staubina:
Thank you for the updates and changes.  I will also wait on the a11y review, it was just a thought as I was using the sandbox.

e-kolpakov:
@staubina sure, thank you for sharing your thoughts!

@cahrens no worries about yesterday!

cahrens:
This has accessibility issues-- when you press Submit (via keyboard, for example), the focus goes to the problem bank instead of the feedback message. You can't even tab to the feedback zone because it disappears on the next tab.

When feedback is shown, my understanding is that it either needs to be read out with aria-live, or keyboard focus should move to the feedback area. FYI @cptvitamin

I also think the message disappears too aggressively. I would expect it to stay until you press the close button, or reload the page. Clicking and tabbing seems to make it go away.

drag_and_drop_v2/drag_and_drop_v2.py
cahrens: @e-kolpakov so should "start" feedback be deleted entirely (Python, studio)?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -321,13 +321,44 @@ function DragAndDropTemplates(configuration) {
...
+        } else {
+            popup_content = h("div.popup-content", {}, msgs.map(function(message) {
+                return h("p", {innerHTML: message.message});
+            }))
+        }
+
+        return h(
+            popupSelector, {style: {display: have_messages ? 'block' : 'none'} },
+            [
+                h('div.close.icon-remove-sign.fa-times-circle'),
```
cahrens: Needs to have aria-hidden with appropriate sr span explaining what the "button" does. 

Also needs an actual button.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -321,13 +321,44 @@ function DragAndDropTemplates(configuration) {
...
+                popup_content
+            ]
+        )
+    };
+
+    var mainTemplate = function(ctx) {
+        var problemTitle = ctx.show_title ? h('h2.problem-title', {innerHTML: ctx.title_html}) : null;
```
cahrens: This needs to be an h3 (and the one below an h4). There is already an h2 on the page (course header). If you look at other components, you see that their display names are h3 instead of h2.

tests/integration/test_interaction.py
```diff
@@ -487,14 +482,17 @@ class AssessmentTestMixin(object):
     Provides helper methods for assessment tests
     """
     @staticmethod
-    def _wait_until_enabled(element):
+    def _wait_until_enabled(element, message=None):
         wait = WebDriverWait(element, 2)
-        wait.until(lambda e: e.is_displayed() and e.get_attribute('disabled') is None)
+        wait.until(lambda e: e.is_displayed() and e.get_attribute('disabled') is None, message)
+
+    def _just_wait(self, seconds):
```
cahrens: I don't think this is doing what you expect. And instead of waiting for a specified amount of time, it is much better to wait for an element to be in its expected state.

http://www.seleniumhq.org/docs/04_webdriver_advanced.jsp
`An implicit wait is to tell WebDriver to poll the DOM for a certain amount of time when trying to find an element or elements if they are not immediately available. The default setting is 0. Once set, the implicit wait is set for the life of the WebDriver object instance.`

cahrens:
@e-kolpakov done reviewing

sstack22:
@e-kolpakov - feedback is looking great! A couple comments/questions:

1) When you reset a problem, you automatically get the feedback that you "did not place 4 required items." When a learner resets the problem, I would not expect that feedback message to appear until after their next submission.

2) If I place 1 item correctly, and the others incorrectly, the correct item is also returned to the tray after submission.  Is that intended? I would expect the correct item to remain.

e-kolpakov:
@cahrens @staubina @sstack22 rebased from c958bc5

e-kolpakov:
@sstack22 I was not able to reproduce the second issue (returning correctly placed item with misplaced items) - are you sure you allowed enough time for "drop_item" AJAX call to complete? You can tell it is complete if the spinner is removed.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @cahrens no, it is still shown in feedback area - it is passed through `get_user_state` handler, rather than here.

sstack22:
@e-kolpakov - that must have been it. I tried to reproduce again and was unable to.

e-kolpakov:
@sstack22 I've addressed feedback hiding on virtually anything - updating the sandbox now (should complete in 5-10 minutes)

sstack22:
@e-kolpakov - I just tested again and I do still see the first issue (when reset, no red X feedback message should appear).

e-kolpakov:
@sstack22 That's right, I've mistakingly updated the wrong sandbox so it still runs on old code. I'm updating it now - will likely take same 10-15 minutes, unless I fail miserably again. :)

I've added a check to prevent clicking "Submit" when item is "dropping"

e-kolpakov:
@sstack22 sandbox updated.

Just to make sure, item count feedback is still shown when learner submits an attempt, *does not* reset the problem and refreshes the page - you've requested that in some comment on #87 (don't have time to search for it now - I've got little time before my vacation and some stuff is not done yet).

sstack22:
@e-kolpakov - awesome, confirmed it's fixed! 👍  from me, looks great.

tests/integration/test_base.py
```diff
@@ -122,3 +122,14 @@ def wait_until_has_class(class_name, elem):
         wait = WebDriverWait(elem, 2)
         wait.until(lambda e: class_name in e.get_attribute('class').split(),
                    u"Class name {} not in {}".format(class_name, elem.get_attribute('class')))
+
+    def wait_for_ajax(self, timeout=15):
```
cahrens: Out of curiousity, why did you guys choose not to use bok choy?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -349,15 +351,23 @@ function DragAndDropTemplates(configuration) {
         return h(
             popupSelector, {style: {display: have_messages ? 'block' : 'none'} },
             [
-                h('div.close.icon-remove-sign.fa-times-circle'),
+                h(
+                    'button.unbutton.close.icon-remove-sign.fa-times-circle',
```
cahrens: I am not accustomed to seeing a button done this way. I suggest you follow the pattern here:

http://ux.edx.org/components/buttons/
and

So there would be a button containing 2 spans-- one that has the icon (always aria-hidden), and one with the sr text. Similar to this--

image

tests/integration/test_base.py
e-kolpakov: @cahrens I've just copied it from #95, so it's a question to @bradenmacdonald

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -487,7 +491,7 @@ function DragAndDropBlock(runtime, element, configuration) {
 
             // Set up event handlers:
 
-            $(document).on('keydown mousedown touchstart', closePopup);
+            $element.on('click', "div.item-feedback-popup button.close", closePopup);
```
cahrens: Nit: It's best just to use the class names (instead of including the html elements). You might wish to be consistent with the naming convention of the other button classes (lines 495-497).

cahrens:
Note: I'm getting a "Studio's having trouble saving your work" error on this unit-- don't know why.
http://studio-dndv2-sandbox2.opencraft.hosting/container/block-v1:Opencraft+DNDv2+1+type@vertical+block@b718f781f3de4937bb7802838b33c9db

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -321,13 +321,58 @@ function DragAndDropTemplates(configuration) {
...
+        }
+
+        return h(
+            popupSelector,
+            {
+                style: {display: have_messages ? 'block' : 'none'},
+                attributes: {"tabindex": "-1"}
+            },
+            [
+                h(
```
cahrens: I am not accustomed to seeing a button done this way. I suggest you follow the pattern here:

http://ux.edx.org/components/buttons/
and

So there would be a button containing 2 spans-- one that has the icon (always aria-hidden), and one with the sr text. Similar to this--

image

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -748,6 +752,15 @@ function DragAndDropBlock(runtime, element, configuration) {
         $root.find('.item-bank .option').first().focus();
     };
 
+    var focusItemFeedbackPopup = function() {
+        var popup = $root.find('.item-feedback-popup');
```
cahrens: $root will still be limited to this drag and drop problem instance?

e-kolpakov: @cahrens yes.

tests/integration/test_interaction.py
```diff
@@ -695,6 +699,27 @@ def test_per_item_feedback_multiple_misplaced(self):
         for message_element in expected_message_elements:
             self.assertIn(message_element, self._get_popup_content().get_attribute('innerHTML'))
 
+    def test_submit_disabled_during_drop_item(self):
+        def delayed_drop_item(item_attempt, suffix=''):  # pylint: disable=unused-argument
+            # some delay to allow selenium check submit button disabled status while "drop_item"
+            # XHR is still executing
+            time.sleep(0.1)
```
cahrens: Why can't you wait until the button becomes disabled? Sleeps should be avoided at all costs because they are likely to be flaky, and then slow down tests (vs. only waiting until the state changes).

tests/integration/test_base.py
bradenmacdonald: @cahrens In the strictest technical sense, the selenium tests are actually using bok choy, but they're not using the bok choy PageObject API. I don't honestly know/remember the reason we're not using bok choy page objects, but it's probably simply that it we didn't realize it was an option and consider it fully.

I know that native selenium element selectors can be filtered and refined but you can't do that with bok choy (or at least not historically) - i.e. you can't go `self.q(css="".xblock).q(css=".problem-bank").q(css=".item")` which I believe is a pattern we use in these tests.

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @cahrens Unfortunately for some reason I was not able to find in 15 minutes, removing `fa-times-circle` from the button makes it ignore the clicks. Not removing it obviously results in double icon.

It's 9pm my timezone, so I won't be able to address that, sorry.

tests/integration/test_interaction.py
e-kolpakov: @cahrens I don't want to wait for something to happen - instead, the idea here is to perform checks while XHR is still in progress; without this slowdown, XHR completes too fast, so the test is extremely unreliable and have ~50% success ratio.

cahrens:
@e-kolpakov and @bradenmacdonald who will be picking up this PR and finishing it, or will it wait for @e-kolpakov's return? Note that I am also on vacation starting tomorrow, returning 8/29.

e-kolpakov:
@cahrens I've got to go now - it's 9pm my timezone, and I've got to get up really early tommorow, so a5e222c92f21f292e5246b64121c23682f772a10 was the last batch of changes from me.

CC @bradenmacdonald

tests/integration/test_interaction.py
cahrens: I see-- I misunderstood the point of the method.

e-kolpakov:
@cahrens I see the error on the sandbox too, but I'm not quite sure what triggers it (I've just updated the sandbox to the latest version). Also, looks like the error is stuck - I can update item titles ([note "Goes to the top1"](http://studio-dndv2-sandbox2.opencraft.hosting/container/block-v1:Opencraft+DNDv2+1+type@vertical+block@b718f781f3de4937bb7802838b33c9db) - I've just added that *1*) and it is shown on any unit page, including demo course.

@bradenmacdonald - might make sense to ask the one to pick the PR to rebuild the VM from scratch.

bradenmacdonald:
@cahrens I will see if someone from our team (possibly even myself) can pick this up today/tomorrow, but due to some other fires that have sprung up, it may need to wait until Monday, when I can assign someone else.

Now that @e-kolpakov removed the problematic validation UX, it seems like there are just some very minor points left, right?

cahrens:
@bradenmacdonald the validation UX is a different PR.

bradenmacdonald:
@cahrens Ah, sorry.

haikuginger:
@itsjeyd, I've updated the feedback close pattern to match that pattern; could you review, please?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -321,13 +321,73 @@ function DragAndDropTemplates(configuration) {
...
+            {
+                style: {display: have_messages ? 'block' : 'none'},
+                attributes: {"tabindex": "-1"}
+            },
+            [
+                h(
+                    'button.unbutton.close',
+                    {
+                        attributes: {
+                            'aria-hidden': !have_messages
```
itsjeyd: @haikuginger I know you didn't add this, but I'm wondering why this `button` needs to have an `aria-hidden` attribute: As per L354 above, if `have_messages` is false/falsy, we don't display the popup at all, so specifying that the close button should be hidden seems redundant?

haikuginger:
@itsjeyd, the sandbox has been updated with all relevant changes - can you give it another pass, please?

Thanks!

bradenmacdonald:
@haikuginger Looks like this will need a rebase as well. If you encounter any flaky tests after rebasing, please try to fix them.

itsjeyd:
@haikuginger I made another pass over the PR to double-check whether all comments had been addressed. One comment seems to have slipped through: https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/93#discussion-diff-74738064. Could you please address it when you rebase? It's a tiny change (remember to update the translation files when you address it though).

I will do a final round of manual testing when you're done rebasing and the sandbox is up-to-date.

haikuginger:
Rebased from [`f7131ee`](https://github.com/edx-solutions/xblock-drag-and-drop-v2/commits/f7131ee048c9f4cd7a66d514c1aeb28ca68b51df).

haikuginger:
@itsjeyd, I've rebased, made that tweak, and deployed to the sandbox. Ready for your final review.

@bradenmacdonald, it does look like there are some flaky tests; I'll take a look and see if I can spot anything that needs changing about them.

haikuginger:
@bradenmacdonald, good news is that it doesn't look like flaky tests - just tests that need to be re-written and that may have been ignored due to the regular failure of other tests. I'll get on that right away.

haikuginger:
@itsjeyd, I am happy to report that integration tests are all green, should stay that way, your comments are addressed, and the latest build is deployed on the sandbox for your review.

itsjeyd:
Thanks for the updates, @haikuginger. The code looks good.

I tested the feature on the sandbox using the unit under "PR Review Testing" > "Assessment mode: Per-item feedback", and ran into some issues:

* It's not possible to close the feedback popup; nothing happens when clicking the "x" button in the top right corner (or hitting RET while focused). I had to reload the page to make it disappear.

* In some cases, submitting a solution with incorrectly placed items leads to the "Reset" and "Submit" buttons being disabled. I had to reload the page to be able to interact with the problem again (instead of just being able to click "Reset"). I got this behavior for any solution that did not include any correctly placed items.

* When placing all required items correctly and additionally placing the decoy item on the board:

  * The "introductory feedback" message displayed below the problem ("Drag the items onto the image above.") changes it's icon and color after clicking "Submit":

    introductory-feedback-standard

    vs.

    introductory-feedback-green

  * The feedback popup is missing the default message ("Some of your answers were not correct.") and has a black background instead of a red one:

    feedback-all-required-items-correct-plus-decoy-item

Lastly, I was unable to create new units in Studio (or access existing units for editing), getting "The Studio servers encountered an error". This is most likely not related to the changes introduced here, so maybe just try installing the feature branch on another one of our DnDv2 sandboxes for now and see if the problem persists.

staubina:
@haikuginger @itsjeyd 
Let me know when you would like myself or @cahrens to take another pass at this PR.  @cahrens is still out, but I think her opinion would be good before merging this.

haikuginger:
@itsjeyd: I can replicate all these issues in Firefox; I'll look into the first two, but I believe that the last issue is something that's [being handled in a separate PR](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/96).

haikuginger:
@itsjeyd, actually, I can't replicate the second point. The Submit button does disable after getting all items incorrect, but it re-enables as soon as you drag an item onto the image again. Presumably this is to prevent empty submissions.

haikuginger:
@itsjeyd, I've pushed a fix for the feedback dialog not closing in Firefox (worked in Safari when I manually tested it); it's all down to what the browser decides was the target of the click. This should now handle the case when the target of the click was the button or one of the spans under the button in the DOM. Given I can't replicate the second issue, and the third is in a separate PR, I think this should do it. It's being loaded into the sandbox now.

itsjeyd:
@haikuginger I can confirm that it's now possible to close the feedback popup, either by using the mouse or by tabbing to it and hitting RET -- that's great :)

> The Submit button does disable after getting all items incorrect, but it re-enables as soon as you drag an item onto the image again. *Presumably this is to prevent empty submissions.*

That makes sense, I'm fine with keeping that behavior.

> ... the last issue is something that's being handled in a separate PR.

I checked the sandbox for #96, and the icon and color of the "introductory feedback" message don't change after submitting a solution that places all required items correctly and places the decoy item on some zone. So we've got that part covered :)

But since that PR doesn't include the code for displaying per-item error feedback in assessment mode, I couldn't verify that it would take care of the second issue as well:

> The feedback popup is missing the default message ("Some of your answers were not correct.") and has a black background instead of a red one: ...

Could you please verify that we get the expected behavior by cherry-picking the commits from #96  in your local dev env and report back? Alternatively, you could wait and see if #96 merges today, and test in the sandbox after rebasing this PR again.

haikuginger:
@itsjeyd, I've filed a one-liner commit that resolves that issue in conjunction with #96.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1115,6 +1115,7 @@ function DragAndDropBlock(runtime, element, configuration) {
 
             if (attemptsRemain()) {
                 data.misplaced_items.forEach(function(misplaced_item_id) {
+                    state.last_action_correct = false
```
itsjeyd: @haikuginger Hm, I see why this would work, but it feels kind of hacky. It would be good if `data.correct` were the only "source of truth" regarding the correctness of a solution (in other words, after setting

```js
state.last_action_correct = data.correct;
```

in L1114 above, ideally it shouldn't be necessary to modify the value of `state.last_action_correct` again). I haven't looked at #96 in detail, but it seems like after that PR, `data.correct` should not be truthy if a solution placed one or more decoy items on the board. Is that not the case?

itsjeyd: CC @mtyaka, since he reviewed #96.

haikuginger: I'll do some additional debugging to see if with #96 in place, `data.correct` is giving us what we need.

haikuginger: Okay, looks like #96 did not adjust the logic that sets the `correct` value on the response. It's not strictly in scope for this task, but I can add it if we want - @itsjeyd, @mtyaka, thoughts?

itsjeyd: @haikuginger I think it would be good to take care of it in this context, if it doesn't blow up the scope too much. You might want to confirm with @bradenmacdonald.

bradenmacdonald: We need to get these PRs merged yesterday... so it's your call as to how this affects the scope. If it's a simple, low-risk cleanup, please get it done now. Otherwise, we can postpone it.

haikuginger: Honestly, looking at it, it's pretty much a mess. I'd rather get this delivered and then have some sort of follow-up to clean things up once we actually have a stable codebase.

e-kolpakov:
rebased from bcb06d8

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @haikuginger @itsjeyd @bradenmacdonald I've removed this line and replaced it with a [proper fix](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/93/commits/042c56724ee3db06d4ea9eb0af2425182d79dc3c) - in two words, the reason behind this was that self.item_state was manipulated before value for `correct` was calculated.

So, since this is not the first time I see this problem (expected invocation order violation) in DnDv2 codebase, I've added [two TODOs](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/93/commits/2c78472bd5fbb92ac7ed354ad1513f0c826e4005) to refactor it to the methods that would benefit from such refactoring most. It doesn't look like a good time to try to address them now, so I've just added TODOs.

@bradenmacdonald let me know if you think we should create tickets for those TODOs.

cahrens:
@e-kolpakov Please let us know when this PR is ready for TNL re-review.

e-kolpakov:
@cahrens Sure - I believe @itsjeyd @haikuginger should review and approve it first.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1021,7 +1097,7 @@ function DragAndDropBlock(runtime, element, configuration) {
             focusFirstDraggable();
         });
     };
-
+
```
itsjeyd: @e-kolpakov Nit: Remove trailing whitespace.

itsjeyd:
@e-kolpakov I reviewed your recent updates to the code and verified that the fix for decoy item feedback works:

* If per-item error feedback is available for a given decoy item, the popup that comes up when placing the item on the board and clicking submit has the expected color (red) and contains the default "Some of your answers were not correct" message, as well as per-item feedback).
* If per-item error feedback is *not* available for a given decoy item, no popup is displayed.

Per-item feedback for "normal"/non-decoy items is working as expected as well.

:+1: 

- [x] I tested this: Feature works as advertised in description
- [x] I read through the code
- [x] Includes documentation

cahrens:
Just to clarify, is this now ready for TNL re-review?

e-kolpakov:
@cahrens Yes, I believe it is now ready.

haikuginger:
:+1: 

- [x] I tested this: Quick runthrough of our last few detected bugs shows that they are now working as expected. 
- [x] I read through the code
- [x] Includes documentation

staubina:
👍

e-kolpakov:
@cahrens are we good to merge this PR, or would you like to take another pass?

cahrens:
I'll take a quick look at it later today.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -350,17 +349,29 @@ def do_attempt(self, data, suffix=''):
         self._validate_do_attempt()
 
         self.attempts += 1
-        self._mark_complete_and_publish_grade()  # must happen before _get_feedback
+        # pylint: disable=fixme
+        # TODO: Refactor this method to "freeze" item_state and pass it to methods that need access to it.
```
cahrens: Will this TODO be done in a future story/PR?

cahrens:
👍

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @cahrens I've asked @bradenmacdonald about this and haven't got an answer yet. My current understanding is that it's not a good time to further increase the scope of v2.1 with optional refactroings, so it won't be a part of v2.1.

e-kolpakov:
@sstack22 There were quite a lot of changes to this PR since your +1 - do you want to take another look? That'll be the last review step before merging, so I would appreciate if you could allocate some time for this today. Also, I'm going to cut v2.0.9 today to include it into next week release and it would be great to have this PR included.

(I hope it does not sound too pushy - I don't mean to put any pressure on you)

e-kolpakov:
Squashed commits; previous revision is dc1e3a6

sstack22:
@e-kolpakov - I've re-reviewed, looks go to go 👍

PR #104, opened 2016-09-15
Enforcing XBlock variable types

haikuginger:
We encountered an issue where excessively long numbers submitted through the configuration pane would be saved as strings in Mongo, and would prevent the XBlock from loading correctly in Studio, because in integer form, they were too large to store. This change will cast all entries to the appropriate type, allowing the Mongo database driver to raise appropriate exceptions, rather than save incorrect values to the database.

**JIRA issue**:  [SOL-2088](https://openedx.atlassian.net/browse/SOL-2088)

**Testing Instructions**
1. With the existing version of the xblock, create a block in a course.
2. With the existing version of the xblock, edit the block.
3. With the existing version of the xblock, submit a large number (e.g., 99999999999999999999999) for the `max_attempts` field after setting `mode` to Assessment.
4. Note that the value is accepted, but thereafter, the block cannot be loaded.
5. Delete the block
6. Install this version of the xblock.
7. Run Studio; note that because recent versions of OpenEdX require Drag and Drop V2, you must do `paver devstack studio --fast` to avoid having this version of the XBlock overwritten with the current release version.
8. Again, create a block in a course.
9. Again, edit the block.
10. Again, submit the same large number for the `max_attempts` field after setting`mode` to Assessment.
11. Note that the value is rejected.
12. Note that the block can still be loaded, without the value that was submitted.

**Notes**

What we're looking for is basically for the XBlock API to coerce any values for integer fields to integers. This ensures that the Mongo database driver's type checking can properly validate all numbers, including large numbers, prior to them being inserted into the database.

**Reviewers**
- [ ] @pomegranited 
- [ ] edX reviewer[s] TBD

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -63,6 +63,7 @@ class DragAndDropBlock(
         help=_("The title of the drag and drop problem. The title is displayed to learners."),
         scope=Scope.settings,
         default=_("Drag and Drop"),
+        enforce_type=True,
     )
 
     mode = String(
@@ -78,7 +79,8 @@ class DragAndDropBlock(
             {"display_name": _("Standard"), "value": Constants.STANDARD_MODE},
             {"display_name": _("Assessment"), "value": Constants.ASSESSMENT_MODE},
         ],
-        default=Constants.STANDARD_MODE
+        default=Constants.STANDARD_MODE,
+        enforce_type=True,
     )
 
     max_attempts = Integer(
@@ -89,6 +91,7 @@ class DragAndDropBlock(
         ),
         scope=Scope.settings,
         default=None,
+        enforce_type=True,
     )
 
     show_title = Boolean(
@@ -96,6 +99,7 @@ class DragAndDropBlock(
         help=_("Display the title to the learner?"),
         scope=Scope.settings,
         default=True,
+        enforce_type=True,
     )
 
     question_text = String(
@@ -103,6 +107,7 @@ class DragAndDropBlock(
         help=_("The description of the problem or instructions shown to the learner."),
         scope=Scope.settings,
         default="",
+        enforce_type=True,
     )
 
     show_question_header = Boolean(
@@ -110,6 +115,7 @@ class DragAndDropBlock(
         help=_('Display the heading "Problem" above the problem text?'),
         scope=Scope.settings,
         default=True,
+        enforce_type=True,
     )
 
     weight = Float(
@@ -117,6 +123,7 @@ class DragAndDropBlock(
         help=_("Defines the number of points the problem is worth."),
         scope=Scope.settings,
         default=1,
+        enforce_type=True,
     )
 
     item_background_color = String(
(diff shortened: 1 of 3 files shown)


edx-webhook:
Thanks for the pull request, @haikuginger! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=104

bradenmacdonald:
@haikuginger You have a pylint error:
```
W:230,8:BasicTests.test_studio_submit_coerce_to_integer: Unused variable 'res'
```

cahrens:
👍 Sorry, I didn't notice the last time I was pinged on the Solutions ticket.

It does make me wonder when you would want enforce_type to be False....

bradenmacdonald:
Thanks @cahrens !

bradenmacdonald:
@haikuginger Can you please rebase and merge?

haikuginger:
Rebased and squashed from 1e2f5e7.

haikuginger:
Okay, so apparently this change now pushes the `drag_and_drop_v2.py` module over Pylint's threshold of acceptable module length of 1000 lines. A lot of that code is actually docstring, but that's neither here nor there.

Now, there are three options, as I see it. I can take an excessive amount of time to refactor the app to reduce the module length. I can take a less excessive amount of time to find 12 lines that can be trimmed via techniques like using ternary `x if y else z` operations instead of multiline `if x: return y; else: return z` statements. Or, I can set the linter to be okay with the fact that the module is a little bit of a whale.

@cahrens, @bradenmacdonald, thoughts?

cahrens:
@cpennington do you have an opinion about this particular Pylint rule? I don't know if this is something you are relaxing for edx-platform.

cpennington:
@cahrens: I'm pretty sure I just disabled that in my edx-lint PR. One option would be to just install edx-lint and use the new standard pylintrc from there. (https://github.com/edx/edx-lint)

bradenmacdonald:
@haikuginger Yeah, Cale's suggestion sounds good to me.

haikuginger:
@cpennington, it looks like the canonical version of edx-lint still has `max-module-lines` set to 1000, so I've set up the pylintrc using that, but set a tweak on that setting to have it at 1500. We can modify that later if needed.

haikuginger:
@cpennington @cahrens @bradenmacdonald, unfortunately, it looks like `edx-lint` isn't compatible with `xblock-sdk`; the prior requires `pylint>=1.0` by inheritance while the latter has a fixed requirement at `pylint==0.28.0`. For right now, if it's all right, I'll revert the change and simply set `max-module-lines` directly in pylintrc.

haikuginger:
@cpennington @cahrens @bradenmacdonald, any objection to the above?

cahrens:
That's fine with me.

haikuginger:
Switched back to the previous, non-edx-lint code, with a change to pylintrc expanding the max module length to 1500 lines. Previous commit: 364c527.

PR #96, opened 2016-08-18
[SOL-1988] Account for decoy items in score

arbrandes:
## Description

In order to take the correct non-placement of decoy items into account, this PR implements the student score as `score = ((correctly placed items) + (correctly unplaced decoys)) / (total items)`

## Dependencies

- #87 (now merged)

## JIRA

- [SOL-1988](https://openedx.atlassian.net/browse/SOL-1988)

## Sandbox URLs

- [LMS](http://dndv2-pr96.opencraft.hosting/courses/course-v1:test+tt101+run/courseware/chapter/drag_and_drop_v2/)
- [Studio](http://studio-dndv2-pr96.opencraft.hosting/container/block-v1:test+tt101+run+type@vertical+block@drag_and_drop_v2)

## Latest commit

- 88f0873 *(updated 2016-08-23)*

## Testing instructions

1. Go to the [pre-deployed DnDv2 instance](http://dndv2-pr96.opencraft.hosting/courses/course-v1:test+tt101+run/courseware/chapter/drag_and_drop_v2/) on the LMS, and log in as `staff@example.com`.
2. Click on `Staff Debug Info`, and `Delete Student State` for the `staff@example.com` user.
3. Close the debug info window, drag all items into their correct places, and finally drag the `I don't belong anywhere` item anywhere onto the image.
4. Click `Submit`.
5. Go to the [progress page](http://dndv2-pr96.opencraft.hosting/courses/course-v1:test+tt101+run/progress), and check that the score for `Drag and Drop v2` is **80%**, or 4 out of 5.

### Suggested submission

screenshot

### Expected grade

screenshot2

### Reviewers

- [x] @mtyaka
- [x] @staubina
- [x] @sstack22

PR #108, opened 2016-10-19
Display grade information

mtyaka:
## Description

Student's grade was not displayed on DnDv2 problems. This patch adds code to display score information beneath the display name of the problem (similar to CAPA).

It also implements the missing `max_score` method which is required by the LMS to correctly calculate and display grades on the "Progress" tab.

screen shot 2016-10-20 at 10 10 22
## JIRA
- [SOL-2030](https://openedx.atlassian.net/browse/SOL-2030)
- [SOL-2094](https://openedx.atlassian.net/browse/SOL-2094)
## Sandbox URLs

Latest commit on the sandboxe: 3ad5a184dfe6ecb6644603cc75973e19738a97da
- [DnD unit in LMS](http://pr13797.sandbox.opencraft.hosting/courses/course-v1:DnDX+DnD101+2016/courseware/ce77c57034524ffb938c2b16e8316b7d/d49d04ffc18f4e738229aab063234adf/1?activate_block_id=block-v1%3ADnDX%2BDnD101%2B2016%2Btype%40vertical%2Bblock%40631785c086ab45278755c34fa7562a31)
- [Progress page in LMS](http://pr13797.sandbox.opencraft.hosting/courses/course-v1:DnDX+DnD101+2016/progress)
- [DnD unit in Studio](http://studio-pr13797.sandbox.opencraft.hosting/container/block-v1:DnDX+DnD101+2016+type@vertical+block@631785c086ab45278755c34fa7562a31)
## Testing instructions
1. Add a new DnDv2 block using default settings (standard mode).
2. Open the problem in LMS. The grey line below the title should say: "1 point possible (graded)" or "1 point possible (ungraded)", depending on whether you created the problem in a graded or ungraded section.
3. Toggle section between graded and ungraded (in Course Outline view in Studio, click the settings icon of the subsection containing the DnDv2 problem and select "Not Graded" or any of the other options if you want the problem to be graded).
4. Refresh the page in LMS. Make sure the text changes between "(graded)" <-> "(ungraded)".
5. Open the "Progress" page. Verify that the section corresponding to the DnD problem says `0/1`.
6. Go back to the DnD problem and start dragging items to the board. As you drop each item, check that the grade updates accordingly. Also check that the grade on the "Progress" page is always in sync with the grade displayed on the DnD problem.
7. Reset the problem. Verify that resetting the problem does not reset the grade (both on the problem itself and on the "Progress" page).
8. Change max score (weight) setting in Studio. Delete student state to clear the existing grade, then check that the grades are scaled correctly for max_score.
9. Repeat the above steps with a DnD problem in "assessment" mode. In assessment mode, the grade should only update when you explicitly submit the problem, not on every drop like it does in standard mode.
## Notes and Concerns
#96 makes decoy items contribute to the final score. This has the surprising effect that the first move you do on a fresh DnD problem that includes at least one decoy item is wrong, you immediately get awarded points for the decoy item.
## Reviewers
- [x] OpenCraft: @itsjeyd
- [ ] a11y: @edx/edx-accessibility
- [x] Product: @sstack22

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -107,8 +107,8 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     )
 
     weight = Float(
-        display_name=_("Maximum score"),
-        help=_("The maximum score the learner can receive for the problem."),
+        display_name=_("Problem Weight"),
+        help=_("Defines the number of points the problem is worth."),
         scope=Scope.settings,
         default=1,
     )
@@ -172,6 +172,23 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     block_settings_key = 'drag-and-drop-v2'
     has_score = True
 
+    def max_score(self):  # pylint: disable=no-self-use
+        """
+        Return the problem's max score, which for DnDv2 always equals 1.
+        Required by the grading system in the LMS.
+        """
+        return 1
+
+    def _learner_raw_score(self):
+        """
+        Calculate raw score for learner submission.
+
+        As it is calculated as ratio of correctly placed (or left in bank in case of decoys) items to
+        total number of items, it lays in interval [0..1]
+        """
+        correct_count, total_count = self._get_item_stats()
+        return correct_count / float(total_count)
+
     @XBlock.supports("multi_device")  # Enable this block for use in the mobile app via webview
     def student_view(self, context):
         """
@@ -232,6 +249,8 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
             "mode": self.mode,
             "zones": self.zones,
             "max_attempts": self.max_attempts,
+            "graded": getattr(self, 'graded', False),
+            "weighted_max_score": self.max_score() * self.weight,
             "max_items_per_zone": self.max_items_per_zone,
             # SDK doesn't supply url_name.
             "url_name": getattr(self, 'url_name', ''),
@@ -413,6 +432,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         return {
             'correct': correct,
             'attempts': self.attempts,
+            'grade': self._get_grade_if_set(),
             'misplaced_items': list(misplaced_ids),
             'feedback': self._present_feedback(feedback_msgs),
             'overall_feedback': self._present_feedback(overall_feedback_msgs)
@@ -571,9 +591,14 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
         feedback_msgs.append(FeedbackMessage(problem_feedback_message, problem_feedback_class))
 
-        if not self.attempts_remain:
+        if self.weight > 0:
+            if self.attempts_remain:
+                grade_feedback_template = FeedbackMessages.GRADE_FEEDBACK_TPL
+            else:
+                grade_feedback_template = FeedbackMessages.FINAL_ATTEMPT_TPL
+
             feedback_msgs.append(
-                FeedbackMessage(FeedbackMessages.FINAL_ATTEMPT_TPL.format(score=self.grade), grade_feedback_class)
+                FeedbackMessage(grade_feedback_template.format(score=self.grade), grade_feedback_class)
             )
 
         return feedback_msgs, misplaced_ids
@@ -608,6 +633,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
         return {
             'correct': is_correct,
+            'grade': self._get_grade_if_set(),
             'finished': self._is_answer_correct(),
             'overall_feedback': self._present_feedback(overall_feedback),
             'feedback': self._present_feedback([item_feedback])
@@ -654,7 +680,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
         """
         # pylint: disable=fixme
         # TODO: (arguable) split this method into "clean" functions (with no side effects and implicit state)
-        # This method implicitly depends on self.item_state (via _is_answer_correct and _get_grade)
+        # This method implicitly depends on self.item_state (via _is_answer_correct and _calculate_grade)
         # and also updates self.grade if some conditions are met. As a result this method implies some order of
         # invocation:
         # * it should be called after learner-caused updates to self.item_state is applied
@@ -666,9 +692,10 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
         # There's no going back from "completed" status to "incomplete"
         self.completed = self.completed or self._is_answer_correct() or not self.attempts_remain
-        grade = self._get_grade()
+        grade = self._calculate_grade()
         # ... and from higher grade to lower
-        if grade > self.grade:
+        current_grade = self._get_grade_if_set()
+        if current_grade is None or grade > current_grade:
             self.grade = grade
             self._publish_grade()
 
@@ -751,6 +778,7 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
             'items': item_state,
             'finished': is_finished,
             'attempts': self.attempts,
+            'grade': self._get_grade_if_set(),
             'overall_feedback': self._present_feedback(overall_feedback_msgs)
         }
 
@@ -872,12 +900,21 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
 
         return ItemStats(required, placed, correctly_placed, decoy, decoy_in_bank)
 
-    def _get_grade(self):
+    def _calculate_grade(self):
         """
-        Returns the student's grade for this block.
+        Calculates the student's grade for this block based on current item state.
         """
-        correct_count, total_count = self._get_item_stats()
-        return correct_count / float(total_count) * self.weight
+        return self._learner_raw_score() * self.weight
+
+    def _get_grade_if_set(self):
+        """
+        Returns student's grade if already explicitly set, otherwise returns None.
+        This is different from self.grade which returns 0 by default.
+        """
+        if self.fields['grade'].is_set_on(self):
+            return self.grade
+        else:
+            return None
 
     def _answer_correctness(self):
         """
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -1,8 +1,6 @@
 function DragAndDropTemplates(configuration) {
     "use strict";
     var h = virtualDom.h;
-    // Set up a mock for gettext if it isn't available in the client runtime:
-    if (!window.gettext) { window.gettext = function gettext_stub(string) { return string; }; }
 
     var itemSpinnerTemplate = function(item) {
         if (!item.xhr_active) {
@@ -184,7 +182,7 @@ function DragAndDropTemplates(configuration) {
         var item_wrapper = 'div.item-wrapper.item-align.item-align-' + zone.align;
         var is_item_in_zone = function(i) { return i.is_placed && (i.zone === zone.uid); };
         var items_in_zone = $.grep(ctx.items, is_item_in_zone);
-        var zone_description_id = 'zone-' + zone.uid + '-description';
+        var zone_description_id = configuration.url_name + '-zone-' + zone.uid + '-description';
         if (items_in_zone.length == 0) {
           var zone_description = h(
             'div',
@@ -443,8 +441,63 @@ function DragAndDropTemplates(configuration) {
         )
     };
 
+    var progressTemplate = function(ctx) {
+        // Formats a number to 4 decimals without trailing zeros
+        // (1.00 -> '1'; 1.50 -> '1.5'; 1.333333333 -> '1.3333').
+        var formatNumber = function(n) { return n.toFixed(4).replace(/\.?0+$/, ''); };
+        var is_graded = ctx.graded && ctx.weighted_max_score > 0;
+        var progress_template;
+        if (ctx.grade !== null && ctx.weighted_max_score > 0) {
+            if (is_graded) {
+                progress_template = ngettext(
+                      // Translators: {earned} is the number of points earned. {possible} is the total number of points (examples: 0/1, 1/1, 2/3, 5/10). The total number of points will always be at least 1. We pluralize based on the total number of points (example: 0/1 point; 1/2 points).
+                    '{earned}/{possible} point (graded)',
+                    '{earned}/{possible} points (graded)',
+                    ctx.weighted_max_score
+                );
+            } else {
+                progress_template = ngettext(
+                      // Translators: {earned} is the number of points earned. {possible} is the total number of points (examples: 0/1, 1/1, 2/3, 5/10). The total number of points will always be at least 1. We pluralize based on the total number of points (example: 0/1 point; 1/2 points).
+                    '{earned}/{possible} point (ungraded)',
+                    '{earned}/{possible} points (ungraded)',
+                    ctx.weighted_max_score
+                );
+            }
+            progress_template = progress_template.replace('{earned}', formatNumber(ctx.grade));
+        } else {
+            if (is_graded) {
+                progress_template = ngettext(
+                    // Translators: {possible} is the number of points possible (examples: 1, 3, 10).
+                    '{possible} point possible (graded)',
+                    '{possible} points possible (graded)',
+                    ctx.weighted_max_score
+                );
+            } else {
+                progress_template = ngettext(
+                    // Translators: {possible} is the number of points possible (examples: 1, 3, 10).
+                    '{possible} point possible (ungraded)',
+                    '{possible} points possible (ungraded)',
+                    ctx.weighted_max_score
+                );
+            }
+        }
+        var progress_text = progress_template.replace('{possible}', formatNumber(ctx.weighted_max_score));
+
+        return h('div.problem-progress', {
+            id: configuration.url_name + '-problem-progress',
+            attributes: {'role': 'status', 'aria-live': 'polite'}
+        }, progress_text);
+    };
+
     var mainTemplate = function(ctx) {
-        var problemTitle = ctx.show_title ? h('h3.problem-title', {innerHTML: ctx.title_html}) : null;
+        var problemProgress = progressTemplate(ctx);
+        var problemTitle = null;
+        if (ctx.show_title) {
+            problemTitle = h('h3.problem-title', {
+                innerHTML: ctx.title_html,
+                attributes: {'aria-describedby': problemProgress.properties.id}
+            });
+        }
         var problemHeader = ctx.show_problem_header ? h('h4.title1', gettext('Problem')) : null;
         // Render only items in the bank here, including placeholders.  Placed
         // items will be rendered by zoneTemplate.
@@ -463,6 +516,7 @@ function DragAndDropTemplates(configuration) {
         return (
             h('section.themed-xblock.xblock--drag-and-drop', [
                 problemTitle,
+                problemProgress,
                 h('section.problem', [
                     problemHeader,
                     h('p', {innerHTML: ctx.problem_html}),
@@ -500,6 +554,12 @@ function DragAndDropTemplates(configuration) {
 function DragAndDropBlock(runtime, element, configuration) {
     "use strict";
 
+    // Set up a mock for gettext if it isn't available in the client runtime:
+    if (!window.gettext) {
+        window.gettext = function gettext_stub(string) { return string; };
+        window.ngettext = function ngettext_stub(strA, strB, n) { return n == 1 ? strA : strB; };
+    }
+
     DragAndDropBlock.STANDARD_MODE = 'standard';
     DragAndDropBlock.ASSESSMENT_MODE = 'assessment';
 
@@ -509,9 +569,6 @@ function DragAndDropBlock(runtime, element, configuration) {
     };
 
     var renderView = DragAndDropTemplates(configuration);
-
-    // Set up a mock for gettext if it isn't available in the client runtime:
-    if (!window.gettext) { window.gettext = function gettext_stub(string) { return string; }; }
 
     var $element = $(element);
     element = $element[0]; // TODO: This line can be removed when we no longer support Dogwood.
@@ -1057,6 +1114,7 @@ function DragAndDropBlock(runtime, element, configuration) {
                 if (configuration.mode === DragAndDropBlock.STANDARD_MODE) {
                     state.last_action_correct = data.correct;
                     state.feedback = data.feedback;
+                    state.grade = data.grade;
                     if (!data.correct) {
                         delete state.items[item_id];
                     }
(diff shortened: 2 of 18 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=108&repo=edx-solutions%2Fxblock-drag-and-drop-v2

tests/unit/test_advanced.py
```diff
@@ -123,51 +132,69 @@ def test_drop_item_correct(self):
...
                 published_grades.append(params)
         self.block.runtime.publish = mock_publish
 
+        self.assertEqual(0, self.block.grade)
+
         self.call_handler(self.DROP_ITEM_HANDLER, {"val": 0, "zone": self.ZONE_1})
 
         self.assertEqual(1, len(published_grades))
-        self.assertEqual({'value': 0.75, 'max_value': 1}, published_grades[-1])
+        self.assertEqual(0.75 * weight, self.block.grade)
```
mtyaka: Done.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -444,8 +442,51 @@ function DragAndDropTemplates(configuration) {
...
+                    ctx.max_score
+                );
+            } else {
+                progress_template = ngettext(
+                    '{possible} point possible (ungraded)',
+                    '{possible} points possible (ungraded)',
+                    ctx.max_score
+                );
+            }
+        }
```
mtyaka: You're right, that function was getting way too long. I moved the progress template to a separate function.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -463,6 +504,7 @@ function DragAndDropTemplates(configuration) {
         return (
             h('section.themed-xblock.xblock--drag-and-drop', [
                 problemTitle,
+                h('div.problem-progress', {key: progress_id, id: progress_id}, progress_text),
```
mtyaka: @itsjeyd The `key` entry does not translate to HTML, it's an internal property used by virtual-dom so that it can distinguish elements when reordering them.
For example if you first render two elements `[h('div', 'a'), h('div', 'b')]`, and then reorder them when rendering the second time (`[h('div', 'b'), h('div', 'a')]`), virtual-dom will not be able to know that the elements have been reordered so it will patch the two corresponding DOM elements rather than reordering them. It usually doesn't matter since the visual end result is the same, but if the two DOM elements contain any internal state (for example if you modify then with jQuery outside of virtual-dom), it can lead to weird issues.

There is no reordering going on here, so the `key` property isn't really needed. I added it while trying to fix weird selenium failures just in case it helped (it didn't; the issues were not caused by virtual-dom). I will remove the `key` property.

mtyaka:
@itsjeyd 

> One question about a11y: Using Orca and Firefox I was able to get the screen reader to announce the contents of the problem-progress element to me (as part of reading the problem-title). But updates to the grade were not announced to me as I was placing items on the board in standard mode, or after clicking "Submit" in assessment mode. Do CAPA problems do this? If so, we should probably try and replicate that behavior for DnDv2.1.

Good idea, thank you. CAPA problems display the grade in the feedback area when you submit the problem and the screen reader reads that.

DnDv2 doesn't display the grade in the feedback area, and in standard mode the grade updates dynamically while you're working on the problem, so it does make sense for the screen reader to announce updates to the score. I added `role="status"` and `aria-live="polite"` to the progress element and verified that VoiceOver announced the changes as I work on a DnDv2 problem.

tests/integration/test_interaction.py
```diff
@@ -271,6 +272,29 @@ def test_final_feedback_and_reset(self, action_key):
     def test_keyboard_help(self, use_keyboard):
         self.interact_with_keyboard_help(use_keyboard=use_keyboard)
 
+    def test_grade_display(self):
+        items_with_zones = self._get_items_with_zone(self.items_map).values()
+        items_without_zones = self._get_items_without_zone(self.items_map).values()
+        total_items = len(items_with_zones) + len(items_without_zones)
+
+        progress = self._page.find_element_by_css_selector('.problem-progress')
+        self.assertEqual(progress.text, '1 point possible (ungraded)')
```
itsjeyd: @mtyaka Thanks for the explanation, I'm OK with leaving the tests as is.

drag_and_drop_v2/public/js/drag_and_drop.js
itsjeyd: @mtyaka Sounds good, thanks!

itsjeyd:
Thanks for the updates, @mtyaka! I had another look at the code and have no more comments. Let me know when the sandbox is up to date so I can do another quick round of testing.

mtyaka:
@itsjeyd The sandbox should be up to date :)

itsjeyd:
@mtyaka I gave this another quick test on the sandbox, and it's working fine. But my screen reader is acting up right now, so I couldn't verify that changes to the grade are getting announced when adding items to the board in standard mode or when submitting a solution in assessment mode.

@haikuginger Since you're sprint firefighter, could you help out and verify just that aspect on the sandbox please? See https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/108#issuecomment-255314435 for reference. I'll approve this PR as soon as that's done.

mtyaka:
@haikuginger In assessment mode only negative feedback popups are shown. Positive feedback is still displayed in the feedback message area, but there is no popup. I think it's the popup that somehow prevents the update to the score from being announced.

@itsjeyd I added code to display grade information in the feedback area when in assessment mode. In assessment mode we already displayed the highest grade when user runs out of attempts, but after this change, the grade is displayed after every submission.

The sandbox has been updated, please take a look.

mtyaka:
@cahrens or @staubina, @edx/edx-accessibility: This is ready for your review.

cahrens:
@mtyaka I created https://openedx.atlassian.net/browse/TNL-5822 for the review of this PR. I put it into our current sprint, but I can't guarantee super-fast turnaround on this one, as I wasn't aware that any more DnD PR reviews were coming.

Should we expect more reviews in the next 2 weeks?

mtyaka:
@cahrens Thank you for creating the ticket! I don't think we plan to do any more work on DnD in the next two weeks, but I'll ask @bradenmacdonald to confirm.

staubina:
@mtyaka 

The link labeled "DnD unit in LMS" is no longer working.  I will follow the steps above to test these changes out.

staubina:
I was able to get to the problem by updating the dates the course was run.  I am guessing this was changed for testing.

staubina:
Is it expected that I get 0.2 points when I answer the problem incorrectly on my first attempt?

screen shot 2016-10-25 at 11 39 25 am

staubina:
If I perform the following steps I would expect my score to drop.  Currently it stays at its highest and never goes down.

1. Drop an answer in the correct location
2. Submit (review score)
3. Move that answer to an incorrect location
4. Submit (review score)

Expect: 
Score to drop when I move my answer to the incorrect location
Actual: 
Score stays the same.

What is the expected behavior here from the AC?

tests/integration/test_interaction_assessment.py
```diff
@@ -220,15 +221,22 @@ def test_do_attempt_feedback_is_updated(self):
...
             FeedbackMessages.correctly_placed(1),
             FeedbackMessages.not_placed(3),
-            START_FEEDBACK
+            START_FEEDBACK,
+            FeedbackMessages.GRADE_FEEDBACK_TPL.format(score=expected_grade)
         ]
         expected_feedback = "\n".join(feedback_lines)
         self.assertEqual(self._get_feedback().text, expected_feedback)
 
+        # Place the item into incorrect zone. The score does not change.
```
staubina: Is this comment true?  I am seeing different behavior as I mentioned in a previous comment.

bradenmacdonald:
@cahrens 
> Should we expect more reviews in the next 2 weeks?

I'm still getting back to speed on everything but AFAIK, no.

Long answer: The only remaining items are not going to come up anytime soon:
* Comprehensive a11y review + follow up (SOL-1978): this hasn't been scheduled yet, so isn't part of our current sprint.
* https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/104 which is a very low priority fix and not part of any upcoming sprints (SOL-2088)
* any handover coordination requested to support TNL with maintenance of this block (if that's still the plan) (SOL-2095)

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -172,6 +172,12 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     block_settings_key = 'drag-and-drop-v2'
     has_score = True
 
+    def max_score(self):
+        """
+        Return the problem's max score. Required by the grading system in the LMS.
+        """
+        return self.weight
```
nasthagiri: Please note that `max_score` is different from `weight`.  The weight is already applied when calculating the problem's score.  So there's no need to specify it in the max_score value as well.  In fact, this may result in unexpected value.  See the formula at https://openedx.atlassian.net/wiki/display/TNL/Robust+Grades+Design#RobustGradesDesign-ProblemScores.

nasthagiri: In your case, max_score may always be `1`.

sstack22:
@mtyaka - this looks great! Just one note - when a problem weight = 0, we also note it as ungraded (even if it is within a graded subsection).

drag_and_drop_v2/drag_and_drop_v2.py
mtyaka: Thanks you @nasthagiri, I missed that! I will change `max_score` to always return `1`.

mtyaka:
Thanks @sstack22! I will update the code so that weight = 0 displays as "ungraded" as well.

mtyaka:
@staubina Thanks a lot for the quick review!

> Is it expected that I get 0.2 points when I answer the problem incorrectly on my first attempt?

The reason is that the demo problem includes one decoy item which counts towards the final score (see "Notes and Concerns" section in PR description). Even if you first move is incorrect, the decoy item is in the correct position (the item bank), which is why you get awarded 0.2 points. The behavior is surprising, but I can't think of an easy ways to work around that.

> If I perform the following steps I would expect my score to drop. Currently it stays at its highest and never goes down. What is the expected behavior here from the AC?

The expected behavior is to display the highest achieved score (see https://openedx.atlassian.net/browse/SOL-2030?focusedCommentId=211959&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-211959). This is different from CAPA problems.

mtyaka:
I changed `max_score` to always return `1`,  removed grade-related feedback in assessment mode when `weight == 0`, and rebuilt the sandbox with the latest code.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -443,8 +441,59 @@ function DragAndDropTemplates(configuration) {
...
+    var progressTemplate = function(ctx) {
+        // Formats a number to 4 decimals without trailing zeros
+        // (1.00 -> '1'; 1.50 -> '1.5'; 1.333333333 -> '1.3333').
+        var formatNumber = function(n) { return n.toFixed(4).replace(/\.?0+$/, ''); };
+        var is_graded = ctx.graded && ctx.max_score > 0;
+        var progress_template;
+        if (ctx.grade !== null) {
+            if (is_graded) {
+                progress_template = ngettext(
+                    '{earned}/{possible} point (graded)',
```
staubina: Add Translation comments here, as shown here: http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/internationalization/i18n.html#python-source-code

staubina:
@sstack22 @mtyaka We made a recent update to Capa to show a score of zero with a weight of 1 as 0/1 point.  This does not seem to be the case here.  Should we update this PR to match that case?

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: @staubina Translations added, thanks for the suggestion.

mtyaka:
@staubina 

> We made a recent update to Capa to show a score of zero with a weight of 1 as 0/1 point. This does not seem to be the case here. Should we update this PR to match that case?

I think it makes sense to display the zero score as 0/weight. The latest commit does that.
I also updated the sandbox with the latest code.

staubina:
@mtyaka Thank you for making these changes.  This PR Looks good to me. :+1:

mtyaka:
Thanks @staubina!

@sstack22 Do you want to take another look?

staubina:
@mtyaka, @nasthagiri Has more experience with the grader and would to take a look.  

@nasthagiri this is ready for your review.

tests/integration/test_interaction.py
```diff
@@ -271,6 +272,29 @@ def test_final_feedback_and_reset(self, action_key):
     def test_keyboard_help(self, use_keyboard):
         self.interact_with_keyboard_help(use_keyboard=use_keyboard)
 
+    def test_grade_display(self):
```
nasthagiri: Is there also an acceptance test that makes sure the grade for DND problems are now displayed on the LMS' progress page?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -443,8 +441,63 @@ function DragAndDropTemplates(configuration) {
...
+            } else {
+                progress_template = ngettext(
+                      // Translators: {earned} is the number of points earned. {possible} is the total number of points (examples: 0/1, 1/1, 2/3, 5/10). The total number of points will always be at least 1. We pluralize based on the total number of points (example: 0/1 point; 1/2 points).
+                    '{earned}/{possible} point (ungraded)',
+                    '{earned}/{possible} points (ungraded)',
+                    ctx.max_score
+                );
+            }
+            progress_template = progress_template.replace('{earned}', formatNumber(ctx.grade));
+        } else {
```
nasthagiri: Why is this `else` clause needed?  It doesn't look like there are any valid cases for it.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -172,6 +172,13 @@ class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
     block_settings_key = 'drag-and-drop-v2'
     has_score = True
 
+    def max_score(self):  # pylint: disable=no-self-use
+        """
+        Return the problem's max score, which for DnDv2 always equals 1.
```
nasthagiri: Note: Currently, you are defining the `max_score` as 1.  However, in the future, you may consider having the `max_score` be equal to the number of items that need to be dragged.  This way, when DND is used as an assessment, learners can get partial credit for some of the items they got correct.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -666,9 +682,10 @@ def _mark_complete_and_publish_grade(self):
 
         # There's no going back from "completed" status to "incomplete"
         self.completed = self.completed or self._is_answer_correct() or not self.attempts_remain
-        grade = self._get_grade()
+        grade = self._calculate_grade()
         # ... and from higher grade to lower
-        if grade > self.grade:
+        current_grade = self._get_grade_if_set()
+        if current_grade is None or grade > current_grade:
```
nasthagiri: Hmmm.. Does this mean if the learner changes their answer so that it's different from the correct answer, they'll see their new incorrect answer, but their grade remains the same?  Won't that be confusing to the learner?  It's also inconsistent with CAPA problems.  FYI @sstack22.
(The sandbox is not responding currently so I can't verify this.)

tests/integration/test_interaction.py
haikuginger: @nasthagiri - taking over for @mtyaka on this temporarily.

We test for the grade event in tests/integration/test_events.py.

drag_and_drop_v2/public/js/drag_and_drop.js
haikuginger: This clause renders a different message in cases where the user hasn't yet submitted an answer.

drag_and_drop_v2/drag_and_drop_v2.py
haikuginger: @nasthagiri, let me look into this one. I know we emit a different `max_value` [in our grade events](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/3ad5a184dfe6ecb6644603cc75973e19738a97da/drag_and_drop_v2/drag_and_drop_v2.py#L697), so I'm not sure why we're returning a static value for the `max_score` method. There may be some legacy considerations here.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -443,8 +441,63 @@ function DragAndDropTemplates(configuration) {
         )
     };
 
+    var progressTemplate = function(ctx) {
+        // Formats a number to 4 decimals without trailing zeros
+        // (1.00 -> '1'; 1.50 -> '1.5'; 1.333333333 -> '1.3333').
+        var formatNumber = function(n) { return n.toFixed(4).replace(/\.?0+$/, ''); };
+        var is_graded = ctx.graded && ctx.max_score > 0;
```
nasthagiri: Does `ctx.max_score` include the `weight` of the problem?  It should - to allow course teams to mark problems as not part of the grade.

nasthagiri: Suggestion: Consider having more descriptive variable names in this code so the reader can easily distinguish between `weighted_max_score` versus `raw_max_score`, for example.
(We did a similar renaming exercise recently in the grading infrastructure in the platform for readability and to prevent bugs.)

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @nasthagiri I believe the same have already [been discussed](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/108#discussion_r85107013). DnDv2 calculates student grade as a [ratio of correct item placements to total count of items](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/3ad5a184dfe6ecb6644603cc75973e19738a97da/drag_and_drop_v2/drag_and_drop_v2.py#L898), so max_score is one, but `grade = student_score * weight`.

We might consider making max_score equal to total number of items, and adjust grade calculations accordingly, but I'm concerned that this might be not backward compatible.

I'm not sure about this part:

> learners can get partial credit for some of the items they got correct.

If I'm not mistaking learners already receive partial credit if they haven't achieved 100% score.

Also, maybe there's a typo in [the formula](https://openedx.atlassian.net/wiki/display/TNL/Robust+Grades+Design#RobustGradesDesign-ProblemScores), but `weight` is in nominator and denominator, so it should cancel out, and the formula should become `learner_raw_score / max_score`.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @nasthagiri @sstack22 keeping the highest score is by design:
* [if you have implemented already, no worries at all"](https://openedx.atlassian.net/browse/SOL-1944?focusedCommentId=205655&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-205655))
* and than [latest version implemented just keeps track of highest achieved score and only submits if the new score was higher](https://openedx.atlassian.net/browse/SOL-1944?focusedCommentId=205666&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-205666)

Than the discussion around this stopped. We might want to revisit it, but I would suggest changing that in a separate PR if necessary.

e-kolpakov:
@nasthagiri I've addressed/answered remaining questions - could you please take another look?

drag_and_drop_v2/drag_and_drop_v2.py
sstack22: Just to follow up on this - from a product perspective, I'm OK with keeping the max score as it was designed. I think this benefits learners. We haven't heard feedback from partners/learners thus far, but we can certainly reconsider in the future. Agreed that if this changes in the future, it would be a separate story.

sstack22:
👍  on my end, given the issue I noted was resolved.

tests/integration/test_interaction.py
nasthagiri: sounds good.  thanks.

drag_and_drop_v2/drag_and_drop_v2.py
nasthagiri: The formula intentionally puts the `weight` in both the numerator and the denominator.  The reason is because the `weighted_possible` (denominator) needs to be the total possible points, including the `weight`.  And therefore, the `weighted_earned` (numerator) is also multiplied by the `weight` in order to keep the total score the same.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -443,8 +441,63 @@ function DragAndDropTemplates(configuration) {
...
+            } else {
+                progress_template = ngettext(
+                      // Translators: {earned} is the number of points earned. {possible} is the total number of points (examples: 0/1, 1/1, 2/3, 5/10). The total number of points will always be at least 1. We pluralize based on the total number of points (example: 0/1 point; 1/2 points).
+                    '{earned}/{possible} point (ungraded)',
+                    '{earned}/{possible} points (ungraded)',
+                    ctx.weighted_max_score
+                );
+            }
+            progress_template = progress_template.replace('{earned}', formatNumber(ctx.grade));
+        } else {
```
nasthagiri: I think this comment was missed.  If we do not verify that `ctx.weighted_max_score > 0` here, then we'll still be displaying scoring information when the course team sets the problem's weight to 0.

FYI @sstack22.  From my perspective, showing "0 points possible" can be confusing.  However, that's currently what CAPA does.  So I'm not flagging this as a merge-blocker.  However, we should consider whether this makes sense for all problem types.

tests/unit/test_basics.py
```diff
@@ -86,13 +89,25 @@ def test_get_configuration(self):
             )
         ])
 
+    @ddt.data(*[random.randint(1, 50) for _ in xrange(5)])  # pylint: disable=star-args
+    def test_grading_interface(self, weight):
+        """
+        Test that the methods required by the LMS grading interface work as expected.
+        """
+        self.block.weight = weight
+        # Max score is different from weight and should always equals 1 for drag and drop problems.
```
nasthagiri: Nit: typo - "equals" -> "equal"

drag_and_drop_v2/drag_and_drop_v2.py
nasthagiri: I LOVE the new DragNDrop.  In my testing of it on prod, I do find it confusing that once my "completion" is locked in - and then I choose to continue to experiment/test with the block, I stop getting feedback on my changes.  Any then it becomes _impossible_ for me to validate which of my changes is the correct answer.  Have you considered locking the UI once the learner "completes" the block in the "Assessment" mode - as is currently done in the "Standard" mode of the block?

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @sstack22 I believe this is a product decision, so please decide and create a task in JIRA if modifications are needed.

drag_and_drop_v2/drag_and_drop_v2.py
e-kolpakov: @nasthagiri I believe it was discussed at least once. I can't find any links to prove that, but as far as I can recall not locking UI after correct attempt was intentional - to allow student to experiment with the problem more without the risk of getting lower grade (i.e. to read error item feedback).

e-kolpakov:
@mtyaka I've merged this PR - please delete the branch if you don't need it anymore.

nasthagiri:
Thanks @e-kolpakov.

mtyaka:
@e-kolpakov Thanks, branch deleted.

e-kolpakov:
@mtyaka One more thing - looks like we both forgot to bump version number, so I've submitted another PR #109 to do that - if you could approve it it would be great :)

PR #110, opened 2016-12-07
Improve keyboard and SR instructions

e-kolpakov:
**Description:** This PR improves keyboard and SR instructions across DnDv2.

**JIRA:** [SOL-2157](https://openedx.atlassian.net/browse/SOL-2157)

**Sandbox:**

Currently at 644b6c7a3917fd171f917acb78ed79b5da6daee7
* LMS: http://pr14104.sandbox.opencraft.hosting/
* Studio: http://studio-pr14104.sandbox.opencraft.hosting/

**Testing instructions:**

Verify the following:
1. Draggable items now have SR only span saying "draggable" when laying on the floor and "draggable, grabbed" when someone picked them up (regardless of keyboard or mouse grab).
2. Space and Command+M can no longer be used to pick and drop items + updated help texts.
3. Keyboard help contains SR specific sections to reassure SR users that they can still use DnDv2 + more clear process description for everyone to see and hear.
4. Verbose help messages are removed from items.
5. If no description provided, zones default to "droppable" label.
6. Item bank SR help moved to aria-label attribute.
7. Feedback popup close buttons now read just "Close".

## Reviewers
- [x] OpenCraft: @mtyaka 
- [x] TNL: @cahrens
- [x] a11y: @cptvitamin

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -43,6 +43,7 @@
     display: inline-block;
     color: #5e5e5e;
     font-size: 0.875em;
+    margin-bottom: 0.5em;
 }
 
 .xblock--drag-and-drop .problem p {
@@ -444,8 +445,6 @@
 .xblock--drag-and-drop .btn-brand {
     display: inline-block;
     font-weight: normal;
-    background: #0079bc;
-    color: #fcfcfc;
     -webkit-transition: color 0.125s ease-in-out 0s, border-color 0.125s ease-in-out 0s, background 0.125s ease-in-out 0s, box-shadow 0.125s ease-in-out 0s;
     -moz-transition: color 0.125s ease-in-out 0s, border-color 0.125s ease-in-out 0s, background 0.125s ease-in-out 0s, box-shadow 0.125s ease-in-out 0s;
     transition: color 0.125s ease-in-out 0s, border-color 0.125s ease-in-out 0s, background 0.125s ease-in-out 0s, box-shadow 0.125s ease-in-out 0s;
@@ -470,6 +469,12 @@
 .xblock--drag-and-drop .btn-small {
   padding: 0.625em 0.625em;
   font-size: 0.875em;
+}
+
+.xblock--drag-and-drop .btn-link {
+    padding: 1px;
+    font-size: 0.875em;
+    font-weight: normal;
 }
 
 .xblock--drag-and-drop .btn-default:hover,
@@ -623,6 +628,10 @@
 
 .xblock--drag-and-drop .btn-icon {
     display: block;
+}
+
+.xblock--drag-and-drop .btn-link .btn-icon {
+    display: inline;
 }
 
 /*** ACTIONS TOOLBAR END ***/
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -126,7 +126,7 @@ function DragAndDropTemplates(configuration) {
             var item_description = h(
               'div',
               { id: item_description_id, className: descriptionClassName },
-              gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')
+              gettext('Press Ctrl-m on an item to select it for dropping, then navigate to the zone you want to drop it on.')
             );
         }
         children.splice(1, 0, item_description);
@@ -858,11 +858,7 @@ function DragAndDropBlock(runtime, element, configuration) {
     };
 
     var isActionKey = function(evt) {
-        var key = evt.which;
-        if (evt.ctrlKey || evt.metaKey) {
-            return key === M;
-        }
-        return key === RET || key === SPC;
+        return evt.which == RET || (evt.ctrlKey && evt.which == M);
     };
 
     var isSpaceKey = function(evt) {
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -301,12 +301,15 @@ function DragAndDropTemplates(configuration) {
                         h('h2.modal-window-title#'+labelledby_id, gettext('Keyboard Help'))
                     ]),
                     h('div.modal-content', [
-                        h('p', gettext('You can complete this problem using only your keyboard.')),
+                        h('p', {className:'sr'}, gettext('This is a screen reader-friendly problem')),
+                        h('p', gettext('You can complete this problem using only your keyboard by following the guidance below:')),
+                        h('p', {className:'sr'}, gettext('Drag and Drop problems consist of draggable items and dropzones. Users should select a draggable item with their keyboard and then navigate to an appropriate dropzone to drop it.')),
                         h('ul', [
-                            h('li', gettext('Use "Tab" and "Shift-Tab" to navigate between items and zones.')),
-                            h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')),
-                            h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" to drop the item on the current zone.')),
+                            h('li', gettext('Use only TAB and SHIFT+TAB to navigate between draggable items and drop zones.')),
+                            h('li', gettext('Press CTRL+M to select a draggable item (effectively picking it up).')),
+                            h('li', gettext('Navigate using TAB and SHIFT+TAB to the appropriate dropzone and press CTRL+M once more to drop it here.')),
                             h('li', gettext('Press "Esc" if you want to cancel the drop operation (for example, to select a different item).')),
+                            h('li', gettext('TAB back to the list of draggable items and repeat this process until all of the draggable items have been placed on their respective dropzones.')),
                         ])
                     ]),
                     h('div.modal-actions', [
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -227,7 +227,7 @@ function DragAndDropTemplates(configuration) {
                 },
                 [
                     h('p', { className: className }, zone.title),
-                    h('p', { className: 'zone-description sr' }, zone.description),
+                    h('p', { className: 'zone-description sr' }, zone.description || gettext("droppable")),
                     h(item_wrapper, renderCollection(itemTemplate, items_in_zone, ctx)),
                     zone_description
                 ]
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -53,6 +53,7 @@ function DragAndDropTemplates(configuration) {
 
     var itemTemplate = function(item, ctx) {
         // Define properties
+        var descriptionClassName = "sr description";
         var className = (item.class_name) ? item.class_name : "";
         var zone = getZone(item.zone, ctx) || {};
         if (item.has_image) {
@@ -118,18 +119,24 @@ function DragAndDropTemplates(configuration) {
             }
             var item_description = h(
                 'div',
-                { key: item.value + '-description', id: item_description_id, className: 'sr' },
+                { key: item.value + '-description', id: item_description_id, className: descriptionClassName },
                 description_content
             );
         } else {
             var item_description = h(
               'div',
-              { id: item_description_id, className: 'sr'},
+              { id: item_description_id, className: descriptionClassName },
               gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')
             );
         }
         children.splice(1, 0, item_description);
         children.splice(1, 0, item_content);
+
+        if (item.grabbed) {
+            var itemGrabbedSRNote = h('span', { className: 'sr dragged' }, gettext("draggable, grabbed"));
+            children.splice(2, 0, itemGrabbedSRNote);
+        }
+
         return (
             h(
                 'div.option',
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -255,44 +255,12 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
-    var popupTemplate = function(ctx) {
-        var popupSelector = 'div.popup';
-        if (ctx.popup_html && !ctx.last_action_correct) {
-            popupSelector += '.popup-incorrect';
-        }
-
-        return (
-            h(
-                "div.popup-wrapper",
-                {
-                    attributes: {
-                        'aria-live': 'polite',
-                        'aria-atomic': 'true',
-                        'aria-relevant': 'additions',
-                    }
-                },
-                [
-                    h(
-                        popupSelector,
-                        {
-                            style: {display: ctx.popup_html ? 'block' : 'none'},
-                        },
-                        [
-                            h('div.close.icon-remove-sign.fa-times-circle'),
-                            h('p.popup-content', {innerHTML: ctx.popup_html}),
-                        ]
-                    )
-                ]
-            )
-        )
-    };
-
     var keyboardHelpPopupTemplate = function(ctx) {
         var labelledby_id = 'modal-window-title-'+configuration.url_name;
         return (
             h('div.keyboard-help-dialog', [
                 h('div.modal-window-overlay'),
-                h('div.modal-window', {attributes: {role: 'dialog', 'aria-labelledby': labelledby_id}}, [
+                h('div.modal-window', {attributes: {role: 'dialog', 'aria-labelledby': labelledby_id, tabindex: -1}}, [
                     h('div.modal-header', [
                         h('h2.modal-window-title#'+labelledby_id, gettext('Keyboard Help'))
                     ]),
@@ -712,11 +680,10 @@ function DragAndDropBlock(runtime, element, configuration) {
         // Show dialog
         var $keyboardHelpDialog = $root.find('.keyboard-help-dialog');
         $keyboardHelpDialog.find('.modal-window-overlay').show();
-        $keyboardHelpDialog.find('.modal-window').show();
+        $keyboardHelpDialog.find('.modal-window').show().focus();
 
         // Handle focus
         $focusedElement = $(':focus');
-        focusModalButton();
 
         // Set up event handlers
         $(document).on('keydown', keyboardEventDispatcher);
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -434,7 +434,7 @@ function DragAndDropTemplates(configuration) {
                         h(
                             'span.sr',
                             {
-                                innerHTML: gettext("Close item feedback popup")
+                                innerHTML: gettext("Close")
                             }
                         ),
                         h(
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -2,9 +2,6 @@ function DragAndDropTemplates(configuration) {
     "use strict";
     var h = virtualDom.h;
 
-    var keyboardHelpButtonClass = "keyboard-help-button";
-    var keyboardHelpText = gettext('Keyboard Help');
-
     var itemSpinnerTemplate = function(item) {
         if (!item.xhr_active) {
             return null;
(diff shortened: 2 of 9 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=110"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -301,12 +301,15 @@ function DragAndDropTemplates(configuration) {
                         h('h2.modal-window-title#'+labelledby_id, gettext('Keyboard Help'))
                     ]),
                     h('div.modal-content', [
-                        h('p', gettext('You can complete this problem using only your keyboard.')),
+                        h('p', {className:'sr'}, gettext('This is a screen reader-friendly problem')),
```
mtyaka: Nit: `h('p.sr', ...)`

mtyaka: Missing period `.` at end of sentence.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -301,12 +301,15 @@ function DragAndDropTemplates(configuration) {
...
+                        h('p', gettext('You can complete this problem using only your keyboard by following the guidance below:')),
```
mtyaka: Should we move this paragraph down one line so that it's the last paragraphs before the `ul` element?
It's a bit weird to end the paragraph with "... following the guidance below:" and then there's another paragraph between this sentence and the list of guidance items.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -531,6 +538,7 @@ function DragAndDropTemplates(configuration) {
                     problemHeader,
                     h('p', {innerHTML: ctx.problem_html}),
                 ]),
+                h('div', {'className': 'sr'}, [forwardKeyboardHelpButtonTemplate(ctx)]),
```
mtyaka: Nit: `h('div.sr', ...)`

mtyaka: According to [Mark's comment on TNL-6022](https://openedx.atlassian.net/browse/TNL-6022?focusedCommentId=231733&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-231733), wrapping focusable content in `sr` only tags is a bad idea.
Can we go with the [design proposed by Christine](https://openedx.atlassian.net/browse/TNL-6022?focusedCommentId=231757&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-231757) instead?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -120,25 +120,23 @@ function DragAndDropTemplates(configuration) {
...
-        if (item.grabbed) {
-            var itemGrabbedSRNote = h('span', { className: 'sr dragged' }, gettext("draggable, grabbed"));
-            children.splice(2, 0, itemGrabbedSRNote);
-        }
+        var itemSRNote = h(
+            'span',
+            { className: 'sr draggable' },
+            (item.grabbed) ? gettext("draggable, grabbed") : gettext("draggable")
+        );
+        children.splice(2, 0, itemSRNote);
```
mtyaka: I'm not sure why we're doing this confusing splicing. Couldn't we simply do this at this point:

```javascript
var children = [itemSpinnerTemplate(item), item_content, itemSRNote, item_description];
```

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -519,14 +519,17 @@ function DragAndDropTemplates(configuration) {
...
-            item_bank_properties.attributes = {
-                'tabindex': 0,
-                'dropzone': 'move',
-                'aria-dropeffect': 'move',
-                'role': 'button'
-            };
+            item_bank_properties.attributes['tabindex'] = 0;
+            item_bank_properties.attributes['dropzone'] = 'move';
+            item_bank_properties.attributes['aria-dropeffect'] = 'move';
+            item_bank_properties.attributes['role'] = 'button';
```
mtyaka: This overrides the `role="group"` from above, and since the item bank isn't really a button, I think we should remove this?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -116,20 +120,24 @@ function DragAndDropTemplates(configuration) {
...
-              'div',
-              { id: item_description_id, className: 'sr'},
-              gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')
-            );
         }
         children.splice(1, 0, item_description);
         children.splice(1, 0, item_content);
+
+        var itemSRNote = h(
+            'span',
```
mtyaka: Nit: could use the class shorthand syntax: `span.sr.draggable`

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -116,20 +120,24 @@ function DragAndDropTemplates(configuration) {
...
+            { className: 'sr draggable' },
+            (item.grabbed) ? gettext("draggable, grabbed") : gettext("draggable")
```
mtyaka: Mark's example has a comma before the word "draggable":

```
(item.grabbed) ? gettext(", draggable, grabbed") : gettext(", draggable")
```

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -116,20 +120,24 @@ function DragAndDropTemplates(configuration) {
...
-            var item_description = h(
-              'div',
-              { id: item_description_id, className: 'sr'},
-              gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')
-            );
         }
         children.splice(1, 0, item_description);
         children.splice(1, 0, item_content);
+
+        var itemSRNote = h(
```
mtyaka: See [Marks' comment](https://openedx.atlassian.net/browse/TNL-6031?focusedCommentId=230658&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-230658) about having to use an `aria-live` region in addition to the SR note, although I'm not sure where that region would go.

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @mtyaka Sure, for some reason I haven't got an update from JIRA on that.

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @mtyaka I wonder if it has some special meaning to screen readers. You have certainly [added it](https://github.com/edx-solutions/xblock-drag-and-drop-v2/commit/0e07b79936020a862f06ca662cb83e34de8cfa96#diff-cef30416b48c3e5d4dc05c6922a2729bR296) for a reason - do you remember what was that?

e-kolpakov:
@mtyaka reviewnotes addressed - please take a look.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: @e-kolpakov I just copied these properties from the zone template when making the item bank act like a droppable zone. We will be removing `role="button"` from zones and draggables for [TNL-6029](https://openedx.atlassian.net/browse/TNL-6029), so you can either delete this line now or leave it in to be removed in [TNL-6029](https://openedx.atlassian.net/browse/TNL-6029) - whatever you prefer.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -294,12 +300,15 @@ function DragAndDropTemplates(configuration) {
...
+                        h('p', gettext('You can complete this problem using only your keyboard by following the guidance below:')),
                         h('ul', [
-                            h('li', gettext('Use "Tab" and "Shift-Tab" to navigate between items and zones.')),
-                            h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" on an item to select it for dropping, then navigate to the zone you want to drop it on.')),
-                            h('li', gettext('Press "Enter", "Space", "Ctrl-m", or "⌘-m" to drop the item on the current zone.')),
+                            h('li', gettext('Use only TAB and SHIFT+TAB to navigate between draggable items and drop zones.')),
+                            h('li', gettext('Press CTRL+M to select a draggable item (effectively picking it up).')),
+                            h('li', gettext('Navigate using TAB and SHIFT+TAB to the appropriate dropzone and press CTRL+M once more to drop it here.')),
                             h('li', gettext('Press "Esc" if you want to cancel the drop operation (for example, to select a different item).')),
```
mtyaka: Sorry I missed this before - can we change `"Esc"` to `ESC` for consistency?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -370,7 +379,7 @@ function DragAndDropTemplates(configuration) {
         }
         return(
             h("section.action-toolbar-item.sidebar-buttons", {}, [
-                sidebarButtonTemplate("keyboard-help-button", "fa-question", gettext('Keyboard Help')),
+                // sidebarButtonTemplate(keyboardHelpButtonClass, "fa-question", keyboardHelpText),
```
mtyaka: Remove commented line.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -441,6 +450,23 @@ function DragAndDropTemplates(configuration) {
...
 
+    var forwardKeyboardHelpButtonTemplate = function(ctx) {
+        return h(
+            'button.unbutton.btn-default.btn-link.'+keyboardHelpButtonClass,
+            {attributes: {tabindex: 0}},
+            [
+                h(
+                    "span.btn-icon.fa.fa-keyboard-o",
+                    {attributes: {"aria-hidden": true}},
+                    []
```
mtyaka: nit: Last argument (`[]`) can be omitted.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -441,6 +450,23 @@ function DragAndDropTemplates(configuration) {
...
+                ),
+                // appending space is the simplest way to avoid sticking text to the button, but also to have
+                // them underlined together on hover. When margin was used there was a gap in underlining
+                " " + keyboardHelpText
```
mtyaka: Instead of concatenating the two strings, we can insert the space as a separate text node, which is a bit cleaner conceptually:

`h(..., [h('span' ...), " ", keyboardHelpText])`

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -2,6 +2,9 @@ function DragAndDropTemplates(configuration) {
     "use strict";
     var h = virtualDom.h;
 
+    var keyboardHelpButtonClass = "keyboard-help-button";
+    var keyboardHelpText = gettext('Keyboard Help');
```
mtyaka: These two variables are now only used in one place and can be defined inline.

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @mtyaka I'll let TNL-6029 handle it in order to avoid possible merge conflicts.

e-kolpakov:
@mtyaka Notes addressed, please take another look.

@smarnach due to problems with creating sandboxes from OC IM console, I've used http://pr14104.sandbox.opencraft.hosting/ - please let me know if you switch it to another appserver.

smarnach:
@e-kolpakov The PR for that sandbox is merged, so it's yours now. :)

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -441,6 +446,23 @@ function DragAndDropTemplates(configuration) {
         )
     };
 
+    var forwardKeyboardHelpButtonTemplate = function(ctx) {
+        return h(
+            'button.unbutton.btn-default.btn-link.'+"keyboard-help-button",
```
mtyaka: This should be changed to a single literal string now that we're no longer using a variable:

```javascript
'button.unbutton.btn-default.btn-link.keyboard-help-button'
```

e-kolpakov:
@cahrens @cptvitamin This PR is ready for your review.

cahrens:
@e-kolpakov I should be able to review this on Monday.

e-kolpakov:
@cahrens :ok:

cahrens:
After I drop an item, if a confirmation dialog shows up, focus is not being moved into it (to the close button). I have to Shift-Tab to move back into the dialog, then press Enter to trigger the close button.

image

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -441,6 +446,23 @@ function DragAndDropTemplates(configuration) {
...
+            {attributes: {tabindex: 0}},
```
cahrens: Why is it necessary to explicitly set a tabindex of 0?

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -472,6 +473,21 @@
   font-size: 0.875em;
 }
 
+.xblock--drag-and-drop .btn-link {
+    padding: 1px;
+    font-size: 0.875em;
+}
+
+.xblock--drag-and-drop .btn-default.btn-link:hover,
```
cahrens: Why is it necessary to handle the focus border explicitly? What is preventing the default behavior from just working?

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
```diff
@@ -521,6 +521,14 @@ msgstr "Pläçéd ïn: {zone_title}  Ⱡ'σяєм ιρѕυм ∂σłσя ѕιт
 msgid "Correctly placed in: {zone_title}"
 msgstr "Çörréçtlý pläçéd ïn: {zone_title} Ⱡ'σяєм ιρѕυм ∂σłσя ѕιт αмєт, #"
 
+#: drag_and_drop_v2/public/js/drag_and_drop.js
+msgid ", draggable, grabbed"
```
cahrens: Is there documentation on how translations are managed in this repo?

cahrens:
@e-kolpakov I'm done with my first pass.

mtyaka:
> After I drop an item, if a confirmation dialog shows up, focus is not being moved into it (to the close button). I have to Shift-Tab to move back into the dialog, then press Enter to trigger the close button.

Focus issues were taken care of in #111.

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @cahrens according to [W3C spec](https://www.w3.org/TR/html5/editing.html#sequential-focus-navigation-and-the-tabindex-attribute), setting it on a `button` element is not required, as by spec buttons are among the elements that have the tabindex focus flag set (but "modulo platform conventions"). On the other hand, setting `tabindex: 0` just sets that flag.

It is safe both to keep it or remove it, but keeping it has a slight advantages of:
1. No need to retest it now.
2. No need to redeploy sandbox.
3. In future if the type of element changes (i.e. button => span, e.g. for styling purposes) it will protect from button being inaccessible using keyboard.

drag_and_drop_v2/public/css/drag_and_drop.css
e-kolpakov: @cahrens `unbotton` class sets border to none. I'm not 100% sure, but it seems like `unbotton` does not do anything now, but at some moment it was used to "unapply" edX button styling from buttons.

Before we start discussing it I'd suggest we don't hijack this PR for going over the styling and removing dead/useless rules. If that's required we can have a separate "CSS cleanup" task/story.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -504,26 +526,29 @@ function DragAndDropTemplates(configuration) {
...
+            }
+        };
         if (ctx.item_bank_focusable) {
-            item_bank_properties.attributes = {
-                'tabindex': 0,
-                'dropzone': 'move',
-                'aria-dropeffect': 'move',
-                'role': 'button'
-            };
+            item_bank_properties.attributes['tabindex'] = 0;
```
e-kolpakov: @cahrens Line 1327: 

```
// In assessment mode, it is possible to move items back to the bank, so the bank should be able to
// gain focus while keyboard placement is in progress.
var item_bank_focusable = (state.keyboard_placement_mode || state.showing_answer) &&
        configuration.mode === DragAndDropBlock.ASSESSMENT_MODE;
```

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
e-kolpakov: @cahrens Yes - at the end of the README there are [two sections about translations](https://github.com/edx-solutions/xblock-drag-and-drop-v2#i18n-compatibility).

Spoiler: it is far from ideal, but we were unable to come up with anything that does not require significant updates to i18n_tool.

drag_and_drop_v2/public/js/drag_and_drop.js
cahrens: I think you should remove explicitly setting the tabindex on buttons--as you mentioned, it is not necessary. Furthermore, changing the type of element from button to span "for styling purposes" would introduce a host of other issues; the reason to use native html components is that you get the correct behavior for screen-readers and keyboard navigation, without having to use aria attributes to make something act "like a button".

drag_and_drop_v2/public/css/drag_and_drop.css
cahrens: I understand limiting scope, but this PR is actually introducing this css. I'd rather not see custom handling of focus indicators added-- it's much better to use the default platform styling (which may be different on different browsers).

e-kolpakov: @cahrens Beccause this css is needed to override `btn-default` styling. Without this CSS the button that supposed to look like a link gets blue border on hover. If we remove `btn-default` rules controlling hover styling (they're exactly below), than all the other buttons loose blue border on hover, which they suppose to have.

Completely removing custom button styling and relying on edX styling is not a good idea too, as (1) it would require significant overhaul of styling and (2) DnDv2 is used with non-edX LMS, so removing custom styles will introduce regressions there.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -69,7 +70,8 @@ function DragAndDropTemplates(configuration) {
             'draggable': !item.drag_disabled,
             'aria-grabbed': item.grabbed,
             'data-value': item.value,
-            'tabindex': item.focusable ? 0 : undefined
+            'tabindex': item.focusable ? 0 : undefined,
+            'aria-live': 'polite'
```
cahrens: I don't think this should have aria-live set.

cahrens:
I have one question (just want to confirm my understanding is correct), and there is a failing test that needs to be fixed. Otherwise 👍

drag_and_drop_v2/public/js/drag_and_drop.js
e-kolpakov: @cahrens Yes - it is here to read out changes in SR block with "draggable" / "draggable, dropped" content. It was [explicitly requested by Mark](https://openedx.atlassian.net/browse/TNL-6031?focusedCommentId=230658&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-230658)

e-kolpakov:
@cahrens Should I merge now, or wait for @cptvitamin 's review?

cptvitamin:
@e-kolpakov This is great.  I have a couple comments.  Focus is moved to the OK button, which is at the end of the modal dialog boxes DOM. This means that a screen reader user has to move backwards if they want to read the contents of the dialog box.  That's why it is best to make the close button the very first element in the dialog box's DOM. Alternatively, you can move focus to the div itself (the one with the role=dialog on it) by placing a tabindex=-1 attribute on it and moving focus to the div.

 Also, are we doing anything to prevent default on the arrow keys while the keyboard dialog box is open?  I can't seem to use the arrow keys (down arrow) to read through the content in the keyboard help dialog box on the windows platform, specifically with NVDA.

e-kolpakov:
@cptvitamin Implemented alternative suggestion.

As of arrow keys - there's a custom keypress handler, but it does not prevent default as far as I can tell. When I use up/down keys, the popup is not affected, but underlying page is scrolled.

e-kolpakov:
@cptvitamin Friendly reminder to take another look at this. And Happy new year!

e-kolpakov:
@cahrens This PR is a little bit stuck again - could you please talk to @cptvitamin to either arrange another review or sign off an approval?

cahrens:
@e-kolpakov Our office has been closed for holidays. @cptvitamin is in the office today though.

e-kolpakov:
@cahrens @cptvitamin Friendly reminder to review/approve this PR.

cptvitamin:
👍  this looks good @e-kolpakov

PR #111, opened 2016-12-08
Usability improvements for screen reader users

mtyaka:
## Description

Implements these usability improvements suggested by Mark S:

- [TNL-6024](https://openedx.atlassian.net/browse/TNL-6024): Move Close button to beginning of dialog.
- [TNL-6020](https://openedx.atlassian.net/browse/TNL-6020): Focus is not moved to feedback dialog box
- [TNL-6028](https://openedx.atlassian.net/browse/TNL-6028): Add a method for moving focus back to the beginning of the problem after dropping and item,

## JIRA Ticket

- [SOL-2158](https://openedx.atlassian.net/browse/SOL-2158)

## Sandbox URLs

Provisioned with 94986e590e95ddb60a4008286c27116b10785c51.
You might have to add a certificate security exception to be able to access the sandbox.

- LMS: http://pr14132.sandbox.stage.opencraft.hosting/
- Studio: http://studio-pr14132.sandbox.stage.opencraft.hosting/

### Keyboard help dialog close button focus

1. Open the Keyboard help dialog
2. The close button should be focused automatically, hitting the RET key will close the dialog

### Per item feedback close button

This only works in standard mode since in assessment mode feedback popup is not displayed when dropping an item.

1. Pick an item that has a negative feedback message defined.
2. Drop it onto a wrong zone.
3. The feedback popup will be shown. Verify that the close button is automatically focused.
4. Hitting the RET key closes the feedback popup and automatically moves focus back to the first item in the item bank.
5. Repeat by dropping an item with a positive feedback message onto the correct zone.

### Go to beginning button

1. TAB through the items and zones until you reach the "Go to Beginning" button which is hidden by default.
2. When it receives focus, the "Go to Beginning" button becomes visible.
3. TAB again to move focus to next button - the "Go to Beginning" button should automatically become hidden again.
4. Shift-TAB back to the "Go to Beginning" item to make it visible.
5. Clicking (or pressing the RET key on) the button should move focus to the first item in the item bank and the button should become hidden.
6. Verify that the button is not focusable when the problem is finished, user has no more attemtps, or there are no more items in the item bank.

### Method for moving focus back to beginning after dropping an item

In standard mode this only works with items that have no per-item feedback message defined. In assessment mode it works with any item since feedback messages are not shown in assessment mode on drop.

1. Pick an item and drop it onto desired zone.
2. Press the TAB key. The focus should automatically move to the "Go to Beginning" button, no matter which zone you dropped the item into.

## Author notes

I placed the "Go to Beginning" button as the first item in the section that contains other buttons (Keyboard Help, Reset, Show Answer). The area is becoming somewhat crowded, but that should be fixed #110, which moves the "Keyboard Help" button out of that area to the beginning of the problem.

screen shot 2016-12-12 at 08 54 15

## Reviewers

- [x] OpenCraft: @haikuginger
- [x] TNL: @cahrens 
- [x] a11y: @cptvitamin

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -671,8 +671,11 @@
     padding: 7px;
     background-color: #e5e5e5;
     text-align: left;
-    direction: ltr;
     z-index: 1500;
+}
+
+.rtl .xblock--drag-and-drop .modal-window {
+    transform: translate(50%, -50%);
 }
 
 .xblock--drag-and-drop .modal-dismiss-button {
@@ -681,6 +684,12 @@
     top: 3px;
     right: 3px;
     padding: 5px 8px;
+}
+
+
+.rtl .xblock--drag-and-drop .modal-dismiss-button {
+    right: inherit;
+    left: 3px;
 }
 
 .xblock--drag-and-drop .modal-header h2 {
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -675,11 +675,24 @@
     z-index: 1500;
 }
 
+.xblock--drag-and-drop .modal-dismiss-button {
+    font-size: 24px;
+    position: absolute;
+    top: 3px;
+    right: 3px;
+    padding: 5px 8px;
+}
+
+.xblock--drag-and-drop .modal-header h2 {
+    height: 30px;
+    line-height: 30px;
+    margin-bottom: 5px;
+}
+
 .xblock--drag-and-drop .modal-content {
     border-radius: 5px;
     background-color: #ffffff;
-    margin-bottom: 5px;
-    padding: 5px;
+    padding: 8px;
 }
 
 .xblock--drag-and-drop .modal-content li {
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -261,8 +261,12 @@ function DragAndDropTemplates(configuration) {
             h('div.keyboard-help-dialog', [
                 h('div.modal-window-overlay'),
                 h('div.modal-window', {attributes: {role: 'dialog', 'aria-labelledby': labelledby_id, tabindex: -1}}, [
+                    h('button.modal-dismiss-button.unbutton', {attributes: {tabindex: 1}}, [
+                        h('span.fa.fa-remove', {attributes: {'aria-hidden': true}}),
+                        h('span.sr', gettext('Close'))
+                    ]),
                     h('div.modal-header', [
-                        h('h2.modal-window-title#'+labelledby_id, gettext('Keyboard Help'))
+                        h('h2.modal-window-title', {id: labelledby_id}, gettext('Keyboard Help'))
                     ]),
                     h('div.modal-content', [
                         h('p.sr', gettext('This is a screen reader-friendly problem.')),
@@ -275,9 +279,6 @@ function DragAndDropTemplates(configuration) {
                             h('li', gettext('Press ESC if you want to cancel the drop operation (for example, to select a different item).')),
                             h('li', gettext('TAB back to the list of draggable items and repeat this process until all of the draggable items have been placed on their respective dropzones.')),
                         ])
-                    ]),
-                    h('div.modal-actions', [
-                        h('button.modal-dismiss-button', gettext("OK"))
                     ])
                 ])
             ])
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -629,7 +629,7 @@ function DragAndDropBlock(runtime, element, configuration) {
             // Set up event handlers:
 
             $element.on('click', '.item-feedback-popup .close-feedback-popup-button', closePopupEventHandler);
-            $element.on('keydown', '.item-feedback-popup .close-feedback-popup-button', registerPopupCloseButtonKeydown);
+            $element.on('keydown', '.item-feedback-popup .close-feedback-popup-button', closePopupKeydownHandler);
             $element.on('keyup', '.item-feedback-popup .close-feedback-popup-button', preventFauxPopupCloseButtonClick);
 
             $element.on('click', '.submit-answer-button', doAttempt);
@@ -736,8 +736,12 @@ function DragAndDropBlock(runtime, element, configuration) {
     // a click event on keyup if the close button received a keydown event prior to the keyup.
     var _popup_close_button_keydown_received = false;
 
-    var registerPopupCloseButtonKeydown = function(evt) {
+    var closePopupKeydownHandler = function(evt) {
         _popup_close_button_keydown_received = true;
+        // Don't let user tab out of the button until the feedback is closed.
+        if (evt.which === TAB) {
+            evt.preventDefault();
+        }
     };
 
     var preventFauxPopupCloseButtonClick = function(evt) {
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -261,7 +261,7 @@ function DragAndDropTemplates(configuration) {
             h('div.keyboard-help-dialog', [
                 h('div.modal-window-overlay'),
                 h('div.modal-window', {attributes: {role: 'dialog', 'aria-labelledby': labelledby_id, tabindex: -1}}, [
-                    h('button.modal-dismiss-button.unbutton', {attributes: {tabindex: 1}}, [
+                    h('button.modal-dismiss-button.unbutton', [
                         h('span.fa.fa-remove', {attributes: {'aria-hidden': true}}),
                         h('span.sr', gettext('Close'))
                     ]),
@@ -321,8 +321,7 @@ function DragAndDropTemplates(configuration) {
                     'button.unbutton.btn-default.btn-small',
                     {
                         className: buttonClass,
-                        disabled: options.disabled || options.spinner || false,
-                        attributes: {tabindex: 0}
+                        disabled: options.disabled || options.spinner || false
                     },
                     [
                         h("span.btn-icon.fa", {className: iconClass, attributes: {"aria-hidden": true}}),
@@ -755,7 +754,7 @@ function DragAndDropBlock(runtime, element, configuration) {
     };
 
     var focusModalButton = function() {
-        $root.find('.keyboard-help-dialog .modal-dismiss-button ').focus();
+        $root.find('.keyboard-help-dialog .modal-dismiss-button').focus();
     };
 
     var showKeyboardHelp = function(evt) {
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -310,21 +310,22 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
-    var sidebarButtonTemplate = function(buttonClass, iconClass, buttonText, disabled, spinner) {
-        if (spinner) {
+    var sidebarButtonTemplate = function(buttonClass, iconClass, buttonText, options) {
+        options = options || {};
+        if (options.spinner) {
             iconClass = 'fa-spin.fa-spinner';
         }
         return (
             h('span.sidebar-button-wrapper', {}, [
                 h(
-                    'button.unbutton.btn-default.btn-small.'+buttonClass,
-                    {disabled: disabled || spinner || false, attributes: {tabindex: 0}},
+                    'button.unbutton.btn-default.btn-small',
+                    {
+                        className: buttonClass,
+                        disabled: options.disabled || options.spinner || false,
+                        attributes: {tabindex: 0}
+                    },
                     [
-                        h(
-                            "span.btn-icon.fa." + iconClass,
-                            {attributes: {"aria-hidden": true}},
-                            []
-                        ),
+                        h("span.btn-icon.fa", {className: iconClass, attributes: {"aria-hidden": true}}),
                         buttonText
                     ]
                 )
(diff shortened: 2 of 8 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=111&repo=edx-solutions%2Fxblock-drag-and-drop-v2"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

haikuginger:
@mtyaka, a couple early notes.

The focus for the feedback dialog box looks good!

For the "method for moving focus back to the beginning", I think it might be better if we arrived at it immediately after pressing tab, rather than needing to tab through to the end of the problem space; if we need to do that, what's the benefit of having it vs. Shift-tab?

The Keyboard Help change fits the request, but the dialog as a whole is a bit tricky to navigate with a keyboard - it'd be great if it was something you could just tab through without needing to mess with arrow keys and keyboard modifiers. That's possibly out of scope, though.

mtyaka:
Thanks for the feedback @haikuginger!

> For the "method for moving focus back to the beginning", I think it might be better if we arrived at it immediately after pressing tab, rather than needing to tab through to the end of the problem space; if we need to do that, what's the benefit of having it vs. Shift-tab?

After you drop an item using the keyboard, the next TAB keyprass **will** take you to the 'Go to Beginning' button immediately, no matter which zone you dropped the item into, unless there is a bug in my code :)

> The Keyboard Help change fits the request, but the dialog as a whole is a bit tricky to navigate with a keyboard - it'd be great if it was something you could just tab through without needing to mess with arrow keys and keyboard modifiers. That's possibly out of scope, though.

Eugeny is making some changes to the help dialog in #110. He's also moving the 'Keyboard Help' button to the top of the problem.

haikuginger:
> After you drop an item using the keyboard, the next TAB keyprass will take you to the 'Go to Beginning' button immediately, no matter which zone you dropped the item into, unless there is a bug in my code.

Ah, it might be because I was getting feedback popups. I'll be sure to test in assessment mode as well where those don't happen.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -335,21 +336,21 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
-    var sidebarButtonTemplate = function(buttonClass, iconClass, buttonText, disabled, spinner) {
-        if (spinner) {
+    var sidebarButtonTemplate = function(buttonClass, iconClass, buttonText, options) {
```
haikuginger: I like the small refactor here - changing it to take an options object rather than continually add more and more parameters.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -335,21 +336,21 @@ function DragAndDropTemplates(configuration) {
...
+        options = options || {};
+        if (options.spinner) {
             iconClass = 'fa-spin.fa-spinner';
         }
         return (
             h('span.sidebar-button-wrapper', {}, [
                 h(
-                    'button.unbutton.btn-default.btn-small.'+buttonClass,
-                    {disabled: disabled || spinner || false, attributes: {tabindex: 0}},
+                    'button.unbutton.btn-default.btn-small', {
```
haikuginger: Nit: Should the object start on a new line?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -360,18 +361,30 @@ function DragAndDropTemplates(configuration) {
...
+                options
             );
         }
+        var go_to_beginning_button_class = 'go-to-beginning-button';
+        if (!ctx.show_go_to_beginning_button) {
+            go_to_beginning_button_class += ' sr';
+        }
         return(
             h("section.action-toolbar-item.sidebar-buttons", {}, [
+                sidebarButtonTemplate(go_to_beginning_button_class, "fa-arrow-up", gettext("Go to Beginning"), {
```
haikuginger: Nit: Since this has to cross multiple lines anyway, I'd like to see one argument per line. Makes it easier to diff later on.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -360,18 +361,30 @@ function DragAndDropTemplates(configuration) {
...
+                    disabled: ctx.disable_go_to_beginning_button
+                }),
                 sidebarButtonTemplate("keyboard-help-button", "fa-question", gettext('Keyboard Help')),
-                sidebarButtonTemplate("reset-button", "fa-refresh", gettext('Reset'), ctx.disable_reset_button),
+                sidebarButtonTemplate("reset-button", "fa-refresh", gettext('Reset'), {
+                    disabled: ctx.disable_reset_button
```
haikuginger: As on line 381.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -634,6 +648,20 @@ function DragAndDropBlock(runtime, element, configuration) {
...
 
+            // We need to register both mousedown and click event handlers because in some browsers the blur
+            // event is emitted right after mousedown, hiding our button and preventing the click event from
+            // being emitted.
+            // We still need the click handler to catch keydown events (other than RET which is handled below),
+            // since in some browser/OS combinations some other keyboard button presses (for example space bar)
+            // are also treated as clicks,
+            $element.on('mousedown click', '.go-to-beginning-button', onGoToBeginningButtonClick);
+            $element.on('keydown', '.go-to-beginning-button', function(evt) {
+                runOnKey(evt, RET, onGoToBeginningButtonClick);
```
haikuginger: Nit: Not sure what the recommended style is for inline functions in JS, but as on line 381, I'd like to see a cleaner separation of one argument per line.

haikuginger:
:+1:, although there are some nits I'd like to see addressed if possible.

- [x] I tested this: Using the old version of the code, I verified the usability bugs in question. Then, I installed this branch and verified that they were no longer present, following the PR instructions.
- [x] I read through the code
- [x] ~~Includes documentation~~ N/A; bugfix

mtyaka:
Rebased from aa6dfd5714240846c9925b0d2aeb25f10a1eb0e0.

mtyaka:
Thanks @haikuginger! I addressed most of your nits and rebased.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: @haikuginger I decided to leave it as it is for now, because the surrounding code uses the same formatting. We use a mixture of different formatting styles. It would be great if we could make that more consistent at some point.

mtyaka:
@cahrens This is ready for your review.

cahrens:
@mtyaka I have created https://openedx.atlassian.net/browse/TNL-6120 for the review. I think we can get to it in the next couple of days.

mtyaka:
Thanks @cahrens!

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -290,8 +290,12 @@ function DragAndDropTemplates(configuration) {
             h('div.keyboard-help-dialog', [
                 h('div.modal-window-overlay'),
                 h('div.modal-window', {attributes: {role: 'dialog', 'aria-labelledby': labelledby_id}}, [
+                    h('button.modal-dismiss-button.unbutton', {attributes: {tabindex: 1}}, [
```
cahrens: This button should not have a tabindex set (buttons are naturally focusable). You can check for errors like this using Chrome's accessibility audit tools, or the Axe Chrome plugin.

image

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -666,11 +666,24 @@
     z-index: 1500;
 }
 
+.xblock--drag-and-drop .modal-dismiss-button {
+    font-size: 24px;
+    position: absolute;
+    top: 3px;
+    right: 3px;
```
cahrens: Have you verified this looks OK with rtl languages? I'm asking because only "right" is specified here.

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
```diff
@@ -582,8 +582,12 @@ msgstr ""
 "séléçt ä dïfférént ïtém). Ⱡ'σяєм ιρѕυм ∂σłσя ѕιт αмєт, ¢#"
 
 #: public/js/drag_and_drop.js
-msgid "OK"
-msgstr "ÖK Ⱡ'σя#"
+msgid "Close"
+msgstr "Çlösé Ⱡ'σя#"
```
cahrens: msgstr is empty in text.po.

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
```diff
@@ -582,8 +582,12 @@ msgstr ""
...
+
+#: public/js/drag_and_drop.js
+msgid "Go to Beginning"
+msgstr "Gö tö Bégïnnïng Ⱡ'σяєм ιρѕυм ∂σłσя ѕιт α#"
```
cahrens: msgstr is empty in text.po.

tests/integration/test_render.py
```diff
@@ -213,6 +214,30 @@ def test_popup(self):
         self.assertEqual(popup_content.text, "")
         self.assertEqual(popup_wrapper.get_attribute('aria-live'), 'polite')
 
+    @data(None, Keys.RETURN)
+    def test_go_to_beginning_button(self, action_key):
+        self.load_scenario()
+        self.scroll_down(250)
+
+        button = self._get_go_to_beginning_button()
+        self.assertEqual(button.get_attribute('tabindex'), '0')
```
cahrens: Why is it important that the tabindex is 0? It seems that you can always tab to the Review button.

tests/integration/test_render.py
```diff
@@ -213,6 +214,30 @@ def test_popup(self):
...
+        # Button is only visible to screen reader users by default.
+        self.assertIn('sr', button.get_attribute('class').split())
+        # Set focus to the element (cannot find a way to do this without execute_script).
+        self.browser.execute_script('$("button.go-to-beginning-button").focus()')
```
cahrens: I think you could continue tabbing until the button is focused.

cahrens:
@mtyaka I'm done with this pass-- the improvements are nice!

drag_and_drop_v2/public/css/drag_and_drop.css
mtyaka: Ah, I missed that, thank you.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: I left the tabindex here by accident while testing issues with integration tests. Thanks for spotting this, I will remove the attribute.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -335,21 +336,22 @@ function DragAndDropTemplates(configuration) {
...
         return (
             h('span.sidebar-button-wrapper', {}, [
                 h(
-                    'button.unbutton.btn-default.btn-small.'+buttonClass,
-                    {disabled: disabled || spinner || false, attributes: {tabindex: 0}},
+                    'button.unbutton.btn-default.btn-small',
+                    {
+                        className: buttonClass,
+                        disabled: options.disabled || options.spinner || false,
+                        attributes: {tabindex: 0}
```
mtyaka: I'm not sure... it was there before, but it seems to be redundant since we're using naturally focusable button elements anyway. I'll remove it.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -885,14 +974,18 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+
     var focusFirstDraggable = function() {
         $root.find('.item-bank .option').first().focus();
     };
 
     var focusItemFeedbackPopup = function() {
         var popup = $root.find('.item-feedback-popup');
         if (popup.length && popup.is(":visible")) {
-            popup.focus();
+            popup.find('.close-feedback-popup-button').focus();
```
mtyaka: I tried that, but for some reason it does not work.

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
mtyaka: @cahrens I'm not sure what you mean? It is not empty in this file. It is empty in the english text.po file, but all msgstrs are empty in the english text.po file.

tests/integration/test_render.py
mtyaka: Not important at all, I will remove this line, thanks.

tests/integration/test_render.py
mtyaka: I tried that initially, but it didn't work. I did some more digging and found out why it does not work - when you TAB to the element in firefox, the `focus` event is not fired unless Firefox is the focused window, which it usually isn't when running the tests. I expanded the comment to explain this.

mtyaka:
Thanks @cahrens! I addressed your comments and updated the sandbox.

drag_and_drop_v2/translations/eo/LC_MESSAGES/text.po
cahrens: Oh, I see. Thanks!

cahrens:
This is related to this specific PR, but I found an accessibility error on this particular problem (bucket demo). It's related to the final (5th) zone.

image

cahrens:
👍 on this PR, but please follow up on the invalid ARIA attribute item that I noted.

mtyaka:
Thanks @cahrens! I'll address the ARIA attribute problem while working on https://openedx.atlassian.net/browse/SOL-2156.

mtyaka:
@cptvitamin This is waiting for your review.

cptvitamin:
@mtyaka this is amazing work!  The experience for keyboard and screen reader users was very logical and efficient.  The only issue I had was that when the feedback modal appeared, I could tab outside of it without closing it.  Ideally, this modal would function like other modals on the platform, trapping focus inside of it until the user explicitly closes the modal.  This can be difficult to accomplish in certain circumstances.  If this is one of those cases, I'm willing to let it go, but if you can easily trap focus within the feedback dialog box, that is how it should behave.  Otherwise 👍  excellent work!

mtyaka:
Rebased from 7e849a4327131520d0f2188203e1505fce8bc26b.

PR #112, opened 2016-12-15
a11y Interaction Improvements

mtyaka:
## Description

Fixes these usability bugs:

- [TNL-6018](https://openedx.atlassian.net/browse/TNL-6018): Feedback not accessible to non-visual user.
- [TNL-6027](https://openedx.atlassian.net/browse/TNL-6027): Can select multiple draggables, but only last one is dropped.
- [TNL-6029](https://openedx.atlassian.net/browse/TNL-6029): role=button is misleading on draggables and dropzones.
- [TNL-6025](https://openedx.atlassian.net/browse/TNL-6025): Improper HTML elements used to define parts of problem.
- [Invalid aria-describedby value on old DnDv2 problems](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/111#issuecomment-267131707)

## JIRA Ticket

- [SOL-2156](https://openedx.atlassian.net/browse/SOL-2156)

## Sandbox URLs

Sandbox has been updated with 79ec8d7aab62b155d12bbce273648a0f25547179.

You might have to add a certificate security exception to be able to access the sandbox.

- LMS: https://pr14169.sandbox.stage.opencraft.hosting/
- Studio: https://studio-pr14169.sandbox.stage.opencraft.hosting/

### Feedback not accessible to non-visual user

1. Standard mode: When dropping items onto the board, the per-item feedback should be read out by the screen reader.
2. Assessment mode: When submitting the problem, all messages in the feedback area, and all messages in the popup feedback (when present) should be read out by the screen reader.

### Can select multiple draggables, but only last one is dropped

Without this patch it was possible to select multiple items in the bank using VoiceOver (and possibly other screen reader) commands like this:

1. Focus one of the items in the item bank, press the RET key to select it. This automatically moves focus to the first droppable zone.
2. Without dropping the selected item, move focus back to another item in the item bank **using the screenreader commands**. For example the VoiceOver activation key + the back arrow.
3. Press RET key. Two items will now be selected at the same time.

With this patch, the above is no longer possible - when you try to select the second item using the steps above, the first selected item gets deselected automatically.

### role=button is misleading on draggables and dropzones

Check that draggables and roles no longer contain the role="button" attribute.

### Improper HTML elements used to define parts of problem

Check that `<section>` elements are no longer used. `<div>` elements with `role="group"` are used where it makes sense instead.

## Author notes

Some parts of [TNL-6029](https://openedx.atlassian.net/browse/TNL-6029) were already taken care of in #110, so I did not duplicate them here.

## Reviewers

- [x] OpenCraft: @pomegranited 
- [x] TNL: @cahrens 
- [x] a11y: @cptvitamin


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=112"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -509,36 +514,38 @@ function DragAndDropTemplates(configuration) {
...
-                h('section.problem', [
+                h('div.problem', [
                     problemHeader,
                     h('p', {innerHTML: ctx.problem_html}),
                 ]),
-                h('section.drag-container', {}, [
+                h('div.drag-container', {}, [
                     h('div.item-bank', item_bank_properties, [
                         h('p', { className: 'zone-description sr' }, gettext('Item Bank')),
                         renderCollection(itemTemplate, items_in_bank, ctx),
```
pomegranited: Does it make more sense to put the `Item Bank` description in as an `aria-label` on the `div.item-bank`, instead of an sr-visible `p` tag?  Could use `role="group"` on that div too.

Then the `div.item-bank` would also match the `div.target` below.

pomegranited:
@mtyaka  Having some issues with the `SR.readTexts` content not being read by VoiceOver?

- [x] I tested this on the sandbox course:
* *OK* Verified that only a single item can be selected at a time using the VoiceOver selection keys.
* *FAIL?* VoiceOver reads item feedback when using keyboard to place items in correct and incorrect zones.
   In Standard mode: the overall feedback is read out, but the item feedback is never read.
   In Assessment mode: the item feedback is read out, but the overall feedback isn't read until I close the item feedback box.

   However, using the JS debugger, I can see that [`SR.readTexts(messages);`](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/112/files#diff-cef30416b48c3e5d4dc05c6922a2729bR877) is being called with the correct text, including both the overall and item feedback messages.  So I don't understand why it isn't being read out?
- [x] I read through the code:
* *OK* Verified that `role="button"` has been removed from all draggable items.
* *OK* Verified that learner-facing `section` elements have been replaced with `div role="group" aria-lable="group name"`.  
* *FAIL?* There's still a number of `section` tags in the Studio template [drag_and_drop_edit.html](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/mtyaka/a11y-interaction-issues/drag_and_drop_v2/templates/html/drag_and_drop_edit.html).  Would it make sense to fix them now, or is Studio out of scope for this work?
- [x] I checked for accessibility issues.  All manual testing was run using keyboard shortcuts and VoiceOver assist.
- [x] Includes documentation

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: Thanks for the suggestion, it definitely makes sense.

mtyaka:
Thanks for the review @pomegranited!

> FAIL? VoiceOver reads item feedback when using keyboard to place items in correct and incorrect zones. In Standard mode: the overall feedback is read out, but the item feedback is never read. In Assessment mode: the item feedback is read out, but the overall feedback isn't read until I close the item feedback box.

Weird that I got almost opposite behavior - in Standard mode per-item feedback was usually read correctly and in Assessment mode, overall feedback was working fine for me, but per-item feedback was not.
I made some tweaks to aria-relevant attributes and I switched to using `SR.readText` instead of `SR.readTexts` (concatenating the messages before passing them to `readText`). I also added a `SR.clear()` before invoking `SR.readText(...)`.
With these tweaks it still doesn't work correctly 100% of the time, but at least it now seems to work correctly _most_ of the time for me (tested on OS X El Capitan with VoiceOver on Safari).

> FAIL? There's still a number of section tags in the Studio template drag_and_drop_edit.html. Would it make sense to fix them now, or is Studio out of scope for this work?

Thanks, I missed those because at first glance it looked like they were being used correctly (used to delimit tabbed pages), but on second look, it turns out they were used to break down individual tabs into subsections. I fixed that so that `<section>` elements are still used in the studio edit template, but I think they are used correctly now :)

drag_and_drop_v2/public/js/drag_and_drop.js
pomegranited: Ah cool, thanks for pointing me towards that change.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -257,38 +257,6 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
-    var popupTemplate = function(ctx) {
```
pomegranited: Nice catch!

mtyaka:
Thanks @pomegranited!

> Also I would guess that it's better to use SR.readTexts than SR.readText in this case, since it seems designed to send multiple messages at once.

The problem with `readTexts` is that it would skip some messages when messages don't change since last time they are read. `readText` seem to force it to read all messages (kind of like `aria-atomic="true"`), although it does feel like a bit of a hack. I'll let upstream comment on this.

pomegranited:
@mtyaka Ah ok, I didn't see the subtlety of meaning there..  And yep, I agree that upstream is best qualified to answer here.  Thanks!

mtyaka:
@cahrens This is another batch of DnDv2 a11y updates that is now ready for your review.

drag_and_drop_v2/templates/html/drag_and_drop.html
```diff
@@ -1,4 +1,4 @@
 {% load i18n %}
-<section class="themed-xblock xblock--drag-and-drop">
+<div class="themed-xblock xblock--drag-and-drop">
     <i class="fa fa-spin fa-spinner initial-load-spinner"></i>{% trans "Loading drag and drop problem." %}
```
cahrens: The i tag, which is for italic, should not be used for icons. Please go through and change any i tags that are being used for icons.

tests/integration/test_base.py
```diff
@@ -47,7 +47,7 @@
 
 
 class BaseIntegrationTest(SeleniumBaseTest):
-    default_css_selector = 'section.themed-xblock.xblock--drag-and-drop'
+    default_css_selector = 'div.themed-xblock.xblock--drag-and-drop'
```
cahrens: Note that we prefer to just use class names for selectors instead of including the HTML tag type. It makes changes like you are doing now easier, and the tag type usually doesn't provide any additional value.

drag_and_drop_v2/templates/html/drag_and_drop_edit.html
```diff
@@ -4,10 +4,12 @@
...
     {{ js_templates|safe }}
 
-    <section class="drag-builder">
-        <div class="tab feedback-tab">
-
-            <section class="tab-content">
+    <div class="drag-builder">
+        <section class="tab feedback-tab">
+            <header class="tab-header">
+                <h3>{% trans "Basic Settings" %}</h3>
```
cahrens: Unfortunately the heading levels in Studio are an absolute mess. I think ultimately this will need be an h4 (with the title of the editing dialog being an h3), but for now you are consistent with what is happening for other component editors.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -239,10 +244,10 @@ function DragAndDropTemplates(configuration) {
         });
 
         return (
-            h('section.feedback', {}, [
+            h('div.feedback', {attributes: {'role': 'group', 'aria-label': gettext('Feedback')}}, [
                 h(
                     "div.feedback-content",
-                    { attributes: { 'aria-live': 'polite' } },
+                    {attributes: {'aria-live': 'polite', 'aria-atomic': 'true', 'aria-relevant': 'all'}},
```
cahrens: If the aria-live functionality on this feedback-content div is not working, we should remove it. As it is, if it should happen to work wouldn't the user be inundated with double messaging due to the usage of SR.readText?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -509,36 +488,38 @@ function DragAndDropTemplates(configuration) {
...
                 'dropzone': 'move',
-                'aria-dropeffect': 'move',
-                'role': 'button'
+                'aria-dropeffect': 'move'
             };
         }
+        var main_element_properties = {
+            attributes: {
+                'role': 'group',
+                'aria-label': gettext('Drag and Drop Problem')
```
cahrens: Did @cptvitamin request this change? Instead of setting a hard-coded label, I would use aria-labelledby and reference the problem-title.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -826,14 +811,48 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+                        messages.push(gettext("Some of your answers were not correct."))
+                    }
+                    messages = messages.concat(
+                        gettext("Hints:"),
+                        pluckMessages(state.feedback)
+                    );
+                }
+            }
+
+            SR.clear();
```
cahrens: The clear is unnecessary-- on readText the whole content of the div is replaced.

cahrens:
I'm done with my first pass.

drag_and_drop_v2/templates/html/drag_and_drop.html
mtyaka: Thanks, I found two improper usages of `<i>` tag and replaced them with `<span>`s.

tests/integration/test_interaction.py
```diff
@@ -295,6 +345,22 @@ def test_grade_display(self):
         # After placing all items, we get the full score.
         self.assertEqual(progress.text, '1/1 point (ungraded)')
 
+    @data(Keys.RETURN, Keys.SPACE, Keys.CONTROL+'m', Keys.COMMAND+'m')
```
mtyaka: Will keep that in mind when merging/rebasing.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: That's a very good point. It looks like the `aria-live` regions were conflicting with `SR.readText` because after removing `aria-live` attributes, VoiceOver seems to read feedback messages correctly most of the time.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: The reason I used a hardcoded label was because the title is optional, so a hardcoded label seemed easier, but now that I looked at it again, it was pretty easy to change the code so that it will use `aria-labelledby` referencing the title when present, otherwise it will fall back to hardcoded `aria-label`; thanks for prompting me.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: Thanks, I removed `SR.clear()`.

mtyaka:
@cahrens I addressed your comments and improved the code a bit. When testing with VoiceOver the feedback messages are now read correctly most of the time (although it still misses to read the feedback occasionally).

drag_and_drop_v2/templates/html/drag_and_drop.html
```diff
@@ -1,4 +1,4 @@
 {% load i18n %}
-<section class="themed-xblock xblock--drag-and-drop">
-    <i class="fa fa-spin fa-spinner initial-load-spinner"></i>{% trans "Loading drag and drop problem." %}
-</section>
+<div class="themed-xblock xblock--drag-and-drop">
+    <span class="fa fa-spin fa-spinner initial-load-spinner"></span>{% trans "Loading drag and drop problem." %}
```
cahrens: The icons need to be marked with aria-hidden. See the HTML markup at http://ux.edx.org/design_elements/icons/.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -8,7 +8,7 @@ function DragAndDropTemplates(configuration) {
         }
         return (
             h("div.spinner-wrapper", {key: item.value + '-spinner'}, [
-                h("i.fa.fa-spin.fa-spinner")
+                h("span.fa.fa-spin.fa-spinner")
```
cahrens: Need also to add aria-hidden (and make sure there is visible or sr text). See markup at http://ux.edx.org/design_elements/icons/.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -387,12 +387,7 @@ function DragAndDropTemplates(configuration) {
             popupSelector,
             {
                 style: {display: have_messages ? 'block' : 'none'},
-                attributes: {
-                    "tabindex": "-1",
-                    'aria-live': 'polite',
-                    'aria-atomic': 'true',
-                    'aria-relevant': 'all',
-                }
+                attributes: {"tabindex": "-1"}
```
cahrens: If you remove tabindex = -1, what is the impact? I'm just trying to understand why it is necessary.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -311,27 +284,33 @@ function DragAndDropTemplates(configuration) {
...
-                  gettext("You have used {used} of {total} attempts.")
-                      .replace("{used}", ctx.attempts).replace("{total}", ctx.max_attempts)
-              )
-          ])
+            h("div.action-toolbar-item.submit-answer", {}, [
+                h(
+                    "button.btn-brand.submit-answer-button",
+                    submitButtonProperties,
+                    [
+                        (ctx.submit_spinner ? h("span.fa.fa-spin.fa-spinner") : null),
```
cahrens: When this is in the "spinning" state, is there any screen reader text to explain what is going on? And is the icon aria-hidden?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -826,14 +805,46 @@ function DragAndDropBlock(runtime, element, configuration) {
...
         var new_vdom = render(state);
         var patches = virtualDom.diff(__vdom, new_vdom);
         root = virtualDom.patch(root, patches);
         $root = $(root);
         __vdom = new_vdom;
     };
 
+    // Uses edX JS accessibility tools to read feedback messages when present.
+    var readScreenReaderMessages = function() {
+        if (window.SR && window.SR.readText) {
```
cahrens: A thought about this-- I think you mentioned that the drag and drop block is used outside of the LMS. In those environments, you wouldn't have any aria-live text for feedback, which would be an accessibility problem.

You might wish to consider adding your own "screen reader" region with the HTML of the drag and drop problem and send your feedback messages there instead of to the edX one.

cahrens:
I'm done with this pass.

drag_and_drop_v2/templates/html/drag_and_drop.html
mtyaka: Ooh, I missed that. Done now, thanks.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: Fixed.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: I don't understand either. It looks like it was added as part of a larger patch and might not have been intentional. I removed it and cannot see any issues.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: Added aria-hidden and SR text; thanks.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: Thank you, that is a very good idea. I implemented the suggested approach, which also allowed me to add `aria-atomic="true"` to the live element to ensure feedback messages are read in entirety each time, even if only part of the feedback changes. I did some testing with VoiceOver and it looks like feedback messages are now always announced correctly.

mtyaka:
@cahrens This is ready for another look.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -299,13 +300,22 @@ function DragAndDropTemplates(configuration) {
             attemptsUsedInfo = h("span.attempts-used", {id: attemptsUsedId}, attemptsUsedText);
         }
 
+        var submitSpinner = null;
+        if (ctx.submit_spinner) {
+            submitSpinner = h('span', [
+                h('span.fa.fa-spin.fa-spinner', {attributes: {'aria-hidden': true}}),
+                h('span.sr', gettext('Saving'))
```
cahrens: Should it be "Saving" or "Submitting"?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -8,7 +8,8 @@ function DragAndDropTemplates(configuration) {
         }
         return (
             h("div.spinner-wrapper", {key: item.value + '-spinner'}, [
-                h("span.fa.fa-spin.fa-spinner")
+                h("span.fa.fa-spin.fa-spinner", {attributes: {'aria-hidden': true}}),
+                h("span.sr", gettext('Saving'))
```
cahrens: Is "Saving" the correct text here?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -299,13 +300,22 @@ function DragAndDropTemplates(configuration) {
...
         return (
             h("div.action-toolbar-item.submit-answer", {}, [
                 h(
                     "button.btn-brand.submit-answer-button",
                     submitButtonProperties,
                     [
-                        (ctx.submit_spinner ? h("span.fa.fa-spin.fa-spinner") : null),
+                        submitSpinner,
+                        ' ',  // whitespace between spinner icon and text
                         gettext("Submit")
```
cahrens: If this has visible text "Submit", it doesn't also need sr text. When is this button visible?

cahrens:
@mtyaka is the sandbox updated? I'm still seeing the tabindex of -1 on popups. And I wanted to try out the new sr region.

cahrens:
This looks generally good-- I had some questions about the text on the buttons with spinners (some of the comments are showing as collapsed; please expand them). I definitely like you using your own sr-region since this problem type is outside of edx-platform.

I'm actually out sick today, and I am on vacation/holiday then until 1/3. I don't know what @cptvitamin's holiday schedule is, but I'd recommend that you create a single sandbox with all 3 PRs on it for him to review. It would be easier than him reviewing 3 different sandboxes.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: "Submitting" is probably better... I changed the text; thanks.

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: Changed to "Submitting".

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: In standard mode the button is not used at all, but it is *always* visible in assessment mode. The spinner icon is only visible while the submit XHR is in progress after the user clicks the button, though. While the XHR is in progress and the spinner visible, the button is also disabled (see screenshots). The spinner gets hidden as soon as the XHR is complete.

screen shot 2016-12-23 at 08 37 19
screen shot 2016-12-23 at 08 37 53

The "Submit" text is always visible, but the spinner icon indicates that submission is in progress, so the icon technically still needs sr text. The sr text currently says "Submitting", which is fine on its own, but screen readers are going to read "Submitting submit", which is not ideal. The XHR typically completes pretty quickly, so it is not very likely users will notice that text, but I would appreciate better ideas as to what sr text to put next to the spinner.

mtyaka:
@cahrens Thanks for the quick response.

> Is the sandbox updated? I'm still seeing the tabindex of -1 on popups. And I wanted to try out the new sr region.

Ah sorry, I forgot to update the sandbox. I'm updating it now.

> I don't know what @cptvitamin's holiday schedule is, but I'd recommend that you create a single sandbox with all 3 PRs on it for him to review. It would be easier than him reviewing 3 different sandboxes.

I think that during the last round of updates to DnDv2, we merged individual PRs first, and then Mark did a final review of the end result and created new tickets for issues he found. It would probably make sense to do the same here... I'll discuss that with others.

drag_and_drop_v2/public/js/drag_and_drop.js
cahrens: I'm not sure any sr text is necessary here since the button is disabled and the text "Submit" is present. Generally one only needs screen-reader text when there is no displayed text.

I'd suggest you just run this case by @cptvitamin and see what he says.

cahrens:
👍 Just please follow up with @cptvitamin about the "Submitting" button state.

cptvitamin:
@mtyaka all of this looks great with one exception. I am reversing my suggestion to remove the role="button" from the draggables and drop targets based on guidance that has just been published:
http://www.ssbbartgroup.com/blog/differences-aria-1-0-1-1-deprecations-additions/

We knew that the W3C aria WG was planning to deprecate aria-grabbed and aria-dropeffect.  However, as pointed out in the article, that decision was based on (future) native API support. It's kinda like how Trump wants to repeal Obamacare without having a replacement available yet.  Until we get native API support, guidance is to continue supporting/using the deprecated attributes.  So lets follow the guidance here: http://whatsock.com/training/#hd13  In summary:

* add role=button back to draggables and dropzones
* add an aria-grabbed attribute to all draggables, defaulting to false, toggling to true when the item is grabbed (continue to change the label to also indicate it is grabbed which was done in another PR)
* add aria-dropeffect="move" to all dropzones.

Then we can call this done.  Sorry for the flop-flop on this, but the guidance in the above reference article was published this week...

mtyaka:
Rebased from 79ec8d7aab62b155d12bbce273648a0f25547179 and added `role="button"` back.

PR #113, opened 2016-12-30
added translation files and transifex config

ziafazal:
This PR added translation files download from transifex and transifex config to support WL-580.
@douglashall could you please review it?


douglashall:
👍

PR #118, opened 2017-03-03
implements scorablexblockmixin and moves drag and drop v2 to use raw grades.

sanfordstudent:
In the course of [implementing rescore ](https://openedx.atlassian.net/browse/TNL-6593)for the drag-and-drop v2 XBlock, I uncovered [some bugs](https://openedx.atlassian.net/browse/TNL-6638) in the XBlock's handling of grades. This PR implements the abstract methods in the scorable XBlock mixin and changes the internal handling of grades to deal solely in raw, unweighted values until they need to be surfaced to learners.

I will remove my edit to requirements.txt when Cliff's work lands in the XBlock repo.

Reviewer: @jcdyer

FYI @staubina (for timing) and @bradenmacdonald (since OpenCraft built this block)

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -14,6 +14,7 @@ from xblock.core import XBlock
 from xblock.exceptions import JsonHandlerError
 from xblock.fields import Scope, String, Dict, Float, Boolean, Integer
 from xblock.fragment import Fragment
+from xblock.scorable import ScorableXBlockMixin, Score
 from xblockutils.resources import ResourceLoader
 from xblockutils.settings import XBlockWithSettingsMixin, ThemableXBlockMixin
 
@@ -31,7 +32,12 @@ logger = logging.getLogger(__name__)
 
 @XBlock.wants('settings')
 @XBlock.needs('i18n')
-class DragAndDropBlock(XBlock, XBlockWithSettingsMixin, ThemableXBlockMixin):
+class DragAndDropBlock(
+    XBlock,
+    XBlockWithSettingsMixin,
+    ThemableXBlockMixin,
+    ScorableXBlockMixin
+):
     """
     XBlock that implements a friendly Drag-and-Drop problem
     """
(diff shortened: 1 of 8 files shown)


sanfordstudent:
Hm... that integration failure on the push job didn't happen on the last run, before I squashed. I'll investigate it but please feel free to review in the meantime.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 git+https://github.com/edx/xblock-utils.git@v1.0.2#egg=xblock-utils==1.0.2
+git+https://github.com/edx/XBlock.git@cdyer/scorable-mixin
```
bradenmacdonald: Is this just a temporary thing? Because normally we'd want to stick with the pattern of letting `xblock-utils` specify the version of `XBlock` that is being used here (for tests). So it would be best to bump the dependency in `xblock-utils` then bump the `xblock-utils` version here.

sanfordstudent: Yes, this is just temporary while some work @jcdyer is doing is still in flight. I want this dependency pulled in for the sake of running tests but I definitely would not merge this file!

sanfordstudent: PR to upgrade xblock-utils: https://github.com/edx/xblock-utils/pull/42

Not sure why I just got all those geckodriver-related failures, though...

bradenmacdonald: @sanfordstudent Try what I mentioned here: https://github.com/edx/xblock-utils/pull/41#issuecomment-285172023 perhaps?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -179,6 +184,42 @@ def max_score(self):  # pylint: disable=no-self-use
         """
         return 1
 
+    def get_score(self):
+        """
+        Return the problem's current score as raw values.
+        """
+        return Score(self.grade, self.max_score())
```
jcdyer: Is there going to be a problem when existing blocks (already answered problems) that are storing a weighted grade in self.grade get processed using this code, and compare the weighted grade to max_score()?

requirements.txt
```diff
@@ -1,2 +1,2 @@
-git+https://github.com/edx/xblock-utils.git@v1.0.2#egg=xblock-utils==1.0.2
+git+https://github.com/edx/xblock-utils.git@v1.0.4#egg=xblock-utils==1.0.4
```
sanfordstudent: So instead of the update I made here: https://github.com/edx/xblock-utils/commit/890bac00bca0df85c92a12c198dbb071b89e7b8d

would you have changed the `==` to a `>=`?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -179,6 +190,44 @@ def max_score(self):  # pylint: disable=no-self-use
         """
         return 1
 
+    def get_score(self):
+        """
+        Return the problem's current score as raw values.
+        """
+        if not self._get_raw_earned_if_set():
```
nasthagiri: Shouldn't this be `if self._get_raw_earned_if_set() is None`?  Otherwise, this will continue to recompute even if self.raw_earned was already computed as 0.

sanfordstudent: yeah...

PR #127, opened 2017-03-21
view should always show a weighted grade

sanfordstudent:
I found this bug while manually testing. Not sure how I missed it when I was testing manually before.

Reviewer: @jcdyer 
FYI: @staubina there's another dnd change coming down the line.

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -477,11 +477,10 @@ class DragAndDropBlock(
             misplaced_items.append(self._get_item_definition(int(item_id)))
 
         feedback_msgs = [FeedbackMessage(item['feedback']['incorrect'], None) for item in misplaced_items]
-
         return {
             'correct': correct,
             'attempts': self.attempts,
-            'grade': self._get_raw_earned_if_set(),
+            'grade': self._get_weighted_earned_if_set(),
             'misplaced_items': list(misplaced_ids),
             'feedback': self._present_feedback(feedback_msgs),
             'overall_feedback': self._present_feedback(overall_feedback_msgs)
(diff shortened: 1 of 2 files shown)


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -477,11 +477,13 @@ def do_attempt(self, data, suffix=''):
             misplaced_items.append(self._get_item_definition(int(item_id)))
 
         feedback_msgs = [FeedbackMessage(item['feedback']['incorrect'], None) for item in misplaced_items]
-
+        grade = self._get_raw_earned_if_set()
+        if grade is not None:
+            grade = self.weighted_grade()
```
sanfordstudent: I should just condense all of this into what it is: `_get_weighted_earned_if_set`.

sanfordstudent: And the idea with this code is to first check if there is a grade at all, then return None if there is no grade, but `weighted_grade()` if there is. A helper function will definitely clean this up.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -952,6 +943,16 @@ def _get_raw_earned_if_set(self):
         else:
             return None
 
+    def _get_weighted_earned_if_set(self):
+        """
+        Returns student's grade if already explicitly set, otherwise returns None,
+        with the problem weight applied.
+        """
```
jcdyer: Nit. Syntax is a little baroque here.  Consider `"Returns student's grade with the problem weight applied if set, otherwise \`None\`"`.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -946,6 +943,16 @@ def _get_raw_earned_if_set(self):
...
```
jcdyer: Nit: Syntax is a little baroque here. Consider "Returns student's grade with the problem weight applied if set, otherwise None".

PR #129, opened 2017-04-03
Sstudent/tnl 6753

sanfordstudent:
Updates drag-and-drop-v2 to respect the concept of moving an item back to the bank in assessment mode.

For [TNL-6753](https://openedx.atlassian.net/browse/TNL-6753)

Reviewer:

- [ ] @jcdyer

FYI: @bradenmacdonald @apyersedx

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*- # pylint: disable=C0302
 #
 """ Drag and Drop v2 XBlock """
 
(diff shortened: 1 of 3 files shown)


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*- # pylint: disable=C0302
```
sanfordstudent: Before anyone says anything: I know.

jcdyer: Are you going to make it better, or just keep knowing?

sanfordstudent: I'm not refactoring the drag and drop block class in this ticket, but I will use the human-readable identifier.

PR #131, opened 2017-05-18
(WIP) WL-415 | marking HTML strings.

hasnain-naveed:


xitij2000:
@hasnain-naveed  This is no longer needed :-)

PR #132, opened 2017-09-20
[MCKIN-5772] Adjust feedback popup styles

cgopalan:
Change the feedback popup styles for mobile.

**JIRA tickets**: Implements SOL-5, SOL-6.

**Dependencies**: None

**Screenshots**:
These are the new screenshots that display with the changes in this PR.

screen shot 2017-09-22 at 3 11 45 am
screen shot 2017-09-22 at 3 12 18 am

**Merge deadline**: ASAP.

**Testing instructions**:

1. Add the Drag and Drop demo course to your courses. Alternatively, you can just add a Drag and Drop component to your existing course (for eg Problem Builder demo course)
2. Go to the  "Multiple Blocks Test" unit in the DnD demo or the PB demo course and drag and drop blocks to the target zone.

1. I have changed the current CSS just so that I can get feedback about the changes. Still have to add conditions to the code to have the changes appear only for mobile.
2. Looking for feedback on the right color codes, icons, etc and more since I have used some defaults just to get the code working at first.

**Reviewers**
- [ ] @mtyaka 
- [ ] @bradenmacdonald

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -112,6 +112,19 @@
     z-index: 10 !important;
 }
 
+@media screen and (max-width: 480px) {
+   /* Horizontal scroll for item bank on mobile */
+    .xblock--drag-and-drop .item-bank {
+        -ms-flex-flow: row nowrap;
+        flex-flow: row nowrap;
+        overflow-x: auto;
+    }
+
+    .xblock--drag-and-drop .drag-container .item-bank .option {
+        flex-shrink: 0;
+    }
+}
+
 .xblock--drag-and-drop .drag-container .option.specified-width img {
     width: 100%;  /* If the image is smaller than the specified width, make it larger */
 }
@@ -405,7 +418,7 @@
     max-width: 100%;
 }
 
-.xblock--drag-and-drop .popup .close-feedback-popup-button {
+.xblock--drag-and-drop .popup .close-feedback-popup-desktop-button {
     cursor: pointer;
     margin-top: 8px;
     color: #ffffff;
@@ -413,21 +426,109 @@
     font-size: 18pt;
 }
 
-.ltr .xblock--drag-and-drop .popup .close-feedback-popup-button {
+.xblock--drag-and-drop .popup .close-feedback-popup-mobile-button {
+    display: none;
+}
+
+.xblock--drag-and-drop .popup .popup-header-icon {
+    display: none;
+}
+
+.xblock--drag-and-drop .popup .popup-header-text {
+    display: none;
+}
+
+.xblock--drag-and-drop .popup .popup-header-icon.popup-header-icon-incorrect {
+    color: #ff143e;
+}
+
+.ltr .xblock--drag-and-drop .popup .close-feedback-popup-desktop-button {
     float: right;
     margin-right: 8px;
     margin-left: 20px;
 }
 
-.rtl .xblock--drag-and-drop .popup .close-feedback-popup-button {
+.rtl .xblock--drag-and-drop .popup .close-feedback-popup-desktop-button {
     float: left;
     margin-right: 20px;
     margin-left: 8px;
 }
 
-.xblock--drag-and-drop .popup .close-feedback-popup-button:focus {
+.xblock--drag-and-drop .popup .close-feedback-popup-desktop-button:focus {
     outline: 2px dotted white;
 }
+
+@media screen and (max-width: 480px) {
+    .xblock--drag-and-drop .popup .popup-content p {
+        text-align: center;
+    }
+
+    .xblock--drag-and-drop .popup {
+        background-color: #fff;
+        border-radius: 10px;
+        text-align: center;
+        box-shadow: 0 0 100px rgba(0,0,0,0.4);
+        border: none;
+        max-height: 90vh;
+    }
+
+    .xblock--drag-and-drop .popup .close-feedback-popup-desktop-button {
+        display: none;
+    }
+
+    .xblock--drag-and-drop .popup.popup-incorrect {
+        background-color: #fff;
+    }
+
+    .xblock--drag-and-drop .popup .popup-header-icon {
+        display: inline-block;
+        margin-top: 10px;
+        color: #629a2c;
+        font-size: 20pt;
+    }
+
+    .xblock--drag-and-drop .popup .popup-header-icon.popup-header-icon-incorrect {
+        color: #ff143e;
+    }
+
+    .xblock--drag-and-drop .popup .popup-header-text {
+        display: block;
+        padding: 2px 0 8px;
+        font-weight: 600;
+    }
+
+    .xblock--drag-and-drop .popup .popup-content {
+        color: #629a2c;
+        margin: 10px 0px 0px 0px;
+        font-size: 14px;
+        background-color: #f4ffe7;
+        padding: 20px;
+        border-color: #eee;
+        border-style: solid;
+        border-width: 1px 0 1px 0;
+    }
+
+    .xblock--drag-and-drop .popup .popup-content ul {
+        list-style-position: inside;
+        padding: 0;
+    }
+
+    .xblock--drag-and-drop .popup .popup-content.popup-content-incorrect {
+        color: #ff143e;
+        background-color: #ffe8ec;
+    }
+
+    .xblock--drag-and-drop .popup .close-feedback-popup-mobile-button {
+        display: inline;
+        cursor: pointer;
+        margin-top: 10px;
+        margin-bottom: 10px;
+        color: #3384ca;
+        font-size: 12pt;
+        padding: 10px;
+    }
+}
+
 
 /*** edX pattern library components ***/
 
@@ -583,19 +684,6 @@
     .rtl .xblock--drag-and-drop .actions-toolbar .action-toolbar-item.sidebar-buttons {
         float: left;
         padding-left: -5px;
-    }
-}
-
-@media screen and (max-width: 480px) {
-   /* Horizontal scroll for item bank on mobile */
-    .xblock--drag-and-drop .drag-container .item-bank .option {
-        flex-shrink: 0;
-    }
-
-    .xblock--drag-and-drop .item-bank {
-        -ms-flex-flow: row nowrap;
-        flex-flow: row nowrap;
-        overflow-x: auto;
     }
 }
 
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -427,11 +427,19 @@ function DragAndDropTemplates(configuration) {
                     return h("li", {innerHTML: message.message});
                 }))
             ];
-            popup_content = h("div.popup-content", {}, have_messages ? content_items : []);
+            popup_content = h(
+                ctx.last_action_correct ? "div.popup-content" : "div.popup-content.popup-content-incorrect",
+                {},
+                have_messages ? content_items : []
+            );
         } else {
-            popup_content = h("div.popup-content", {}, msgs.map(function(message) {
-                return h("p", {innerHTML: message.message});
-            }))
+            popup_content = h(
+                ctx.last_action_correct ? "div.popup-content" : "div.popup-content.popup-content-incorrect",
+                {},
+                msgs.map(function(message) {
+                    return h("p", {innerHTML: message.message});
+                })
+            );
         }
 
         return h(
@@ -441,7 +449,7 @@ function DragAndDropTemplates(configuration) {
             },
             [
                 h(
-                    'button.unbutton.close-feedback-popup-button',
+                    'button.unbutton.close-feedback-popup-button.close-feedback-popup-desktop-button',
                     {},
                     [
                         h(
@@ -460,7 +468,37 @@ function DragAndDropTemplates(configuration) {
                         )
                     ]
                 ),
-                popup_content
+                h(
+                    ctx.last_action_correct ? 'div.popup-header-icon' : 'div.popup-header-icon.popup-header-icon-incorrect',
+                    {},
+                    [
+                        h(
+                            ctx.last_action_correct ? 'span.icon.fa.fa-check-circle' : 'span.icon.fa.fa-exclamation-circle',
+                            {
+                                attributes: {
+                                    'aria-hidden': true
+                                }
+                            }
+                        )
+                    ]
+                ),
+                h(
+                    'div.popup-header-text',
+                    {},
+                    ctx.last_action_correct ? gettext("Correct") : gettext("Incorrect")
+                ),
+                popup_content,
+                h(
+                    'button.unbutton.close-feedback-popup-button.close-feedback-popup-mobile-button',
+                    {},
+                    [
+                        h(
+                            'span',
+                            {},
+                            gettext("Close")
+                        )
+                    ]
+                )
             ]
         )
     };
@@ -1092,7 +1130,7 @@ function DragAndDropBlock(runtime, element, configuration) {
     var focusItemFeedbackPopup = function() {
         var popup = $root.find('.item-feedback-popup');
         if (popup.length && popup.is(":visible")) {
-            popup.find('.close-feedback-popup-button').focus();
+            popup.find('.close-feedback-popup-button:visible').focus();
             return true;
         }
         return false;
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ install:
     - "pip install selenium==2.53.0"
     - "pip uninstall -y xblock-drag-and-drop-v2"
     - "python setup.py sdist"
-    - "pip install dist/xblock-drag-and-drop-v2-2.0.18.tar.gz"
+    - "pip install dist/xblock-drag-and-drop-v2-2.1.1.tar.gz"
 script:
     - pep8 drag_and_drop_v2 tests --max-line-length=120
     - pylint drag_and_drop_v2
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+Version 2.1.1 (2017-09-26)
+---------------------------
+
+* Enforce XBlock variable types (PR #104)
+* Improvements for mobile (PRs #132, #133, #134)
+
 Version 2.0.14 (2017-01-17)
 ---------------------------
 
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ def package_data(pkg, root_list):
 
 setup(
     name='xblock-drag-and-drop-v2',
-    version='2.0.18',
+    version='2.1.1',
     description='XBlock - Drag-and-Drop v2',
     packages=['drag_and_drop_v2'],
     install_requires=[


edx-webhook:
Thanks for the pull request, @cgopalan! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=132"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

cgopalan:
@bradenmacdonald @mtyaka please review when you get a chance. As explained above, I have not added the conditional code to do this only for mobile. I would like an initial review on my changes to see if I am going in the right direction.

clemente:
@cgopalan Your reviewers will give you a detailed review, but here are some notes/ideas from my side (without running it):
- looks nice!
- you removed `float:right` for the ✘ button but you didn't really hide it, it may still be positioned somewhere there. Better add `display:none` if you want to hide it. Or better, find the class that makes the button show (maybe `those button.unbutton.close-feedback-popup-button`…) and don't give that class
- nit: is the centering consistent? Either „Correct“ is too to the left, or „Close“ too to the right.
- the tone of blue for „Close“ is different
- the icon for „!“ seems smaller and darker than in the goal screenshot. Well, maybe this doesn't matter as long as it looks nice, which it does
- nit: the pink box still has a 1px white space (padding or margin) at the left and right side, though not very noticeable. Maybe it comes from the popup itself
- you don't need „a mobile version“, you need „a version for small screens“, and for that you need to wrap some of the code in a block that says „only for small screens (widths until XXXX px)“, where XXXX is some measure you see appropriate for this particular design (or maybe there's some standard in edx about what is small screen). You do this with media queries: https://www.w3schools.com/css/css3_mediaqueries_ex.asp

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -411,9 +411,10 @@ function DragAndDropTemplates(configuration) {
             ];
             popup_content = h("div.popup-content", {}, have_messages ? content_items : []);
         } else {
-            popup_content = h("div.popup-content", {}, msgs.map(function(message) {
-                return h("p", {innerHTML: message.message});
-            }))
+            popup_content = h(
+                              (ctx.last_action_correct ? "div.popup-content" : "div.popup-content.popup-content-incorrect"),   
+                              {}, msgs.map(function(message) {
+                                return h("span", {innerHTML: message.message});}))
```
UmanShahzad: Nit: this looks pretty hard to read, relative to what it was before.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -423,26 +424,44 @@ function DragAndDropTemplates(configuration) {
...
             [
                 h(
+                  (ctx.last_action_correct ? 'div.popup-header-icon' : 'div.popup-header-icon.popup-header-icon-incorrect'),
+                  {},
+                  [
+                    h(
+                      (ctx.last_action_correct ? 'span.icon.fa.fa-check-circle' : 'span.icon.fa.fa-exclamation-circle'),
+                      {
+                            attributes: {
+                            'aria-hidden': true
```
UmanShahzad: Nit: I *think* some spaces would be nice here.

mtyaka:
@clemente and @UmanShahzad Thank you for your comments!

@cgopalan This is looking good. In addition to addressing Daniel's and Uman's comments, please use these colors:

- Correct background color: `#f4ffe7`
- Correct text & icon color: `#629a2c`
- Incorrect background color: `#ffe8ec`
- Incorrect text & icon color: `#ff143e`
- The close button: `#3384ca`

For box shadow, try using something like: `0 0 100px rgba(0,0,0,0.4)`

cgopalan:
@clemente 
"you removed float:right for the ✘ button but you didn't really hide it, it may still be positioned somewhere there. Better add display:none if you want to hide it. Or better, find the class that makes the button show (maybe those button.unbutton.close-feedback-popup-button…) and don't give that class"
The x button is actually added in the [JS file](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/008b2dcc028e50e39b0a66f16f06b33a39a97fa3/drag_and_drop_v2/public/js/drag_and_drop.js#L453). I did remove that line and added the text instead. And then now the text is centered since I removed `float:right` and gave a `text-align:center` to the whole popup

drag_and_drop_v2/public/js/drag_and_drop.js
cgopalan: @UmanShahzad is it just a matter of formatting the code? How do you suggest I make it easier to read?

drag_and_drop_v2/public/js/drag_and_drop.js
cgopalan: @UmanShahzad I am not very familiar with how JS formatting should be done. What do you have in mind for this?

drag_and_drop_v2/public/js/drag_and_drop.js
UmanShahzad: Yeah, I just skimmed this, and wasn't taking a look at logic or functionality, but just the code readability. I'd suggest looking at pre-existing, surrounding code, including possibly some of the code in the deletion diff.

drag_and_drop_v2/public/js/drag_and_drop.js
UmanShahzad: Usually no matter where you are, you'd format things according to the surrounding code, because consistency tends to be preferred.

mtyaka:
@cgopalan 

> The x button is actually added in the JS file. I did remove that line and added the text instead. And then now the text is centered since I removed float:right and gave a text-align:center to the whole popup

This is fine while testing, but in the end, we want both the x button and the 'close' text button to be added to the DOM, and then one of them hidden via CSS depending on the size of the screen (via media queries).

cgopalan:
@clemente @UmanShahzad I think I have addressed all your comments.
@mtyaka I have replaced the colors with the ones you listed.
To classify it as "mobile" I have used a media query of `@media screen and (max-width: 1001px)`. Just used some arbitrary standard for now. Let me know if theres a more appropriate media query that we can use.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +448,62 @@
     outline: 2px dotted white;
 }
 
+@media screen and (max-width: 1001px) {
+    .xblock--drag-and-drop .popup {
+        background-color: rgba(255, 255, 255, 1);
```
mtyaka: I would prefer to use `#fff`.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +448,62 @@
...
+        text-align: center;
+        box-shadow: 0 0 100px rgba(0,0,0,0.4);
+    }
+
+    .xblock--drag-and-drop .popup .close-feedback-popup-button {
+        display: none;
+    }
+
+    .xblock--drag-and-drop .popup.popup-incorrect {
+        background-color: rgba(255, 255, 255, 1);
```
mtyaka: Please use `#fff` instead.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +448,62 @@
...
+    .xblock--drag-and-drop .popup .popup-header-text {
+        display: inherit;
+    }
+
+    .xblock--drag-and-drop .popup .popup-content {
+        color: #629a2c;
+        margin: 10px 0px 0px 0px;
+        font-size: 14px;
+        background-color: #f4ffe7;
+        padding: 20px;
```
mtyaka: The mockups have a borders on top and bottom of the content. Try adding these rules here:

```css
border-color: #eee;
border-style: solid;
border-width: 1px 0 1px 0;
```

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +448,62 @@
...
+        background-color: #ffe8ec;
+    }
+
+    .xblock--drag-and-drop .popup .close-feedback-popup-mobile-button {
+        display: inline;
+        cursor: pointer;
+        margin-top: 10px;
+        margin-bottom: 10px;
+        color: #3384ca;
+        font-size: 12pt;
```
mtyaka: Also add `padding: 10px;` - this increases the touch sensitive area so it's easier to hit on mobile and also looks nicer IMHO.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +448,62 @@
...
+
+    .xblock--drag-and-drop .popup.popup-incorrect {
+        background-color: rgba(255, 255, 255, 1);
+    }
+
+    .xblock--drag-and-drop .popup .popup-header-icon {
+        display: inline;
+        margin-top: 8px;
+        color: #629a2c;
+        font-size: 12pt;
```
mtyaka: Increase font size to 20pt to make it more similar to the mockups.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +448,62 @@
...
+    .xblock--drag-and-drop .popup .close-feedback-popup-button {
+        display: none;
+    }
+
+    .xblock--drag-and-drop .popup.popup-incorrect {
+        background-color: rgba(255, 255, 255, 1);
+    }
+
+    .xblock--drag-and-drop .popup .popup-header-icon {
+        display: inline;
```
mtyaka: Replace this with `display: inline-block;`, otherwise it ignores the margin.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +448,62 @@
...
+        margin-top: 8px;
```
mtyaka: Nit: I would increase the margin to 10px.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +448,62 @@
...
+        color: #629a2c;
+        font-size: 12pt;
+    }
+
+    .xblock--drag-and-drop .popup .popup-header-icon.popup-header-icon-incorrect {
+        color: #ff143e;
+    }
+
+    .xblock--drag-and-drop .popup .popup-header-text {
+        display: inherit;
```
mtyaka: Use `block` instead of `inherit`.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -412,6 +416,22 @@
...
+.xblock--drag-and-drop .popup .close-feedback-popup-mobile-button {
+    display: none;
+}
+
+.xblock--drag-and-drop .popup .popup-header-icon {
+    display: none;
+}
+
+.xblock--drag-and-drop .popup .popup-header-text {
+    display: none;
```
mtyaka: Please add these rules:

```css
padding: 2px 0 8px;
font-weight: 600;
```

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -594,9 +632,11 @@ function DragAndDropBlock(runtime, element, configuration) {
     DragAndDropBlock.STANDARD_MODE = 'standard';
     DragAndDropBlock.ASSESSMENT_MODE = 'assessment';
 
+    var MEDIA_QUERY = '(max-width: 1001px)';
```
mtyaka: I don't think we have to differentiate between mobile and desktop in JS at this point. I would prefer to only use media queries in CSS.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -594,9 +632,11 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+
     var Selector = {
         popup_box: '.popup',
-        close_button: '.popup .close-feedback-popup-button'
+        close_button: window.matchMedia( MEDIA_QUERY ).matches ? '.popup .close-feedback-popup-mobile-button' : '.popup .close-feedback-popup-button'
```
mtyaka: No need for media query, you can set the selector to match both buttons:

```
close_button: '.popup .close-feedback-popup-mobile-button, .popup .close-feedback-popup-button'
```

mtyaka: Or maybe event better to add the same class to both. Maybe replace the current usage of `close-feedback-popup-button` with `close-feedback-popop-desktop-button`, and then add `close-feedback-popup-button` to both buttons to be able to match them with a single selector.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -649,9 +689,18 @@ function DragAndDropBlock(runtime, element, configuration) {
 
             // Set up event handlers:
 
-            $element.on('click', '.item-feedback-popup .close-feedback-popup-button', closePopupEventHandler);
-            $element.on('keydown', '.item-feedback-popup .close-feedback-popup-button', closePopupKeydownHandler);
-            $element.on('keyup', '.item-feedback-popup .close-feedback-popup-button', preventFauxPopupCloseButtonClick);
+            // Depending on media setup the close button handlers.
+            if (window.matchMedia( MEDIA_QUERY ).matches) {
```
mtyaka: Here I would just bind events to both buttons - the one that is hidden is not going to receive them anyway.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -431,18 +434,53 @@ function DragAndDropTemplates(configuration) {
...
+                                    'aria-hidden': true
+                                }
+                            }
                         )
                     ]
+                  ),
+                h(
+                    'div.popup-header-text',
+                    {
+                        innerHTML: gettext(ctx.last_action_correct ? "Correct" : "Incorrect")
```
mtyaka: This needs to be rewritten as:

```
innerHTML: ctx.last_action_correct ? gettext("Correct") : gettext("Incorrect")
```

Otherwise gettext will not be able to parse the strings.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -431,18 +434,53 @@ function DragAndDropTemplates(configuration) {
...
                             {
                                 attributes: {
                                     'aria-hidden': true
                                 }
                             }
+                         )
+                    ]
+                 ),
+                h(
+                    ctx.last_action_correct ? 'div.popup-header-icon' : 'div.popup-header-icon.popup-header-icon-incorrect',
```
mtyaka: Nit: for consistency, it would be nice if we also used `popup-header-icon-correct`class.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -431,18 +434,53 @@ function DragAndDropTemplates(configuration) {
                             {
                                 innerHTML: gettext("Close")
                             }
-                        ),
+                         ),
```
mtyaka: Remove extra space.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -411,9 +411,12 @@ function DragAndDropTemplates(configuration) {
             ];
             popup_content = h("div.popup-content", {}, have_messages ? content_items : []);
         } else {
-            popup_content = h("div.popup-content", {}, msgs.map(function(message) {
-                return h("p", {innerHTML: message.message});
-            }))
+            popup_content = h(ctx.last_action_correct ? "div.popup-content" : "div.popup-content.popup-content-incorrect",
```
mtyaka: Nit: It would be nice if we also use `popup-content-correct` correct class, for consistency reasons.

mtyaka: Formatting is still weird here. Can you Format the code like you did below (i.e. with `h(` on it's own line, and nested lines indented 4 spaces).

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -411,9 +411,12 @@ function DragAndDropTemplates(configuration) {
...
+                              {},
+                              msgs.map(function(message) {
+                                return h("span", {innerHTML: message.message});
```
mtyaka: Why was this changed from `p` to `span`? `p` makes more sense semantically.

mtyaka:
@cgopalan Also the styles need a bit of tweaking for Assessment mode. If you edit a DnDv2 block in the Studio, there's a dropdown where you can chose between "Standard" and "Assessment" modes.

In assessment mode, the user can drag items to any zone, and the popup is only displayed when you click the "Submit" button. Multiple messages are combined into a single popup. See screenshot:

screen shot 2017-09-22 at 10 53 01

mtyaka:
@cgopalan It would be great if the popup was centered on the screen when running on mobile. Currently the position depends on the scroll position of the page (see screenshot). I think that in order to achieve that, we would have to wrap the popup in an extra div, which we could position with `position: fixed; left: 0; right: 0; top: 0; bottom: 0;` We could also apply a semi-transparent background to this div to act as an overlay.

It should then be possible to center the popup with a combination of `display: flex` some flex properties on the popup.

Feel free to skip this though if it's too hard. It could be an improvement for a separate task.

screen shot 2017-09-22 at 17 08 49

drag_and_drop_v2/public/js/drag_and_drop.js
cgopalan: `p` element seems to have a style defined saying `text-align:left`. So the popup text is always aligned left. I was not sure how to override the style for the `p` element just for this case.

cgopalan:
@mtyaka I have addressed all the comments, except for the last one about centering the popup. I would prefer to do it as a separate task.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -428,6 +446,66 @@
     outline: 2px dotted white;
 }
 
+@media screen and (max-width: 480px) {
+    .xblock--drag-and-drop .popup {
+        background-color: #fff;
+        border-radius: 10px;
+        text-align: center;
+        box-shadow: 0 0 100px rgba(0,0,0,0.4);
```
mtyaka: Sorry I missed this during previous pass, but can you also add this rule here: `border: none`. It removes a 1px wide white border which is a bit distracting.

mtyaka: Ans also: `max-height: 90vw;` to allow it to use more vertical space.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -409,11 +409,19 @@ function DragAndDropTemplates(configuration) {
                     return h("li", {innerHTML: message.message});
                 }))
             ];
-            popup_content = h("div.popup-content", {}, have_messages ? content_items : []);
+            popup_content = h(
+                                ctx.last_action_correct ? "div.popup-content" : "div.popup-content.popup-content-incorrect",
+                                {},
+                                have_messages ? content_items : []
+                            );
```
mtyaka: Please change indentation of these lines, so that they are indented like this:

```javascript
            popup_content = h(
                ctx.last_action_correct ? "div.popup-content" : "div.popup-content.popup-content-incorrect",
                {},
                have_messages ? content_items : []
            );
```

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -409,11 +409,19 @@ function DragAndDropTemplates(configuration) {
...
-            popup_content = h("div.popup-content", {}, msgs.map(function(message) {
-                return h("p", {innerHTML: message.message});
-            }))
+            popup_content = h(
+                                ctx.last_action_correct ? "div.popup-content" : "div.popup-content.popup-content-incorrect",
+                                {},
+                                msgs.map(function(message) {
+                                    return h("span", {innerHTML: message.message});
+                                })
+                            )
```
mtyaka: Please indent it like this:

```javascript
            popup_content = h(
                ctx.last_action_correct ? "div.popup-content" : "div.popup-content.popup-content-incorrect",
                {},
                msgs.map(function(message) {
                    return h("span", {innerHTML: message.message});
                })
            );
```

drag_and_drop_v2/public/js/drag_and_drop.js
mtyaka: You can override the styles of `p` elements inside the popup like this:

```css
@media screen and (max-width: 480px) {
    .xblock--drag-and-drop .popup .popup-content p {
        text-align: center;
    }
}
```

mtyaka:
@cgopalan I added some more comments inline. Please change `span` element back to `p` and override the text alignment like this:

```css
@media screen and (max-width: 480px) {
    .xblock--drag-and-drop .popup .popup-content p {
        text-align: center;
    }
}
```

There's another problem that I missed earlier: when the popup is displayed, its close button should automatically become focused. This works for desktop styles, but currently not for the new mobile styles. I think you should be able to fix that by changing [this line](https://github.com/open-craft/xblock-drag-and-drop-v2/blob/a0edcb1ca72ca9f70f38f08398ef09acb9c9acc6/drag_and_drop_v2/public/js/drag_and_drop.js#L1097) to:

```javascript
popup.find('.close-feedback-popup-button:visible, .close-feedback-popup-mobile-button:visible').focus();
```

cgopalan:
@mtyaka thanks for all the time you put into this and the extensive feedback. I will work on it after the meeting and finish it today.

cgopalan:
@mtyaka up for review again!

mtyaka:
Thanks @cgopalan! :+1:

I noticed some minor issues, some of them due to bad advice from me, so instead of going through another round, I fixed the issues in e86c6a5d6142f64da09f1e83ec53c0d4f8e1ec53.

- [x] I tested this: I verified that the popups look good and function correctly both in desktop and mobile mode.
- [x] I read through the code
- [ ] ~~Includes documentation~~ n/a

mtyaka:
Rebased from e86c6a5d6142f64da09f1e83ec53c0d4f8e1ec53.

PR #133, opened 2017-09-21
[MCKIN-5774] Wait 500 milliseconds before starting drag on touch.

mtyaka:
Wait 500 milliseconds before starting drag on touch.

This allows users to scroll the item bank with their finger without accidentally starting to drag items.

Drag events initiated via mouse still start without a delay, since scrolling works differently on desktops.

**JIRA tickets**: OC-3059, MCKIN-5774

**Dependencies**: None

**Testing instructions**:

1. Add the Drag and Drop demo course to your course in various configurations.
2. Test mouse dragging. It should behave just like before this patch.
3. Test touch dragging. Touch dragging should only start 500ms after you hold the finger on the item without moving. If you put your finger on the item, but move the finger before 500ms pass, the drag should not start and default browser behaviour should kick in (i.e. page scrolling).
4. Make sure to test in both Standard and Assessment modes. In assessment mode, items can be moved between zones after they are placed. They can also be dragged back to the item bank.

1. I found no way to implement this with jQuery UI draggable, so I had to re-implement drag and drop interactions, which in some ways makes things cleaner because it doesn't bypass virtualDOM.
2. In assessment mode, when you drag an item that was already placed in one of the zones, at the moment the item is picked up, the dimensions of the item may change (and are equal to item's dimensions in the bank). It would be possible to prevent that, but would take more time than I was able to spend on this ticket. It would be a nice future improvement.
3. The existing tests all pass, but I am not sure how to add tests for touch interactions, since FF driver for Selenium doesn't support them. Suggestions welcome.

**Reviewers**
- [ ] @bradmerlin


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=133"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1101,132 +1160,285 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+        };
+
+
+        // Touch devices emulate mousedown events after touchstart, which would cause our drag
+        // start event to be handled twice.
+        // One way to prevent that would be to call evt.preventDefault() in the touchstart handler,
+        // but that would also prevent scrolling, which we do not want.
+        // Instead of preventing the default, we use the handled_by_touch flag, which we set to true
+        // in the touchstart handler (and set it back to false after the event is processed).
+        // If the mousedown handler sees the flag set to true, it will simply ignore the event.
```
bradenmacdonald: @mtyaka Thanks for including these great comments :)

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -262,12 +262,9 @@ def student_view(self, context):
         fragment = Fragment()
         fragment.add_content(loader.render_template('/templates/html/drag_and_drop.html'))
         css_urls = (
-            'public/css/vendor/jquery-ui-1.10.4.custom.min.css',
-            'public/css/drag_and_drop.css'
+            'public/css/drag_and_drop.css',
         )
         js_urls = (
-            'public/js/vendor/jquery-ui-1.10.4.custom.min.js',
-            'public/js/vendor/jquery-ui-touch-punch-0.2.3.min.js',  # Makes it work on touch devices
```
bradenmacdonald: Wow, great that we can remove these dependencies.

bradenmacdonald:
Quite the diff! Awesome work @mtyaka :)

bradmerlin:
:+1:

- [x] I tested this: 
  - On Chrome I tested that drag and drop works the same with and without the patch (drag happens immediately)
  - On Chrome while mimicking an iPhone, and on an Android simulator running the edX app, I confirmed that drag and drop happens immediately without the patch, and requires a 500ms+ delay to initiate the drag with the patch, or normal scrolling behaviour occurs.
- [x] I read through the code
- [ ] ~I checked for accessibility issues~ N/A
- [x] Includes documentation – code is well-documented

One concern that I have is for this test case:

> If you put your finger on the item, but move the finger before 500ms pass, the drag should not start and default browser behaviour should kick in (i.e. page scrolling).

The dotted border that is applied on `:hover` / `:focus` (e.g. on desktop, to indicate that an item can be dragged) is applied immediately on first touch on mobile and then after 500ms the item becomes transparent and can be dragged. The actual behaviour works as expected, but the visual change of the item before drag behaviour is applied might be a bit confusing. Or is this intentional?

mtyaka:
@bradmerlin Thanks for the review!

> The dotted border that is applied on :hover / :focus (e.g. on desktop, to indicate that an item can be dragged) is applied immediately on first touch on mobile and then after 500ms the item becomes transparent and can be dragged. The actual behaviour works as expected, but the visual change of the item before drag behaviour is applied might be a bit confusing. Or is this intentional?

It's not intentional and i has been bothering me as well, so I did some more research on how to avoid that, and learned about a `(hover: none)` media query, which applies on devices that don't support native hover. I used this to fix the annoying outline on first tap in my last commit, thanks for pointing it out!

PR #135, opened 2017-09-26
When running on mobile, make the target image scroll in the x direction

cgopalan:
On desktop the target image automatically shrinks/grows with its container. On mobile this often is not convenient because the image can shrink too much. We want to restrict the minimum width of the image on mobile and make the image scrollable in the x direction.

**Dependencies**: None

**Screenshots**: Here are the before and after screenshots for the different demo apps

Buckets:

Before:
buckets_orig

After:
buckets

Maps:

Before:
maps_orig

After:
maps

Multiple blocks:

Before:
blocks_orig

After:
multipleblocks

**Merge deadline**: ASAP.

**Testing instructions**:
1. Add a drag and drop demo unit to the course.
2. On a mobile device (less than 480px wide) check that the target image does not shrink and can be scrolled horizontally.

The testing was mostly done using firefox responsive design mode, and one LG android phone.

**Reviewers**
- [ ] @bradenmacdonald

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -249,6 +249,23 @@
     height: auto;
 }
 
+@media screen and (max-width: 480px) {
+    .xblock--drag-and-drop .target {
+        overflow: auto;
+        display: block;
+    }
+
+    .xblock--drag-and-drop .target-img {
+        max-width: none;
+        width: auto;
+    }
+
+    .xblock--drag-and-drop .target-img-wrapper {
+        display: inline-block;
+        position: relative;
+    }
+}
+
 .xblock--drag-and-drop .zone {
     position: absolute;
 
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -3,6 +3,18 @@
     max-width: 770px;
     margin: 0;
     padding: 0;
+    position: relative;
+}
+
+.xblock--drag-and-drop .resize-detector {
+    position: absolute;
+    top: 0;
+    left: 0;
+    height: 100%;
+    width: 100%;
+    overflow: hidden;
+    pointer-events: none;
+    z-index: -1;
 }
 
 .xblock--drag-and-drop .initial-load-spinner {
@@ -64,6 +76,7 @@
 /* drag-container holds the .items and the .target image */
 
 .xblock--drag-and-drop .drag-container {
+    box-sizing: border-box;
     width: auto;
     padding: 1%;
     background-color: #ebf0f2;
@@ -122,6 +135,8 @@
 
     .xblock--drag-and-drop .drag-container .item-bank .option {
         flex-shrink: 0;
+        flex-grow: 0;
+        max-width: 80%;
     }
 }
 
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -2,6 +2,10 @@ function DragAndDropTemplates(configuration) {
     "use strict";
     var h = virtualDom.h;
 
+    var isMobileScreen = function() {
+        return window.matchMedia('screen and (max-width: 480px)').matches;
+    };
+
     var itemSpinnerTemplate = function(item) {
         if (!item.xhr_active) {
             return null;
@@ -33,12 +37,12 @@ function DragAndDropTemplates(configuration) {
         if (item.widthPercent) {
             // The item bank container is often wider than the background image, and the
             // widthPercent is specified relative to the background image so we have to
-            // convert it to pixels. But if the browser window / mobile screen is not as
-            // wide as the image, then the background image will be scaled down and this
-            // pixel value would be too large, so we also specify it as a max-width
-            // percentage.
+            // convert it to pixels. But if the browser window is not as wide as the image,
+            // then the background image will be scaled down and this pixel value would be too large,
+            // so we also specify it as a max-width percentage.
+            // On mobile, the image is never scaled down, so we don't specify the max-width.
             style.width = (item.widthPercent / 100 * ctx.bg_image_width) + "px";
-            style.maxWidth = item.widthPercent + "%";
+            style.maxWidth = isMobileScreen() ? 'none' : item.widthPercent + "%";
         }
         return style;
     };
@@ -627,9 +631,21 @@ function DragAndDropTemplates(configuration) {
           }
         });
         bank_children = bank_children.concat(renderCollection(itemPlaceholderTemplate, items_placed, ctx));
-
+        var drag_container_style = {};
+        var target_img_style = {};
+        // If drag_container_max_width is null, we are going to render the container width after this render.
+        // To be able to accurately measure the natural container width, we have to set max-width of the target
+        // image to 100%, so that it doesn't expand the container.
+        if (ctx.drag_container_max_width === null) {
+            target_img_style.maxWidth = '100%';
+        } else {
+            drag_container_style.maxWidth = ctx.drag_container_max_width + 'px';
+        }
         return (
             h('div.themed-xblock.xblock--drag-and-drop', main_element_properties, [
+                h('object.resize-detector', {
+                    attributes: {type: 'text/html', tabindex: -1, data: 'about:blank'}
+                }),
                 problemTitle,
                 problemProgress,
                 h('div', [forwardKeyboardHelpButtonTemplate(ctx)]),
@@ -637,12 +653,16 @@ function DragAndDropTemplates(configuration) {
                     problemHeader,
                     h('p', {innerHTML: ctx.problem_html}),
                 ]),
-                h('div.drag-container', {}, [
+                h('div.drag-container', {style: drag_container_style}, [
                     h('div.item-bank', item_bank_properties, bank_children),
                     h('div.target', {attributes: {'role': 'group', 'arial-label': gettext('Drop Targets')}}, [
                         itemFeedbackPopupTemplate(ctx),
                         h('div.target-img-wrapper', [
-                            h('img.target-img', {src: ctx.target_img_src, alt: ctx.target_img_description}),
+                            h('img.target-img', {
+                                src: ctx.target_img_src,
+                                alt: ctx.target_img_description,
+                                style: target_img_style
+                            }),
                             renderCollection(zoneTemplate, ctx.zones, ctx)
                         ]),
                     ]),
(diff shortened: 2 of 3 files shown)


edx-webhook:
Thanks for the pull request, @cgopalan! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=135"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

cgopalan:
@mtyaka I am still researching why 4 tests are failing, but the code seems to be working fine to make the target image scroll. I tested it with "bucket demo", "canada demo", "multiple blocks test" and "rgb demo" sections in the drag and drop demo course.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -237,15 +237,35 @@
...
+    }
+
+    .xblock--drag-and-drop .target-img {
+        max-width: none;
+    }
+}
+
+
+@media screen and (min-width: 481px) {
+    /* Style for non-mobile */
```
mtyaka: @cgopalan Style for non mobile should just go under default styles (no need to wrap into `(min-width: 481px)` media query). You already reset `max-width` to `none` in mobile-only styles. You can also unset the default `width: 100%` using `width: auto` in mobile styles.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -237,15 +237,35 @@
...
+
 .xblock--drag-and-drop .target-img-wrapper {
     /* This element is required for Firefox due to https://bugzilla.mozilla.org/show_bug.cgi?id=975632 */
     display: table-row;
 }
 
 .xblock--drag-and-drop .target-img {
     display: table-cell;
-    width: 100%;
-    max-width: 100%;
```
mtyaka: As noted above, `width` and `max-width` should stay here under default styles, but should be overridden under mobile specific styles.

drag_and_drop_v2/public/css/drag_and_drop.css
cgopalan: @mtyaka I tried this, but if I use `width : auto` and remove the non-mobile style, the target image behaves like before. This is because for some reason the mobile css does not override the non-mobile css. See screenshot. Any idea why?

screen shot 2017-09-28 at 11 43 17 am

mtyaka: @cgopalan I think it's because the mobile styles are defined *above* the default styles. If you move the mobile styles below, it should override correctly.

cgopalan: @mtyaka please ignore the above comment. It turned out I was defining the mobile style before the default style, so the mobile style always got overriden. Its working now.

cgopalan: @mtyaka I think we both typed the responses at the same time! Thanks for confirming.

bradenmacdonald:
@sanfordstudent FYI, we also have this PR in progress.

@cgopalan Would you be able to add before-and-after screenshots?

cgopalan:
@bradenmacdonald sure!

cgopalan:
Squashed from c1f1686

bradenmacdonald:
@cgopalan To fix the issue with zone sizing on mobile that you're having, you need this additional rule:

```css
@media screen and (max-width: 480px) {
    .xblock--drag-and-drop .target-img-wrapper {
        position: relative;
        display: inline-block;
    }
}
```

bradenmacdonald:
@cgopalan @mtyaka @bradmerlin Did you test the latest changes to this XBlock on a `master` devstack? When I do, the mobile version is "overflowing" the page, and causing all sorts of layout issues. (It's always challenging to mix percentage widths with auto-sized content.) Specifically, since 592df93fef44ab4cf37cfdf4275766a67e7befec (before this PR), the panels of this xblock that are supposed to use "overflow: auto" and scroll within the panel instead are wider than the page, causing the whole page to be too wide:

frame

After struggling with this for quite a while, the only fix I could find is something we'd have to change at the platform level, and it seems a bit hacky:
```
.chromeless .course-content {
     max-width: 100vw;
}
```

I can't approve this PR until we also have a fix for ^

I also notice that I can no longer test the drag and drop in Chrome's mobile device emulator. When I click and hold for half a second, the context menu appears right when the drag should start (super weird). Did anyone else get that?

tests/integration/test_sizing.py
```diff
@@ -213,13 +213,13 @@ def _check_sizes(self, block_index, expectations, expected_img_width=None, is_de
                     container_width=item_bank_width,
                     expected_percent=expect.width_percent,
                 )
-            if expect.fixed_width_percent is not None:
-                self._check_width(
-                    item_description="Unplaced item {} with fixed width".format(expect.item_id),
-                    item=self._get_unplaced_item_by_value(expect.item_id),
-                    container_width=target_img_width,
```
cgopalan: @bradenmacdonald i was fixing the tests when I noticed this line in the section where the unplaced items are being tested. Shouldnt the unplaced items be tested against `item_bank_width` rather than `target_img_width`?

mtyaka:
@cgopalan @bradenmacdonald 

> After struggling with this for quite a while, the only fix I could find is something we'd have to change at the platform level, and it seems a bit hacky.

I've been playing with this for a while, but wasn't able to find any better CSS only solutions. We *could* add `.chromeless .course-content { max-width: 100vw; }` to the drag and drop xblock stylesheet, so we don't necessarily have to add it to the platform, although I'm not sure that makes it any better.

Another solution, which is a bit more involved, but I think would be better, is to measure the available width at render time and set the `max-width` of the `.drag-container` div dynamically.
See https://github.com/open-craft/xblock-drag-and-drop-v2/commit/53e36bd26c8af85bea04aa0af567c65c0519e026 for a quick prototype.

> I also notice that I can no longer test the drag and drop in Chrome's mobile device emulator. When I click and hold for half a second, the context menu appears right when the drag should start (super weird). Did anyone else get that?

That's an annoying feature of the Chrome mobile emulator. I paste this into the JS console to disable the context menu when testing:

```javascript
window.oncontextmenu = function() { return false; };
```

tests/integration/test_sizing.py
bradenmacdonald: @cgopalan I'm not sure, but that sounds like it would make more sense, yep.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -643,8 +643,9 @@ function DragAndDropTemplates(configuration) {
                         itemFeedbackPopupTemplate(ctx),
                         h('div.target-img-wrapper', [
                             h('img.target-img', {src: ctx.target_img_src, alt: ctx.target_img_description}),
+                            renderCollection(zoneTemplate, ctx.zones, ctx)
                         ]),
-                        renderCollection(zoneTemplate, ctx.zones, ctx)
+                        // renderCollection(zoneTemplate, ctx.zones, ctx)
```
mtyaka: Please remove this.

tests/integration/test_sizing.py
```diff
@@ -213,13 +213,13 @@ def _check_sizes(self, block_index, expectations, expected_img_width=None, is_de
...
-                    expected_percent=expect.fixed_width_percent,
-                )
+            # if expect.fixed_width_percent is not None:
+            #     self._check_width(
+            #         item_description="Unplaced item {} with fixed width".format(expect.item_id),
+            #         item=self._get_unplaced_item_by_value(expect.item_id),
+            #         container_width=target_img_width,
+            #         expected_percent=expect.fixed_width_percent,
+            #     )
```
mtyaka: Can these tests be fixed?

tests/integration/test_sizing.py
mtyaka: @cgopalan I think `target_img_width` is correct, since the percent is supposed to be relative to the background image, see https://github.com/open-craft/xblock-drag-and-drop-v2/blob/671ff9bd8bbbf7c53cc3c8977eb586f77a6f7a82/drag_and_drop_v2/public/js/drag_and_drop.js#L34-L40.

tests/integration/test_sizing.py
cgopalan: @mtyaka they were pretty complicated to fix. I will try though.

cgopalan:
@mtyaka The only way I could fix the tests was to specify a range to check the width against rather than exact values. I think we could be more specific rather than ranges, but might involve a whole lot more of refactoring of the test code itself.

mtyaka:
Rebased from c9ac2323cc825bf286a23e2df3339156750b5102.

mtyaka:
@cgopalan I merged my changes into this PR and replaced size ranges in tests with different values that depend on whether testing desktop or mobile. Please review when you get a chance.

mtyaka:
Hmm... some tests are failing on travis even though they pass for me locally. I'll have to look into that.

cgopalan:
@mtyaka I am afraid I don't know enough CSS to be able to review your changes. I did test the changes and they look fine. Can you tag somebody else from the team who knows more CSS to do this?

mtyaka:
@pomegranited or @e-kolpakov Could one of you do a quick review of this? Only the second commit needs attention, the first one was already reviewed by Braden and myself. The second commit fixes a bug that Braden discovered during review and it also fixes tests.

The changes were already tested by @cgopalan and myself, so you don't necessarily have to manually test them, but it would be good if someone could look through the new code before we merge this.

mtyaka:
Actually, I just found another problem while testing on master devstack :/ Will try to come up with w fix.

pomegranited:
@mtyaka  CC @e-kolpakov  I can review this; just ping me when it's ready.

mtyaka:
@pomegranited The problem was fixed and this is now ready for review.

pomegranited:
:+1: @mtyaka CC @cgopalan 

- [x] I tested this on a master devstack using the [DnDv2 demo course](https://github.com/open-craft/demo-courses).
  I used the chromeless view (e.g. http://localhost:8000/xblock/block-v1:OpenCraft+DnDv2+2017+type@drag-and-drop-v2+block@048d46b311cc45c3b27e6315d9d23359)
  I tested the various demos rotating the device to change orientation, testing scrolling the images, and zooming the screen. 
- [x] I read through the code
- [x] I checked for accessibility issues by ensuring I could still answer problems using the keyboard alone.  I also checked that messages and alerts are still properly read through VoiceOver.
- [x] Includes documentation - bugfix so user docs not required, but the changes are well-documented in the JS.

PR #138, opened 2017-10-03
Set max-width of drag container to available width.

mtyaka:
Measure the available width before rendering the drag container. Set the drag-container's max-width to the measured available width.

Tests TBD.

**Screenshots**:

**Before**:

screen shot 2017-10-03 at 12 29 11

**After**:

screen shot 2017-10-03 at 12 28 35

**Testing instructions**:

1. Set up a DnDv2 problem with a wide background image.
1. View the problem on a small screen (or resize you browser window).
1. Observe that without this patch, the DnDv2 problem overflows.
1. Observe that this patch fixes the problem and does not cause any issues in desktop view.

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -64,6 +64,7 @@
 /* drag-container holds the .items and the .target image */
 
 .xblock--drag-and-drop .drag-container {
+    box-sizing: border-box;
     width: auto;
     padding: 1%;
     background-color: #ebf0f2;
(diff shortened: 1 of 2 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=138&repo=edx-solutions%2Fxblock-drag-and-drop-v2"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mtyaka:
Closed in favor of https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/135.

PR #140, opened 2017-10-17
[MCKIN-6121] Prevent context menu on long draggable item tap

mtyaka:
When performing a long tap on a draggable item which contains an image, Mobile Chrome would show a contextmenu/popup, which made it impossible to start dragging the item.
In Desktop Chrome mobile emulation, the context menu would pop up even on draggables that did not contain an image.

We avoid this by preventing the default action when the `contextmenu` event is emitted on draggables.

This patch also contains a small fix where we draggable items would jump slightly to the left when being picked up from the bank. The reason for that was that we called `grabItem()`, which performs a redraw too early. We have to measure the position of the item in the bank *before* invoking `grabItem()`.

**Testing instructions**:

1. In Desktop Chrome mobile emulation mode, verify that you can drag items without the context menu popping up.
1. If you have an android device or emulator, perform the same check on Mobile Chrome or in the edX app. Note that on Android, the context menu would only pop up when tapping draggable items that contain an image, so make sure to test with draggables that contain images.

**Reviewers**:

- [x] @bradenmacdonald 
- [x] @sanfordstudent (?)

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -1384,7 +1384,6 @@ function DragAndDropBlock(runtime, element, configuration) {
             var item_id = $item.data('value');
             var item = getItemById(item_id);
             var $document = $(document);
-            grabItem($item, 'mouse');
             publishEvent({
                 event_type: 'edx.drag_and_drop_v2.item.picked_up',
                 item_id: item_id
@@ -1392,7 +1391,6 @@ function DragAndDropBlock(runtime, element, configuration) {
 
             var max_left = $container.innerWidth() - $item.outerWidth();
             var max_top = $container.innerHeight() - $item.outerHeight();
-            var item_size = {width: $item.width(), height: $item.height()};
             // We need to get the item position relative to the $container.
             var item_offset = $item.offset();
             var container_offset = $container.offset();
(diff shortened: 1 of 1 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=140&repo=edx-solutions%2Fxblock-drag-and-drop-v2"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mtyaka:
@sanfordstudent This is a small bug fix for Android/Mobile Chrome. Would you be able to review it?

bradenmacdonald:
@mtyaka (Not blocking for this PR): When testing this on a solutions devstack, I'm still getting issues with the width of the page. Do I need to change/update something to avoid this?

e.g. at http://localhost:8000/xblock/i4x://OpenCraft/DnDv2/drag-and-drop-v2/a070c0ddcd0b4ae9a8ef68c1d7c19b34 in mobile emulation mode, if I scroll right it looks like this:

screen shot 2017-10-18 at 1 52 09 pm

mtyaka:
Thanks @bradenmacdonald!

> When testing this on a solutions devstack, I'm still getting issues with the width of the page.

It looks like the sizing sometimes does not work correctly when there are many items in the item bank. I'll add a fix for that to #141.

PR #141, opened 2017-10-18
Fix: in IE11, the resize detector was blocking content

bradenmacdonald:
This PR fixes a bug introduced in 77192dce2f4cbac825b50d67fdb0bad45bdf43f2, which causes the Drag and Drop XBlock to be unusable in IE11. The problem was that the "about:blank" `<object>` (used for consistent resize detection) was obscuring the entire XBlock.

**Before** (with the size of the problem element reduced to make the effect obvious):
screen shot 2017-10-18 at 3 30 54 pm

**After**:
screen shot 2017-10-18 at 3 31 06 pm

**Reviewers**:

TBD

**Testing Instructions**:

1. Try to use this XBlock in IE11. Confirm that it works. Note there is a separate issue with IE11 and blocks that use SVG images for the background, so test a block with a .jpg or .png background.
2. Test in some other browsers, and test the resize detection to make sure there's no adverse impact from this change.

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -15,6 +15,7 @@
     overflow: hidden;
     pointer-events: none;
     z-index: -1;
+    visibility: hidden;
 }
 
 .xblock--drag-and-drop .initial-load-spinner {
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -648,6 +648,7 @@ function DragAndDropTemplates(configuration) {
         // image to 100%, so that it doesn't expand the container.
         if (ctx.drag_container_max_width === null) {
             target_img_style.maxWidth = '100%';
+            item_bank_properties.style = {display: 'none'};
         } else {
             drag_container_style.maxWidth = ctx.drag_container_max_width + 'px';
         }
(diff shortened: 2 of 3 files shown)


bradenmacdonald:
FYI @sanfordstudent - please review if you'd like. We'd like to merge this ASAP as it's a learner-facing issue (though I don't think the affected version got deployed on edx.org so not as urgent for you guys).

mtyaka:
Rebased from b4dd424621d4a2ee8ed3c363aa45bcb1eccdccf5.

mtyaka:
@bradenmacdonald Some were failing, so I added a fix for that. I also added a fix for the sizing issue you noticed in https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/140#issuecomment-337725224, and rebased. Please check and merge if it looks good to you.

mtyaka:
:+1:

- [x] I tested this: I verified the DnDv2 problems load and work correctly in IE11.
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~ n/a
- [ ] ~~Includes documentation~~ n/a - bugfix

bradenmacdonald:
Oooh, thanks so much for fixing that sizing issue, @mtyaka ! 👍  I tested this version in Chrome, Chrome mobile emulator, and IE11, and it's working nicely.

sanfordstudent:
So sorry for getting to this so late! Just squash the three commits and you're good to go.

bradenmacdonald:
Thanks @sanfordstudent ! Squashing and merging.

PR #142, opened 2017-10-20
Fixes failure when loading background image on IE11

kaizoku:
Image.height and Image.width returns 0 for SVGs loaded in JavaScript on IE11, includes a workaround to get the height and width of the SVG when they are set to 0.

**Testing Instructions**
1. Attempt to load a drag and drop block which uses an SVG for the background in Internet Explorer 11 (the [opencraft drag-and-drop-v2 demo course](https://github.com/open-craft/demo-courses) contains a few)
2. Load the same block in other browsers to ensure this doesn't alter functionality elsewhere

**Reviewers**
TBD

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -978,6 +978,14 @@ function DragAndDropBlock(runtime, element, configuration) {
         var promise = $.Deferred();
         var img = new Image();
         img.addEventListener("load", function() {
+            if (img.width == 0 || img.height == 0) {
+                // Workaround for IE11 issue with SVG images
+                document.body.appendChild(img);
+                var width = img.offsetWidth;
+                var height = img.offsetHeight;
+                document.body.removeChild(img);
+                img.width = width, img.height = height;
+            }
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @kaizoku! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=142&repo=edx-solutions%2Fxblock-drag-and-drop-v2"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

bradenmacdonald:
@sanfordstudent Would you like to do a second review of this?

bradenmacdonald:
@sanfordstudent Would you like to review this and/or #141 before I merge?

sanfordstudent:
Sorry @bradenmacdonald - just saw this in my email. Definitely go ahead and merge!

PR #145, opened 2017-10-30
Added publishing progress event no matter if grade is updated or not

e-kolpakov:
**Description:** This PR makes DnDv2 always publish progress event when grading occurs, no matter if grade have improved or not.

**JIRA:** This is an attempt to fix [MCKIN-6011](https://edx-wiki.atlassian.net/browse/MCKIN-6011)

**Testing instructions:**

All the normal operations should happen as usual. This code potentially fixes one particular use case:

1. Create new DnDv2
2. Fake StudentModule data for that XBlock to contain raw_score: 1.0
3. Open block in Apros, submit the solution
4. Make sure Apros progress indicators (on module page and on progress page) are updated after page refresh.

**Reviewers:**

- [x] @pomegranited 
- [ ] @bdero 
- [ ] edX reviewer TBD


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=145&repo=edx-solutions%2Fxblock-drag-and-drop-v2"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

pomegranited:
@e-kolpakov Unfortunately, this change registers any attempt at a DnDv2 unit as "complete", whether the problem has been completed or not.  This is how I tested that:

1. Create new DnDv2 (default setup is fine)
1. Open block in Apros, and drag one piece into place, but not the rest.
1. Refresh the page -- the unit shows up as complete.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -773,6 +773,9 @@ def _mark_complete_and_publish_grade(self):
             self.raw_earned = current_raw_earned
             self._publish_grade(Score(self.raw_earned, self.max_score()))
```
pomegranited: @e-kolpakov Maybe instead, you could publish the grade even `if current_raw_earned is None or current_raw_earned >= saved_raw_earned:`?

PR #146, opened 2017-11-01
Mention where to locate the block instead of how to install it

clemente:
Explicitly mention where to find this block, because it doesn't appear in the usual place (Advanced) as other xblocks. Remove installation instruction because it doesn't need installation.
Note that the [documentation](http://edx.readthedocs.io/projects/xblock-tutorial/en/latest/edx_platform/devstack.html#add-an-instance-of-the-xblock-to-a-unit) is asking users to look only under „Advanced“.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,17 +79,11 @@ potentially diverse audiences.
 Enabling in Studio
 ------------------
 
-You can enable the Drag and Drop XBlock in Studio through the Advanced
-Settings.
+Drag and Drop v2 XBlock is already included in Open edX.
 
-1. From the main page of a specific course, navigate to `Settings ->
-   Advanced Settings` from the top menu.
-2. Check for the `Advanced Module List` policy key, and add
-   `"drag-and-drop-v2"` to the policy value list.
-3. Click the "Save changes" button.
-4. You should see the new block `Drag and Drop` in the
-   `"Add New Component"` box. Click the green `Problem` button,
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @clemente! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx-solutions%2Fxblock-drag-and-drop-v2&number=146"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

bradenmacdonald:
@clemente Is it even still necessary to add "drag_and_drop_v2" to the advanced settings in Studio first?

clemente:
@bradenmacdonald Probably not, because it's already installed as a requirement. Maybe the documentation must cover two cases: older versions, recent versions.

bradenmacdonald:
@clemente No, the documentation does not need to cover older versions. I think you can remove steps 1-3 if you confirm they're not necessary.

clemente:
@bradenmacdonald Confirmed in a recent sandbox that it comes enabled by default, and simplified instructions.

PR #148, opened 2017-12-07
Make code more robust to handle edge cases

awaisdar001:
Do not break when a dict does not contain a key.

[EDUCATOR-1832](https://openedx.atlassian.net/browse/EDUCATOR-1832)

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -718,7 +718,7 @@ class DragAndDropBlock(
         item = self._get_item_definition(item_attempt['val'])
         is_correct = self._is_attempt_correct(item_attempt)
         if item_attempt['zone'] is None:
-            del self.item_state[str(item['id'])]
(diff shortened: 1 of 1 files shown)


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -717,12 +717,14 @@ def _drop_item_assessment(self, item_attempt):
 
         item = self._get_item_definition(item_attempt['val'])
         is_correct = self._is_attempt_correct(item_attempt)
+        str_item_id = str(item['id'])
         if item_attempt['zone'] is None:
-            del self.item_state[str(item['id'])]
```
ziafazal: I think if we replace this line with this statement we don't need to do anything else in this PR
```
self.item_state.pop(str(item['id']), None)
```

awaisdar001: Makes sense. 👍

awaisdar001:
@ziafazal I have updated the PR. Please have a look.

PR #160, opened 2018-06-13
MCKIN-7633 added horizontal slider

jawad-khan:
Added Horizontal slider.
[MCKIN-7633](https://edx-wiki.atlassian.net/browse/MCKIN-7633)
Added [bxslider](http://jsfiddle.net/VgHb7/85/)


pomegranited:
@jawad-khan  I'll be your first reviewer here, so please ping me when you're done with dev and tests are passing?  Thanks!

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1227,7 +1275,7 @@ function DragAndDropBlock(runtime, element, configuration) {
     };
 
     var focusFirstDraggable = function() {
-        $root.find('.item-bank .option').first().focus();
+        $root.find('.item-bank .option[draggable=true]').first().focus();
```
pomegranited: I'm surprised this change was necessary?  Aren't all `.item-bank .option` divs draggable?  (I could be wrong.)

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -880,7 +908,17 @@ function DragAndDropBlock(runtime, element, configuration) {
...
             }
         }
-    }
+    };
+
+    var initializeSlider = function() {
+        if (pageLoaded) {
+            itemSlider = $root.find('.item-bank').bxSlider({
+                pager: false,
+                touchEnabled: false,
```
pomegranited: Why disable touch?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -880,7 +908,17 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+                infiniteLoop: false
+            });
```
pomegranited: Looks like you can either style the "prev" and "next" text control links pretty heavily (which seems painful), or provide custom prev/next controls and pass them as [`nextSelector`/`prevSelector` options](https://bxslider.com/options/).

This [stackoverflow post](https://stackoverflow.com/a/16065654) also shows an approach that hooks up the click handlers, but I don't think you need to do that if you use the selectors.

drag_and_drop_v2/public/js/drag_and_drop.js
jawad-khan: With new requirement every placed item will have a placehoder in item bank.

drag_and_drop_v2/public/js/drag_and_drop.js
jawad-khan: We will discuss this in styling of this slider.

jawad-khan:
@pomegranited This PR is ready from my end. Please review it.

drag_and_drop_v2/public/js/drag_and_drop.js
jawad-khan: It was disabled just for testing purpose. I have enabled it.

drag_and_drop_v2/public/js/drag_and_drop.js
jawad-khan: https://edx-wiki.atlassian.net/browse/MCKIN-7633

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -842,7 +870,8 @@ function DragAndDropBlock(runtime, element, configuration) {
 
             initDraggable();
             initDroppable();
-
+            pageLoaded = true
```
pomegranited: Is there some other way to do this which doesn't require setting a (sort of) global variable?

e.g. by checking for the presence of a loaded page element?

drag_and_drop_v2/public/js/drag_and_drop.js
pomegranited: 👍

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1143,9 +1190,12 @@ function DragAndDropBlock(runtime, element, configuration) {
     var updateDOM = function() {
         var new_vdom = render(state);
         var patches = virtualDom.diff(__vdom, new_vdom);
+        destroySlider();
```
pomegranited: bxslider also offers a `reloadSlider` method -- would this work as a replacement for running destroy and then initialize again?

jawad-khan: Actually I used it deliberately, because in the next line I need root state when there is no bx slider wrappers on it(to compare it with old root version using virtualdom). Otherwise there are errors when comparing them.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -199,15 +214,19 @@ function DragAndDropTemplates(configuration) {
             className += " specified-width";  // The author has specified a width for this item.
         }
         var style = bankItemWidthStyles(item, ctx);
-        // Placeholder should never be visible.
-        style.visibility = 'hidden';
+        style.background = 'gray';
```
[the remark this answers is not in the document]
jawad-khan: Its temporary here, we are going to replace a placeholder here after styling.

jawad-khan:
@pomegranited I have made recommended changes. Please look at it again.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -41,6 +43,21 @@ function DragAndDropTemplates(configuration) {
...
+            orderedDragables[itemsOrder[i]] = dragables[i];
+        }
+        for (var i = 0; i < orderedDragables.length; i=i+items_per_slide) {
+            var first_row = h("div.row", orderedDragables.slice(i, i+(items_per_slide/2)));
+            var second_row = h("div.row", orderedDragables.slice(i+(items_per_slide/2), i+items_per_slide));
+            dividedDragables.push(h("div.slide", [first_row, second_row]));
+        }
+        return dividedDragables;
+    };
+
```
pomegranited: Instead of hardcoding the number of rows to 2, you can make that configurable as well, like this:
```diff
diff --git a/drag_and_drop_v2/public/js/drag_and_drop.js b/drag_and_drop_v2/public/js/drag_and_drop.js
index 35962b5..b56e84f 100644
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -3,7 +3,8 @@ function DragAndDropTemplates(configuration) {

     var gettext;
     var ngettext;
-    var items_per_slide = 8;
+    var cols_per_slide = 2;
+    var rows_per_slide = 3;

     if ('DragAndDropI18N' in window) {
         // Use DnDv2's local translations
@@ -43,17 +44,21 @@ function DragAndDropTemplates(configuration) {
         });
     };

-    var divideDragablesInTwoRows = function(dragables, itemsOrder) {
+    var divideDragablesIntoRows = function(dragables, itemsOrder) {
         var dividedDragables = [];
         var orderedDragables = [];
         for (var i=0; i<itemsOrder.length; i++)
         {
             orderedDragables[itemsOrder[i]] = dragables[i];
         }
-        for (var i = 0; i < orderedDragables.length; i=i+items_per_slide) {
-            var first_row = h("div.row", orderedDragables.slice(i, i+(items_per_slide/2)));
-            var second_row = h("div.row", orderedDragables.slice(i+(items_per_slide/2), i+items_per_slide));
-            dividedDragables.push(h("div.slide", [first_row, second_row]));
+        var i = 0;
+        while (i < orderedDragables.length) {
+            var rows = [];
+            for (var j = 0; j < rows_per_slide; j=j+1) {
+                rows[j] = h("div.row", orderedDragables.slice(i, i+cols_per_slide));
+                i += cols_per_slide;
+            }
+            dividedDragables.push(h("div.slide", rows));
         }
         return dividedDragables;
     };
@@ -684,7 +689,7 @@ function DragAndDropTemplates(configuration) {
         bank_children = bank_children.concat(renderCollection(itemPlaceholderTemplate, items_placed, ctx));
         var itemsOrder = inBankItemsOrder.concat(placedItemsOrder);

-        bank_children = divideDragablesInTwoRows(bank_children, itemsOrder);
+        bank_children = divideDragablesIntoRows(bank_children, itemsOrder);
         var drag_container_style = {};
         var target_img_style = {};
         // If drag_container_max_width is null, we are going to measure the container width after this render.
```

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -842,7 +872,8 @@ function DragAndDropBlock(runtime, element, configuration) {
 
             initDraggable();
             initDroppable();
-
+            initializeSlider = sliderInitializer;
+            sliderInitializer();
```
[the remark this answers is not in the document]
msaqib52: I guess i am tagged mistakenly here :)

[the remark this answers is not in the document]
pomegranited: Sorrt @msaqib52 !  Was reviewing your PR at the same time.  Will update.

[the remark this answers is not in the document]
jawad-khan: @pomegranited Actually We have to initialize bxslider after the complete page load i.e. after js and images have been loaded. It fails mostly when there are images in dragable tiles. So I used this pageload global variable as indicator that page has been loaded and we can safely initialize bxsliser.

https://github.com/stevenwanderski/bxslider-4#step-3-call-the-bxslider
https://stackoverflow.com/questions/48804763/bxslider-not-working-on-first-load-for-chrome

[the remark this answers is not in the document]
pomegranited: Ah ok, I didn't test the draggable tiles with images in, so didn't see that failing.

In that case, I think your original mechanism was clearer -- setting the `pageLoaded` global instead of replacing the `initializeSlider` function pointer.  At least that way, we can follow the process in the debugger.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -942,15 +941,16 @@ function DragAndDropBlock(runtime, element, configuration) {
...
-    var initializeSlider = function() {
-        if (pageLoaded) {
-            itemSlider = $root.find('.item-bank').bxSlider({
-               pager: false,
-                infiniteLoop: false,
-                startSlide: currentSlideIndex,
-                speed: 0
-            });
-        }
+    var sliderInitializer = function() {
```
pomegranited: @murad-hubib The tests are breaking because your commits have overridden @jawad-khan 's `initializeSlider` method, and so the slider isn't getting created.

Can you please fix this?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -903,7 +934,24 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+            infiniteLoop: false,
+            startSlide: currentSlideIndex,
+            speed: 0
+        });
+    };
+
+    var destroySlider = function() {
+        if (itemSlider){
+            currentSlideIndex = itemSlider.getCurrentSlide();
+            itemSlider.destroySlider();
```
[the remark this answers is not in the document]
pomegranited: 👍

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -903,7 +940,28 @@ function DragAndDropBlock(runtime, element, configuration) {
                 }
             }
         }
-    }
+    };
+
+    var sliderInitializer = function() {
```
pomegranited: @murad-hubib The tests are breaking because this commit removes @jawad-khan 's `initializeSlider` method, and so the slider isn't getting created.

Can you please remove these commits and submit a separate PR for the styling changes, so we can get this functionality merged to the feature branch ASAP?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -942,13 +942,15 @@ function DragAndDropBlock(runtime, element, configuration) {
...
-        itemSlider = $root.find('.item-bank').bxSlider({
-            pager: false,
-            infiniteLoop: false,
-            startSlide: currentSlideIndex,
-            speed: 0
-        });
+    var initializeSlider = function() {
+        if (pageLoaded) {
+            itemSlider = $root.find('.item-bank').bxSlider({
+               pager: false,
```
pomegranited: nit: indent 1 more space

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -942,13 +942,15 @@ function DragAndDropBlock(runtime, element, configuration) {
...
 
-    var sliderInitializer = function() {
-        itemSlider = $root.find('.item-bank').bxSlider({
-            pager: false,
-            infiniteLoop: false,
-            startSlide: currentSlideIndex,
-            speed: 0
-        });
+    var initializeSlider = function() {
+        if (pageLoaded) {
```
pomegranited: Thanks for reinstating this @jawad-khan -- could you please add a comment explaining why this is needed, so it doesn't get removed in the future?  e.g.
```javascript
// Need to wait for the full page load before loading the bxSlider,
// so that images contained in the droppable items are loaded properly.
```

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -942,13 +942,15 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+            itemSlider = $root.find('.item-bank').bxSlider({
```
pomegranited: @jawad-khan  If we disable touch (like you did during testing), then this is at least usable on mobile, because click events are registered instead.  So I'd accept adding the following as a workaround, including a comment to explain why:
```javascript
// Workaround: mobile touch drag and drop behavior is broken,
// but disabling touch allows click events to come through.
touchEnabled: false,
```

jawad-khan:
@pomegranited  I have made required changes and now tests are passing. Please review it.

jawad-khan:
@pomegranited Please squash commits and merge it. It seems like I don't have access to merge it.

bradenmacdonald:
@pomegranited Can you please use our normal OpenCraft review template for approvals here? In particular, it's critical that we check accessibility (esp. for keyboard and non-visual users) and documentation (esp. of API changes) with each PR.

pomegranited:
My apologies @bradenmacdonald -- The [comment I referenced](https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/160#pullrequestreview-137293988) in my approval comment uses our template, and shows the steps I took for accessibility testing and documentation, though here it was only in the form of code comments.  It didn't make sense to update the README with new screenshots yet, since the styling will be done in a subsequent PR.

The API was untouched here (I assume you mean the user data API?), and so I did not test it.

bradenmacdonald:
@pomegranited Ah, no worried then. Thanks.

PR #161, opened 2018-06-21
MCKIN-7668: toggle highlight on zones when item is dragged or dropped

msaqib52:
When a user starts dragging an item target zones should appear with green outline and a + sign once a user clicks on a draggable tile or uses keyboard shortcuts to drag a tile. Target zones disappear once draggable item is placed onto target area.

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -1320,6 +1320,7 @@ function DragAndDropBlock(runtime, element, configuration) {
                         placeGrabbedItem($zone);
                     }
                     releaseGrabbedItems();
+                    $root.find('.drag-container').removeClass('dragging');
                 }
             } else if (isTabKey(evt) && !evt.shiftKey) {
                 // If the user just dropped an item to this zone, next TAB keypress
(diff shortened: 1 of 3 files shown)


pomegranited:
@sanfordstudent Could we get a review of this added functionality?

msaqib52:
@pomegranited i have updated this PR with test cases.

mduboseedx:
@marcotuts Any recommendations on who could review something like this in the future? Typically, for drag and drop changes done for McKinsey, @sanfordstudent would review before merging it into the master branch.

bradenmacdonald:
@msaqib52 Will you be opening one big PR to merge this to `master` or should we open an additional PR for each ticket? I want to make sure these "v4" PRs all get merged to master and soon. If a huge PR would be challenging to review, we should do it now we each individual PR too.

CC @pomegranited

msaqib52:
@bradenmacdonald we are planning to open one big PR when v4 functionality is complete. Merging these PRs now would push partial v4 functionality if we make a release from master, but if that is not a concern and huge PR would be challenging to review than we can do individual PRs too.

pomegranited:
Thanks for your reply @msaqib52  -- what about doing feature-based upstream PRs?  Like including all the highlight zone PRs into one change, and all of the draggable item container changes in another? Straightforward bugfixes could stand alone.

msaqib52:
@pomegranited yes we can do feature-based upstream PRs. But these PRs are also changing UI structure and colors, so again doing feature-based upstream PRs would push some components with new UI/color scheme and some with old, which might look odd. I will confirm with our PO if we are ok with this approach.

xitij2000:
@msaqib52  I think that is OK, we will only tag a release once all the functionality is finished.

PR #166, opened 2018-07-17
Item Bank slider and popups slider styling

jawad-khan:
MCKIN 7636
MCKIN 7638
MCKIN 7741

### DIFF
--- a/drag_and_drop_v2/public/css/drag_and_drop.css
+++ b/drag_and_drop_v2/public/css/drag_and_drop.css
@@ -100,20 +100,24 @@
     align-items: center;
 
     position: relative;
-    border: 1px solid rgba(0, 0, 0, 0.1);
     border-radius: 3px;
     padding: 0; /* padding: 5px looks better but makes some blocks to change in size when dropped onto the target; */
 }
 
+.xblock--drag-and-drop .drag-container .bx-wrapper .option {
+    width: 150px;
+    height: 55px;
+
+}
+
 .xblock--drag-and-drop .drag-container .option {
     display: inline-block;
-    width: auto;
     min-width: 4em;
     max-width: 30%;
     border: 1px solid transparent;
-    border-radius: 3px;
+    border-radius: 2px;
     box-sizing: border-box;
-    margin: 5px;
+    margin: 5px 5px 0;
     padding: 10px;
     text-align: center;
     background-color: #1d5280;
@@ -125,6 +129,46 @@
     /* Some versions of the drag and drop library try to fiddle with this */
     z-index: 10 !important;
     -webkit-touch-callout: none;
+    position: relative;
+}
+
+.xblock--drag-and-drop .drag-container .dropped {
+    background-color: gray;
+    color: gray;
+}
+
+/*Item bank slider*/
+
+.bx-wrapper{
+    background: #fff;
+    padding: 0 55px;
+    position: relative;
+}
+
+.item-bank .slide{
+    padding: 10px 0;
+    text-align: center;
+}
+
+.bx-controls .bx-controls-direction > a{
+    position: absolute;
+    left: 18px;
+    top: 50%;
+    transform: translateY(-15px);
+}
+
+.bx-controls .bx-controls-direction > a .fa{
+    font-size: 50px;
+    line-height: 30px;
+    color: #1d5280;
+}
+.bx-controls .bx-controls-direction > a.bx-next{
+    right: 18px;
+    left: auto;
+}
+
+.bx-controls .bx-controls-direction > a.disabled .fa{
+    color: #e3e8f0;
 }
 
 @media screen and (max-width: 480px) {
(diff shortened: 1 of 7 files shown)


jawad-khan:
@pomegranited Please review it.

drag_and_drop_v2/public/css/drag_and_drop.css
```diff
@@ -100,20 +100,20 @@
...
     position: relative;
-    border: 1px solid rgba(0, 0, 0, 0.1);
     border-radius: 3px;
     padding: 0; /* padding: 5px looks better but makes some blocks to change in size when dropped onto the target; */
 }
 
 .xblock--drag-and-drop .drag-container .option {
     display: inline-block;
-    width: auto;
+    width: 150px;
```
kaizoku: Should we be hardcoding the size of these here?
The width can be set as a percentage in the studio editor, and this would likely override that.

murad-hubib: @kaizoku Thanks for the review. I have pushed the said changes except the width of tiles as they are fixed in the deisgn. You probably don't have all the changes for drag and drop in this PR so the tiles are not fitting in. Please let his PR merged in drag-and-drop-v4 first to have the final results.

Thanks

bradenmacdonald:
@murad-hubib @kaizoku 

> I have pushed the said changes except the width of tiles as they are fixed in the deisgn. 

> My thinking is that since there is a "Preferred Width" option in studio to set the width of each option, we should allow that to set the width. If we override it here, that setting would have no longer have any purpose.

We need to support both approaches, at least for now. @sstack22 is going to investigate how many course authors are currently using the "Preferred width %" option, and if it's small we may be able to remove that. 

But in any case, for now, please add a new setting on the first screen of the course author settings:

Tile size: [Automatic | Rectangle | Square | Custom]

Automatic: use current algorithm (flexbox based) to fix according to text
Rectangle: use the fixed rectangular size from the new designs for all tiles
Square: use the fixed square size from the new designs for all tiles
Custom: Allow authors to set a preferred % hint for the tile sizes, for each tile. The height will be computed automatically to fit the tile contents. (We hope to deprecate that option but for now we need to continue to support it).

In general, with all of this work, please find an approach like this that preserves the existing functionality while also supporting the new features and UX.

CC @pomegranited @xitij2000

bradenmacdonald:
@janomanningham sorry for the delay in replying.

> If we provide all four tile size options then how do we render the item bank that supports the selected tile size? For example, since the rectangle and square tiles have predefined sizes, the item bank that houses those tiles was built specifically for those tile sizes and uses a slider to navigate horizontally.

Any HTML-based item bank should be able to support arbitrary tile sizes inside it? They might not fit into a neat grid but as long as the item bank uses what we call "flexbox layout" or a "table layout", the tiles should fit into it fine and be scrollable, basically looking like the current item bank does, but with a scrollbar. But if there's some limitation with the scrolling item bank container that we're using, then what you said about "When a custom tile is created, the rectangle and square tiles become custom and it will render the item bank that currently is used in production today. " sounds fine to me too.

@jawad-khan Is there any technical reason why the item bank slider won't work if it contains tiles of different sizes?

--------

I do think we should stick to having all tiles square, all tiles rectangular, or all tiles custom in any given drag and drop exercise, which will make things easier than allowing a mix.

kaizoku:
@jawad-khan I'm still unsure as PR will still likely break the styling of any other users by setting a fixed width for the draggable options.

Are there no other users of the `drag-and-drop-v4` branch?
If we're sure that this isn't used by anyone else I'm fine with approving and merging this one.

jawad-khan:
@bradenmacdonald We made slider and tiles according to new design, which doesn't support tiles with different sizes. Therefore different tile sizes wouldn't work with this slider. There is no technical reason behind it, it wasn't made to support different tile sizes. 
If we want custom tile sizing then we first have to change our designs according to it, and the easiest way to do it was to support old design for custom sizing and new for automatic.
We can support custom sizing in this branch but we need clear designs.

xitij2000:
@jawad-khan I think what @bradenmacdonald is saying is that when you have a custom tile size, then don't use the new slider, but use the existing layout for tiles. Of course, the best case situation is to allow the existing slider to show custom tiles, but if that isn't possible, just show the old tile bank vs the new slider based on whether there is a custom tile. 

This way you can get the new slider layout by only using square/rectangle tiles, but if you do select a custom tile, everything still works, you just get the old item bank.

jawad-khan:
@kaizoku Can you please review it again?

kaizoku:
Thanks for making this backward compatible @jawad-khan.
This should help a lot of users migrate to the new version.
:+1:

- [✓] I tested this: (deployed in devstack, manually tested both fixed and free sizing modes)
- [✓] I read through the code
- [✓] I checked for accessibility issues
- [✓] Includes documentation

jawad-khan:
@kaizoku Can you please squash and merge it?

PR #169, opened 2018-08-16
MCKIN-8225: Add item style/size option

msaqib52:
Allow course authors to set item style/size option from studio

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -129,20 +129,21 @@ class DragAndDropBlock(
         enforce_type=True,
     )
 
-    item_size = String(
-        display_name=_("Item size"),
+    item_sizing = String(
+        display_name=_("Item sizing"),
         help=_(
-            "Standard size: makes items square or rectangular based on "
-            "the number of characters in an item. "
-            "Fit to size: makes items fit to text content of an item, also "
-            "supports custom width option for item sizing."
+            "Standard sizing: The item will snap to a rectangular size for "
+            "items with a small amount of text, or double in size to a square "
+            "for items containing an image or large amounts of text. "
+            "Free sizing: The item size will fit the specified maximum width, "
+            "or freely expand with the size of the content."
         ),
         scope=Scope.settings,
         values=[
-            {"display_name": _("Standard size"), "value": Constants.STANDARD_SIZE},
-            {"display_name": _("Fit to size"), "value": Constants.FIT_TO_SIZE},
+            {"display_name": _("Standard sizing"), "value": Constants.STANDARD_SIZING},
+            {"display_name": _("Free sizing"), "value": Constants.FREE_SIZING},
         ],
-        default=Constants.STANDARD_SIZE,
+        default=Constants.STANDARD_SIZING,
         enforce_type=True,
     )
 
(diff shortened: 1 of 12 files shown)


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -129,6 +129,23 @@ class DragAndDropBlock(
         enforce_type=True,
     )
 
+    item_size = String(
+        display_name=_("Item size"),
+        help=_(
+            "Standard size: makes items square or rectangular based on "
+            "the number of characters in an item. "
+            "Fit to size: makes items fit to text content of an item, also "
+            "supports custom width option for item sizing."
```
xitij2000: I think these two options are a bit confusing. Both of them seem to fit to the text content, so what is the difference?

While I understand the difference here, because I have been part of these discussions, I think for a course author these two options would be a bit confusing.

msaqib52:
@xitij2000 thank you for the review, i have updated this PR with suggested changes, could you take another look?

xitij2000:
@msaqib52 It's looking good and working as expected, I'm still going through the code, but one thing I think you should add is a test for the new functionality. i.e. that hiding / showing the preferred width based on which kind of sizing is used.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -129,6 +129,24 @@ class DragAndDropBlock(
...
+            "for items containing an image or large amounts of text. "
+            "Free sizing: The item size will fit the specified maximum width, "
+            "or freely expand with the size of the content."
+        ),
+        scope=Scope.settings,
+        values=[
+            {"display_name": _("Standard sizing"), "value": Constants.STANDARD_SIZING},
+            {"display_name": _("Free sizing"), "value": Constants.FREE_SIZING},
+        ],
+        default=Constants.STANDARD_SIZING,
```
xitij2000: Could you remove this default value? Setting this as the default means that all old blocks that don't have this setting will automatically use standard sizing and this might break their layout. If there is no default then the old behaviour (Free Sizing) can be used when no value is present.

msaqib52: How about we set free sizing as default to stay backward compatible, also rename standard sizing to fixed sizing because the name `standard` implies default?

xitij2000: Setting it as the default will also set it as the default in the UI. I'm not sure about calling it Fixed sizing since it can toggle between a square and rectangle if we were to go that way perhaps it could be called Automatic Sizing? I'll post on the ticket.

xitij2000:
@msaqib52 Could you take a look at the test failures.

drag_and_drop_v2/drag_and_drop_v2.py
msaqib52: `Setting it as the default will also set it as the default in the UI`, It would still be default in UI If we don't set it as default because if there is no default we will still be setting free sizing in UI to stay backward compatible.

xitij2000: My point was that we wouldn't want it as the default in the UI. Right now Standard Sizing is the first option so it's kind of selected by default even if it isn't explicitly set.

msaqib52:
@xitij2000 build is green now, also removed `Standard sizing` as default.

PR #173, opened 2018-09-06
MCKIN-8314: Combine authored hints and persisted feedback in one popup window for fixed item sizing

msaqib52:
Combine authored hints and persisted feedback in one popup window for fixed item sizing in assessment mode.

@xitij2000 could you please review this

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -346,6 +346,57 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
+    var getAssessmentModeFixedSizingFeedbackDetailTemplate = function(ctx) {
+        var message = ctx.max_attempts
+            ? gettext('{used} / {total}').replace('{used}', ctx.attempts).replace('{total}', ctx.max_attempts)
+            : ctx.attempts.toString();
+        var attempts_used_html = h('div.attempts', [
+            message,
+            h('small', gettext('Attempts'))
+        ]);
+        var misplaced_items_html = h('p', 
+            gettext('You have misplaced {misplaced_items_count} items').replace(
+                "{misplaced_items_count}", ctx.misplaced_items_count
+            )
+        );
+        var see_answers_button_html = h('button.see-correct-answers-button', gettext('See Correct Answers'));
+        var try_gain_button_html = h('button.try-again-button', gettext('Try Again'));
+
+        if (ctx.answer_correctness == DragAndDropBlock.SOLUTION_CORRECT) {
+            return [
+                h('div.icon', [h('i.fa.fa-check')]),
+                h('h3', gettext('Congratulations!')),
+                h('p', gettext('You have placed all the items in the correct drop zones')),
+                attempts_used_html,
+                h('button.finish-button', gettext('Finish'))
+            ];
+        } 
+
+        return [
+            h('div.icon', [h('i.fa.fa-check')]),
+            h('h3', ctx.answer_correctness == DragAndDropBlock.SOLUTION_INCORRECT ? gettext('Incorrect Answers') : gettext('Partially Correct!')),
+            misplaced_items_html,
+            attempts_used_html,
+            ctx.attempts == ctx.max_attempts ? see_answers_button_html : try_gain_button_html
+        ];
+    };
+
+    var fixedSizingFeedbackTemplate = function(ctx) {
+        var hints = ctx.feedback_messages || [];
+        var feedbackClass = (hints.length > 0) ? 'show-hints ' : ' ';
+        feedbackClass += ctx.answer_correctness;
+        
+        return h('div.fixed-sizing-feedback', { className: feedbackClass, style: {display: ctx.show_fixed_sizing_feedback ? 'block' : 'none'}}, [
+            h('div.feedback-detail', getAssessmentModeFixedSizingFeedbackDetailTemplate(ctx)),
+            h('div.hint-detail', [
+                h('h5', gettext('Hints')),
+                h("ol", {}, hints.map(function(hint) {
+                    return h("li", {innerHTML: gettext(hint.message)});
+                }))
+            ])
+        ]);
+    };
+
     var keyboardHelpPopupTemplate = function(ctx) {
         var labelledby_id = 'modal-window-title-'+configuration.url_name;
         return (
@@ -389,7 +440,7 @@ function DragAndDropTemplates(configuration) {
             var attemptsUsedTemplate = gettext("{used} / {total}");
             var attemptsUsedText = attemptsUsedTemplate.
                 replace("{used}", ctx.attempts).replace("{total}", ctx.max_attempts);
-            attemptsUsedInfo = h("div.attempts-used", {id: attemptsUsedId}, [h("span",attemptsUsedText), h('span.text', "Attempts")]);
+            attemptsUsedInfo = h("div.attempts-used", {id: attemptsUsedId}, [h("span",attemptsUsedText), h('span.text', gettext("Attempts"))]);
         }
 
         var submitSpinner = null;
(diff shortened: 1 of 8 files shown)


drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -331,6 +335,54 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
+    var getFixedSizingFeedbackDetailTemplate = function(ctx) {
+        var attempts_used_html = h('div.attempts', [
+            gettext('{used} / {total}').replace('{used}', ctx.attempts).replace('{total}', ctx.max_attempts),
```
xitij2000: If `max_attempts` is unset it will show up as `null` here. If `max_attempts` is `null` the message should not display it at all.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -331,6 +335,54 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
+    var getFixedSizingFeedbackDetailTemplate = function(ctx) {
```
xitij2000: I think this should be called `getAssessmentModeFeedbackDetailTemplate` since this has more to do with assessment mode and less to do with fixed sizing.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -374,7 +426,7 @@ function DragAndDropTemplates(configuration) {
             var attemptsUsedTemplate = gettext("{used} / {total}");
             var attemptsUsedText = attemptsUsedTemplate.
                 replace("{used}", ctx.attempts).replace("{total}", ctx.max_attempts);
-            attemptsUsedInfo = h("div.attempts-used", {id: attemptsUsedId}, [h("span",attemptsUsedText), h('span.text', "Attempts")]);
+            attemptsUsedInfo = h("div.attempts-used", {id: attemptsUsedId}, [h("span",attemptsUsedText), h('span.text', gettext("Attempts"))]);
```
xitij2000: :+1:

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -2002,7 +2083,9 @@ function DragAndDropBlock(runtime, element, configuration) {
             problem_html: configuration.problem_text,
             show_problem_header: configuration.show_problem_header,
             show_submit_answer: configuration.mode == DragAndDropBlock.ASSESSMENT_MODE,
-            show_show_answer: configuration.mode == false,
+            show_show_answer: (configuration.mode == DragAndDropBlock.ASSESSMENT_MODE && 
+                configuration.item_sizing == DragAndDropBlock.FREE_SIZING),
+            show_feedback_bar: !canShowFeedbackBar(),
```
xitij2000: Please update `canShowFeedbackBar` to return the correct result so it doesn't need to be negated.

xitij2000:
@msaqib52 Here is what I'm seeing on the LMS:
DND V4 popul

drag_and_drop_v2/public/js/drag_and_drop.js
msaqib52: @xitij2000 this will be called for fixed item sizing in assessment mode only.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1884,7 +1956,12 @@ function DragAndDropBlock(runtime, element, configuration) {
             state.feedback = data.feedback;
             state.overall_feedback = data.overall_feedback;
             state.last_action_correct = data.correct;
+            state.answer_correctness = data.answer_correctness;
+            state.misplaced_items_count = data.misplaced_items.length;
 
+            if (isItemSizingFixed(configuration)) {
+                state.fixed_sizing_feedback_display = 'block';
```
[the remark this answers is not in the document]
msaqib52: @xitij2000 having `state.fixed_sizing` and setting it to true or false would give an impression that we are some how turning fixed sizing on/off which might be confusing.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -2109,7 +2112,7 @@ function DragAndDropBlock(runtime, element, configuration) {
             disable_go_to_beginning_button: !canGoToBeginning(),
             show_go_to_beginning_button: state.go_to_beginning_button_visible,
             screen_reader_messages: (state.screen_reader_messages || ''),
-            fixed_sizing_feedback_display: (state.fixed_sizing_feedback_display || 'none'),
+            show_fixed_sizing_feedback: (state.show_fixed_sizing_feedback || false),
```
xitij2000: This can just be ``show_fixed_sizing_feedback: state.show_fixed_sizing_feedback,``.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -335,11 +331,14 @@ function DragAndDropTemplates(configuration) {
...
-            h('small', gettext('Attempts'))
-        ]);
+    var getAssessmentModeFixedSizingFeedbackDetailTemplate = function(ctx) {
+        var attempts_used_html = null;
+        if (ctx.max_attempts) {
+            attempts_used_html = h('div.attempts', [
+                gettext('{used} / {total}').replace('{used}', ctx.attempts).replace('{total}', ctx.max_attempts),
+                h('small', gettext('Attempts'))
+            ]);
+        }
```
xitij2000: Instead of completely disabling the display, perhaps it could be something like this:
```javascript
let message = ctx.max_attempts 
    ? gettext('{used} / {total}').replace('{used}', ctx.attempts).replace('{total}', ctx.max_attempts)
    : ctx.attempts.toString();
attempts_used_html = h('div.attempts', [
    message,
    h('small', gettext('Attempts'))
]);
```
This will show the message as "8 / 12 Attempts"  if there is a ``max_attempts``, and "8 Attempts" otherwise.

drag_and_drop_v2/public/js/drag_and_drop.js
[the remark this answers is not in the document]
xitij2000: Sure, I think the variable name you picked is a lot more clear so :+1:

xitij2000:
@msaqib52 I don't see any CSS changes in the latest push. As you can see the current syling on the LMS without the yonk theme is not looking good. Could you look into that?

msaqib52:
@xitij2000 yeah murad was working on CSS changes i have pushed those changes.

xitij2000:
@msaqib52 I'm having a look at this now. Meanwhile could you fix the conflicts?

xitij2000:
@msaqib52 Unfortunately there are still a couple of issues. 
1. The popup styling is still a bit off as you can see here:
Congratulations is misaligned
The congratulations text (feedback heading) is misaligned. 

2. It seems the accessibility ordering is quite off now. 
  - After placing an item, the focus no longer goes to the next item. 
  - After placing all items, the focus no longer goes to the submit button. 
  - When the popup shows up the focus isn't on the "Try Again" / "Finished" button, in fact I wasn't able to tab to it at all.

msaqib52:
@xitij2000 this strange because i tested it on two machines and UI was looking fine in studio, do you have latest changes for platform? also can you try after clearing the cache? meanwhile i can fix the accessibility issues.

xitij2000:
@msaqib52 The issue I am seeing is on the LMS, not studio. I don't think it's a caching issue, since the colour of the text has changed, and that was an issue before so changes are being picked up. 

I think adding a `text-align: center` [here](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/2e5ec823385603ce2b82bd390bea29d7e940ae2c/drag_and_drop_v2/public/css/drag_and_drop.css#L598) should fix it.

msaqib52:
@xitij2000 i have css issue in studio, as far as accessibility is concerned i have fixed your third point.
Other points are working as they were previously.

`After placing an item, the focus no longer goes to the next item.`: The focus is supposed to be on `Go to beginning` button when you press `TAB` after placing an item and on pressing `Go to beginning` button focus goes to next item, this is how it was working previously and same is working now.

`After placing all items, the focus no longer goes to the submit button.`: In this case one had to press `Tab` key to focus `Submit` button and same is working now.

PR #177, opened 2018-09-26
MCKIN-8027 Assessment notification for assessment mode in fixed sizing

jawad-khan:
Assessment notification for assessment mode in fixed sizing.

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -826,6 +826,7 @@ function DragAndDropTemplates(configuration) {
                 ]),
                 h('div.drag-container', {style: drag_container_style, className: ctx.show_instructions ? 'instructions-visible' : ''}, [
                     instructionsPopupTemplate(ctx),
+                    assessmentNotificationTemplate(ctx),
                     h('div.target', {attributes: {'role': 'group', 'arial-label': gettext('Drop Targets')}}, [
                         h('div.target-img-wrapper', [
                             h('img.target-img', {
@@ -839,7 +840,6 @@ function DragAndDropTemplates(configuration) {
                     h('div.item-bank', item_bank_properties, bank_children),
                     ctx.show_feedback_bar ? itemFeedbackPopupTemplate(ctx) : null,
                     h('div.dragged-items', renderCollection(itemTemplate, items_dragged, ctx)),
-                    assessmentNotificationTemplate(ctx),
                 ]),
                 h("div.actions-toolbar", { attributes: {'role': 'group', 'aria-label': gettext('Actions')}}, 
                 actionButtonsTemplate(ctx)),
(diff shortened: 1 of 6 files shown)


drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -321,6 +321,26 @@ function DragAndDropTemplates(configuration) {
...
+    var assessmentNotificationTemplate = function (ctx) {
+        if (configuration.mode === DragAndDropBlock.STANDARD_MODE) {
+            return
+        }
+        var icon = h('div.icon' , h('i.fa.fa-exclamation'));
+        var button = h('div.action', h('button.close-assessment-notification', gettext("Continue")));
+         var attemptText = h('p',
+            gettext(
+                "You can review answers / resubmit {max_attempts} times"
+            ).replace("{max_attempts}", ctx.max_attempts)
```
xitij2000: If `ctx.max_attempts` is null, this shows up as "You can review answers / resbumit null times.". I think the language should be changes in case of null or 0 `max_attempts` to be:
"You can review answers / resubmit as many times as you want."

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -321,6 +321,26 @@ function DragAndDropTemplates(configuration) {
         );
     };
 
+    var assessmentNotificationTemplate = function (ctx) {
+        if (configuration.mode === DragAndDropBlock.STANDARD_MODE) {
+            return
+        }
+        var icon = h('div.icon' , h('i.fa.fa-exclamation'));
+        var button = h('div.action', h('button.close-assessment-notification', gettext("Continue")));
+         var attemptText = h('p',
```
xitij2000: nit: these lines seem to be misaligned.

xitij2000:
@jawad-khan  Could you look into the failing tests? I will give it another review pass once those are addressed.

jawad-khan:
@xitij2000 Tests are passing now. Please have a look.

xitij2000:
@jawad-khan Unfortunately the focus still doesn't seem to be on the continue button to begin with. I need to tab though all the zones and all the items and links and then I get to the continue button. The continue button should be the first in the tab order when visiting the block.

jawad-khan:
@xitij2000 Actually for me it is working fine. How are you testing it?

xitij2000:
@jawad-khan I'm testing it on Linux on Firefox with a screen reader. 

I just noticed you pushed a new change to rectify this, but that might not work well since it will focus the xblock first on the page, and if there are two such problems on the page the second problem might end up coming first. 

I think the solution here would be to restructure the html such that the popup html comes first, and its visual position can be controlled by CSS.

xitij2000:
@jawad-khan I did a quick test, and simply moving ``assessmentNotificationTemplate(ctx),`` above to right after `instructionsPopupTemplate(ctx),` should solve this problem.

UmanShahzad:
@jawad-khan Please ping me when tests are passing, btw, and I will merge.

jawad-khan:
@UmanShahzad Tests are passing now.
Can you please merge it?

UmanShahzad:
@jawad-khan Yep, will do.

UmanShahzad:
@jawad-khan by the way, I didn't see any new commits. Are the tests flaky and you just restarted them? We should fix flaky tests too.

jawad-khan:
@UmanShahzad Yes we already have some flaky tests and we have fixed many tests in another branch which is not merged yet. So after that branch they will be gone.

PR #178, opened 2018-09-27
MCKIN-8163: show feedback when last correct item place in fixed sizing

msaqib52:
@xitij2000 can you review this?


kaizoku:
@msaqib52 Thanks for adding this. I had a bit of a styling issue on firefox, the icon had a strange pink background, but I think it was just a caching issue. After reloading the LMS it looks good.

:+1:

- [✓] I tested this: (deployed in devstack, ran through problems of standard and assessment types)
- [✓] I read through the code
- [✓] I checked for accessibility issues
~- [n/a] Includes documentation~

PR #179, opened 2018-10-01
MCKIN-8525: add answer_correctness to student_view_user_state

msaqib52:
This PR adds user answer correctness status to student_view_user_state handler.


xitij2000:
@msaqib52 I've added this to next sprint. Will be reviewed then.

PR #180, opened 2018-10-09
MCKIN-7971 Create rectangle and square tile sizes for Fixed Sizing

jawad-khan:
Create rectangle and square tile sizes for fixed sizing.
[https://edx-wiki.atlassian.net/browse/MCKIN-7971](MCKIN-7971)


xitij2000:
@jawad-khan Could you close this PR if it is no longer needed?

PR #192, opened 2018-11-10
Fix JavaScript error when the XBlock is in an IFrame

bradenmacdonald:
When this XBlock runs in a sandboxed IFrame (on a different origin than the LMS), it was throwing an error, because `domNode.contentDocument` was null.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ install:
     - "pip install selenium==2.53.0"
     - "pip uninstall -y xblock-drag-and-drop-v2"
     - "python setup.py sdist"
-    - "pip install dist/xblock-drag-and-drop-v2-2.2.0.tar.gz"
+    - "pip install dist/xblock-drag-and-drop-v2-*.tar.gz"
 script:
     - pep8 drag_and_drop_v2 tests --max-line-length=120
     - pylint drag_and_drop_v2
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -850,6 +850,23 @@ function DragAndDropBlock(runtime, element, configuration) {
         });
     };
 
+    // Handler to react when the size of this XBlock changes:
+    var last_width = 0;
+    var last_height = 0;
+    var raf_id = null;
+    var handleResizeEvent = function(watchedObject) {
+        cancelAnimationFrame(raf_id);
+        raf_id = requestAnimationFrame(function() {
+            var new_width = watchedObject.width();
+            var new_height = watchedObject.height();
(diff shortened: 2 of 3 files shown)


mtyaka:
@bradenmacdonald We're using a hidden `object` element to detect changes to the available space using [this method|http://www.backalleycoder.com/2013/03/18/cross-browser-event-based-element-resize-detection/], which is basically a polyfill for Resize Observer. We're adding an `object` element and attaching a resize handler *to that object*, that's why we use `object.contentDocument` because 
it's not the same as the top level `document`.

While most of the time the available space will change when the window gets resized, it can also change without resizing the window in some cases - for example when surrounding elements are added, removed, or hidden dynamically, expanding the available space for the element we're interested in.

I do not know if that's the case in LMS, and perhaps a resize handler on the `document` will catch all cases. In that case we should simplify the code to remove the `.resize-observer` object element and just bind the resize handler directly to the current window.

I am not sure why `domNode.contentDocument` is null when running in an iframe. I guess it's some kind of a security feature? Although it doesn't make much sense to me since both the `object` element and the JS that's trying to access its inner document are running in the same iframe.

mtyaka:
@bradenmacdonald An alternative fix could be to replace that `object` with a blank `iframe`, add some JS to the iframe to detect resizes of its content window, and then report the size up to DnD via `postMessage`?

I feel a bit uneasy removing that resize observer because I'm sure it was put in for a reason, although I don't remember what exactly the reason was.

bradenmacdonald:
@mtyaka That would work but require a lot of code to implement the message passing.

How about instead we just try the existing approach, but if `object.contentDocument` is `null`, we listen to `window`'s resize event instead? After all, in an IFrame context, the only way the object's size is going to change is if the IFrame's size changes, because there is no surrounding HTML that may change size.

bradenmacdonald:
@mtyaka Ok, I've updated this along those lines and confirmed it works in both cases.

mtyaka:
Thanks @bradenmacdonald, that makes sense.

I added a request for a small change, but otherwise this looks good and was tested in both LMS and LX :+1:

- [x] I tested this: I verified that with this change the block resizes appropriately in both LMS/Studio and when presented in an iframe in an external application.
- [x] I read through the code
- [ ] ~~Includes documentation~~ n/a

bradenmacdonald:
Thanks for the excellent review and suggestions, @mtyaka :)

Fixed and squashed.

bradenmacdonald:
@natabene Does someone from edX want to give this PR a second review before it's merged? It's just a small technical fix, no product changes.

natabene:
@bradenmacdonald  I think you are fine to merge as is.

PR #204, opened 2019-04-04
(BB-1079) Fix bug affecting dragging on mobile

xirdneh:
When having a large draggable item the limits are not calculated correctly and the xblock doesn't let the user to drag the item all the way to the right. This happens because the limit is calculated once when the drag is [initiated](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/public/js/drag_and_drop.js#L1465). But when the item is moved the actual div that gets moved is a different one created when an item is marked as [dragged](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/public/js/drag_and_drop.js#L175) and a different [style](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/122da26ee82dc4f4873e2172b0a618c50a5883c4/drag_and_drop_v2/public/css/drag_and_drop.css#L112) is applied to it making the width only 30% of the original. This makes the limit be wrong. The solution is to calculate the limits by using the actual div that's being dragged.
This PR calculates the limits correctly and it only does it once by using a sort of memoization for performance.

**Screenshots**: 
Wrong limits calculated in mobile view
dnd-before

Behavior with the updates in this PR
dnd-after

**Testing instructions**:

1. Crate a Drag and Drop problem.
2. Add an item with large text.
3. Try dragging the item using a mobile view.

**Reviewers**
- [ ] @xitij2000 
- [ ] edX reviewer[s] TBD

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -1462,8 +1462,9 @@ function DragAndDropBlock(runtime, element, configuration) {
                 item_id: item_id
             });
 
-            var max_left = $container.innerWidth() - $item.outerWidth();
-            var max_top = $container.innerHeight() - $item.outerHeight();
+            var container_width = $container.innerWidth();
+            var container_height = $container.innerHeight();
+
             // We need to get the item position relative to the $container.
             var item_offset = $item.offset();
             var container_offset = $container.offset();
@@ -1547,11 +1548,16 @@ function DragAndDropBlock(runtime, element, configuration) {
                     var dx = x - drag_origin.x;
                     var dy = y - drag_origin.y;
 
+                    if (typeof item.max_left === 'undefined') {
+                        var $draggedItem = $container.find('.option[data-value=' + item_id + ']');
+                        item.max_left =  container_width - $draggedItem.outerWidth();
+                        item.max_top = container_height - $draggedItem.outerHeight();
+                    }
                     var left = original_position.left + dx;
                     var top = original_position.top + dy;
-                    left = Math.max(0, Math.min(max_left, left));
-                    top = Math.max(0, Math.min(max_top, top));
-                    item.drag_position = {left: left, top: top};
+                    left = Math.max(0, Math.min(item.max_left, left));
+                    top = Math.max(0, Math.min(item.max_top, top));
+                    item.drag_position = {left: left, top: top};;
                     applyState();
                     raf_id = null;
                 });
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ def package_data(pkg, root_list):
 
 setup(
     name='xblock-drag-and-drop-v2',
-    version='2.2.2',
+    version='2.2.3',
     description='XBlock - Drag-and-Drop v2',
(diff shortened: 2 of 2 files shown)


edx-webhook:
Thanks for the pull request, @xirdneh! I've created [OSPR-3286](https://openedx.atlassian.net/browse/OSPR-3286) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1547,11 +1548,16 @@ function DragAndDropBlock(runtime, element, configuration) {
                     var dx = x - drag_origin.x;
                     var dy = y - drag_origin.y;
 
+                    if (typeof item.max_left === 'undefined') {
```
xitij2000: Not sure how much this optimisation gets us since it only in use if an item is used multiple times. I think you can just directly set max_left and max_top above instead of setting them on item.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1547,11 +1548,16 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+                        var $item = $container.find('.option[data-value=' + item_id + ']');
```
xitij2000: $item is already a variable and this will shadow that, perhaps call this $dragged_item.

xitij2000:
@xirdneh  Also please update the PR title to describe the issue being fixed.

natabene:
@xirdneh Will you need an edX review for this PR?

xitij2000:
@natabene Yes, we would be needing an edX review.

natabene:
@xitij2000 Ok, then please let me know once it is ready for our review.

xitij2000:
Certainly.

drag_and_drop_v2/public/js/drag_and_drop.js
xirdneh: The thing is that this code is executed with every move event. Which happens a lot of times while an item is being moved and we only need to calculate the limits once.
Also, we can't calculate the limits on `onDragStart` because the actual div that's dragged is not created yet. It's rendered when `onDragMove` is fired.

xirdneh:
@natabene This is ready for review. Mind taking a look?
Thanks!

natabene:
@ormsbee Can you review this when you have a chance?

ormsbee:
I'll tag a new version as soon as #206 merges

ormsbee:
Okay, I merged the Changelog change and tagged v2.2.3

xitij2000:
@ormsbee Thanks for the merge. Sorry about missing the changelog update. Will be more careful in the future.

ormsbee:
@xitij2000: No worries, I missed it too. 😛 Thank you.

PR #205, opened 2019-04-04
MCKIN-10393 Fix bug affecting dragging on mobile

xirdneh:
When having a large draggable item the limits are not calculated correctly and the xblock doesn't let the user to drag the item all the way to the right. This happens because the limit is calculated once when the drag is [initiated](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/public/js/drag_and_drop.js#L1465). But when the item is moved the actual div that gets moved is a different one created when an item is marked as [dragged](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/public/js/drag_and_drop.js#L175) and a different [style](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/122da26ee82dc4f4873e2172b0a618c50a5883c4/drag_and_drop_v2/public/css/drag_and_drop.css#L112) is applied to it making the width only 30% of the original. This makes the limit be wrong. The solution is to calculate the limits by using the actual div that's being dragged.
This PR calculates the limits correctly and it only does it once by using a sort of memoization for performance.

**Screenshots**: 
Wrong limits calculated in mobile view
dnd-before

Behavior with the updates in this PR
dnd-after

**Testing instructions**:

1. Crate a Drag and Drop problem.
2. Add an item with large text.
3. Try dragging the item using a mobile view.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ install:
     - "pip install selenium==2.53.0"
     - "pip uninstall -y xblock-drag-and-drop-v2"
     - "python setup.py sdist"
-    - "pip install dist/xblock-drag-and-drop-v2-2.2.0.tar.gz"
+    - "pip install dist/xblock-drag-and-drop-v2-2.2.1.tar.gz"
 script:
     - pep8 drag_and_drop_v2 tests --max-line-length=120
     - pylint drag_and_drop_v2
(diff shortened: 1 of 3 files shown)


edx-webhook:
Thanks for the pull request, @xirdneh! I've created [OSPR-3290](https://openedx.atlassian.net/browse/OSPR-3290) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

xirdneh:
@xitij2000 This is the cherry picked PR.

PR #217, opened 2019-07-11
add subsection due date check for DnD Xblock

DawoudSheraz:
### Description
Drag & Drop is one of the important problem types that are in use in edx-platform. Despite it being an essential part of the platform, some important features are still missing from it. One of such features is subsection due date check for the Assessment mode. For an instructor-paced course, a DnD problem with Assessment mode can accept submissions after the subsection, in which the DnD problem exists, has passed its due date. This poses a problem importantly in the cases when the DnD problem is part of an exam and the submissions made after the exam has passed can alter the grades. Such incidents consequentially decrease the trust of the course teams in this problem type. This PR adds the due dates checks to avoid submissions after the due date.

### Problem Resolution
To make the decision for enabling/disabling submit button with correspondence to the pacing and due date, the following pieces of information were required:

- Subsection due date
- Course Pacing
- Grace period

None of these fields is available in **DragAndDropXblock**. The first notion was to add these fields directly in the Xblock and have them populated when the problem will be created on the Studio. This, however, was a risky move since all the DnD problems in the production would have been affected by the added fields and would have required a re-publish to get populated with the correct data. For the context, the following is an example of how a DnD problem is stored in the mongo:

`{ "_id" : ObjectId("5bf266a6fdf956134ac2565e"), "fields" : { "data" : { "zones" : [ { "uid" : "top", "title" : "The Top Zone", "align" : "center", "height" : 178, "width" : 196, "y" : 30, "x" : 160, "description" : "Use this zone to associate an item with the top layer of the triangle." }, { "uid" : "middle", "title" : "The Middle Zone", "align" : "center", "height" : 138, "width" : 340, "y" : 210, "x" : 86, "description" : "Use this zone to associate an item with the middle layer of the triangle." }, { "uid" : "bottom", "title" : "The Bottom Zone", "align" : "center", "height" : 135, "width" : 485, "y" : 350, "x" : 15, "description" : "Use this zone to associate an item with the bottom layer of the triangle." } ], "items" : [ { "displayName" : "Goes to the top", "feedback" : { "incorrect" : "No, this item does not belong here. Try again.", "correct" : "Correct! This one belongs to The Top Zone." }, "imageURL" : "", "imageDescription" : "", "zones" : [ "top" ], "id" : 0 }, { "displayName" : "Goes to the middle", "feedback" : { "incorrect" : "No, this item does not belong here. Try again.", "correct" : "Correct! This one belongs to The Middle Zone." }, "imageURL" : "", "imageDescription" : "", "zones" : [ "middle" ], "id" : 1 }, { "displayName" : "Goes to the bottom", "feedback" : { "incorrect" : "No, this item does not belong here. Try again.", "correct" : "Correct! This one belongs to The Bottom Zone." }, "imageURL" : "", "imageDescription" : "", "zones" : [ "bottom" ], "id" : 2 }, { "displayName" : "Goes anywhere", "feedback" : { "incorrect" : "", "correct" : "Of course it goes here! It goes anywhere!" }, "imageURL" : "", "imageDescription" : "", "zones" : [ "top", "middle", "bottom" ], "id" : 3 }, { "displayName" : "I don't belong anywhere", "feedback" : { "incorrect" : "You silly, there are no zones for this one.", "correct" : "" }, "imageURL" : "", "imageDescription" : "", "zones" : [ ], "id" : 4 } ], "targetImgDescription" : "An isosceles triangle with three layers of similar height. It is shown upright, so the widest layer is located at the bottom, and the narrowest layer is located at the top.", "feedback" : { "start" : "Drag the items onto the image above.", "finish" : "Good work! You have completed this drag and drop problem." }, "targetImg" : "/xblock/resource/drag-and-drop-v2/public/img/triangle.png" } }, "edit_info" : { "edited_on" : ISODate("2018-11-19T07:30:46.050Z"), "edited_by" : 2, "original_version" : ObjectId("5bf26695fdf956134ac2565c"), "previous_version" : ObjectId("5bf26695fdf956134ac2565c") }, "block_type" : "drag-and-drop-v2", "schema_version" : 1 }`

This was intriguing as if a DnD problem was viewed from **Staff Debug Info** on the LMS, it showed the information about the due date, pacing and grace period. This led to viewing how the other problems are stored in the mongo. Here is an example of a random problem stored in the Mongo:

`{ "_id" : ObjectId("5c0625a6fdf9560726d4681b"), "fields" : { "data" : "<problem>\n  <multiplechoiceresponse>\n    <label>Which of these statements about rods and cones is true?</label>\n    <choicegroup type="MultipleChoice">\n      <choice correct="true">Rods respond best in a dimly lit environment.</choice>\n      <choice correct="false">Rods are for high light conditions; cones are for low light conditions.</choice>\n      <choice correct="false">The human retina has more cones than rods.</choice>\n      <choice correct="false">Rods come in three types: S, M, and L.</choice>\n    </choicegroup>\n    <solution>\n      <div class="detailed-solution">\n        <p>Explanation</p>\n        <p>Rods respond best in dim light, whereas cones respond best in bright light and to color. There are around 20 times more rods in the retina than there are cones. Cones come in three flavors: S, M, and L.</p>\n      </div>\n    </solution>\n  </multiplechoiceresponse>\n</problem>\n" }, "edit_info" : { "edited_on" : ISODate("2018-12-04T06:58:46.981Z"), "edited_by" : NumberLong(2), "original_version" : ObjectId("5c0625a6fdf9560726d4681b"), "previous_version" : null }, "block_type" : "problem", "schema_version" : 1 }`

As evident from the data, the regular problem types also don't save the due date, pacing and grace period information. This information is embedded at the run time through the help of Mixins. Looking at the DnD xblock created on the studio/LMS side, it is not **DragAndDropXblock**, but **DragAndDropXblockWithMixins**. The xblock created by the studio contained all the necessary information. Based on this information, the idea of adding new fields in the DnDXblock was dropped as this would duplicate the information. The fix in this PR relies on the information coming from the Mixins and no data duplication will happen. 

### Test
 - [ ] Integration tests
 - [x] Unit Tests

### Post Review
 - [ ] Squash & Rebase commits

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -1073,19 +1073,17 @@ class DragAndDropBlock(
         Return the date submissions should be closed from.
 
         Calculates and return the due date after which the submissions
-        will no longer be accepted. Returns None for the AttributeError
-        which happens for pure DragAndDropXblock as due date & graceperiod
-        are not its attributes. They are present for DragAndDropXblockWithMixins
+        will no longer be accepted. Returns None if both due & graceperiod
+        are missing which happens for pure DragAndDropXblock as they are
+        not its attributes. They are present for DragAndDropXblockWithMixins
         """
-        try:
-            due_date = self.due
-
-            if self.graceperiod is not None and due_date:
-                return due_date + self.graceperiod
+        if hasattr(self, 'due') and hasattr(self, 'graceperiod'):
+            due_date = self.due            # pylint: disable=no-member
+            if self.graceperiod is not None and due_date:          # pylint: disable=no-member
+                return due_date + self.graceperiod                 # pylint: disable=no-member
             else:
                 return due_date
-
-        except AttributeError:
+        else:
             return None
 
     def is_self_paced(self):
@@ -1093,15 +1091,14 @@ class DragAndDropBlock(
         Returns if the course is self-paced or not.
 
         Returns the boolean indicating if the parent course is
-        self-paced or not. Returns False for the AttributeError
+        self-paced or not. Returns False if the attribute is missing
         which happens for pure DragAndDropXblock. self_paced attribute is
         present only for DragAndDropXblockWithMixins.
         """
-        try:
-            return self.self_paced
-        except AttributeError:
+        if hasattr(self, 'self_paced'):
+            return self.self_paced           # pylint: disable=no-member
+        else:
             return False
-
 
     @staticmethod
     def workbench_scenarios():
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -356,7 +356,7 @@ class DragAndDropBlock(
             "item_background_color": self.item_background_color or None,
             "item_text_color": self.item_text_color or None,
             "due": self.calculate_due_date(),
-            "self_paced":self.self_paced,
+            "self_paced": self.is_self_paced(),
             # final feedback (data.feedback.finish) is not included - it may give away answers.
         }
 
@@ -1071,13 +1071,37 @@ class DragAndDropBlock(
     def calculate_due_date(self):
         """
         Return the date submissions should be closed from.
-        """
-        due_date = self.due
 
-        if self.graceperiod is not None and due_date:
-            return due_date + self.graceperiod
-        else:
-            return due_date
+        Calculates and return the due date after which the submissions
+        will no longer be accepted. Returns None for the AttributeError
+        which happens for pure DragAndDropXblock as due date & graceperiod
+        are not its attributes. They are present for DragAndDropXblockWithMixins
+        """
+        try:
+            due_date = self.due
+
+            if self.graceperiod is not None and due_date:
+                return due_date + self.graceperiod
+            else:
+                return due_date
+
+        except AttributeError:
+            return None
+
+    def is_self_paced(self):
+        """
+        Returns if the course is self-paced or not.
+
+        Returns the boolean indicating if the parent course is
+        self-paced or not. Returns False for the AttributeError
+        which happens for pure DragAndDropXblock. self_paced attribute is
+        present only for DragAndDropXblockWithMixins.
+        """
+        try:
+            return self.self_paced
+        except AttributeError:
+            return False
+
 
     @staticmethod
     def workbench_scenarios():
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -355,6 +355,8 @@ class DragAndDropBlock(
             "target_img_description": self.target_img_description,
             "item_background_color": self.item_background_color or None,
             "item_text_color": self.item_text_color or None,
+            "due": self.calculate_due_date(),
+            "self_paced":self.self_paced,
             # final feedback (data.feedback.finish) is not included - it may give away answers.
         }
 
@@ -1065,6 +1067,17 @@ class DragAndDropBlock(
             bool: True if current answer is correct
         """
         return self._answer_correctness() == self.SOLUTION_CORRECT
+
+    def calculate_due_date(self):
+        """
+        Return the date submissions should be closed from.
+        """
+        due_date = self.due
+
+        if self.graceperiod is not None and due_date:
+            return due_date + self.graceperiod
+        else:
+            return due_date
 
     @staticmethod
     def workbench_scenarios():
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -10,7 +10,7 @@ import logging
 import urllib
 import webob
 import pkg_resources
-from django.utils import translation
+from django.utils import timezone, translation
 from xblock.core import XBlock
 from xblock.exceptions import JsonHandlerError
 from xblock.fields import Scope, String, Dict, Float, Boolean, Integer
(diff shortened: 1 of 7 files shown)


drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1934,6 +1943,8 @@ function DragAndDropBlock(runtime, element, configuration) {
             display_zone_labels: configuration.display_zone_labels,
             display_zone_borders: configuration.display_zone_borders,
             zones: configuration.zones,
+            due: configuration.due,
+            self_paced: configuration.self_paced,
```
DawoudSheraz: these lines are to be removed as they are not doing anything.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -1066,6 +1068,38 @@ def _is_answer_correct(self):
...
         return self._answer_correctness() == self.SOLUTION_CORRECT
 
+    def calculate_due_date(self):
+        """
+        Return the date submissions should be closed from.
+
+        Calculates and return the due date after which the submissions
+        will no longer be accepted. Returns None if both due & graceperiod
+        are missing which happens for pure DragAndDropXblock as they are
+        not its attributes. They are present for DragAndDropXblockWithMixins
```
bradenmacdonald: Please say specifically that the `due` and `graceperiod` fields get added by [`InheritanceMixin`](https://github.com/edx/edx-platform/blob/7eede0926787f3d991916e28da2b4651845e472e/common/lib/xmodule/xmodule/modulestore/inheritance.py#L31-L74).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -1066,6 +1068,38 @@ def _is_answer_correct(self):
...
+        """
+        if hasattr(self, 'due') and hasattr(self, 'graceperiod'):
+            due_date = self.due            # pylint: disable=no-member
+            if self.graceperiod is not None and due_date:          # pylint: disable=no-member
+                return due_date + self.graceperiod                 # pylint: disable=no-member
+            else:
+                return due_date
+        else:
+            return None
```
bradenmacdonald: It seems wrong that this sort of logic must be duplicated in every XBlock. What do you think about instead expanding and moving this `calculate_due_date()` method to `InheritanceMixin` (check `self_paced`, `due`, and `graceperiod`, and produce either a final due date or `None` if not applicable)?

Then this PR would be reduced to a few lines:

```
def student_view_data(self, context=None):
    ...
    "submit_deadline": self.calculate_due_date(),
    ...
```

and

```
    var problemNotDue = function () {
        if(configuration.submit_deadline === null){
            return true;
        }
        var now = new Date();
        var due_date = new Date(configuration.submit_deadline);
        return (now.getTime() < due_date.getTime());
    };
```

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1827,7 +1827,7 @@ function DragAndDropBlock(runtime, element, configuration) {
     };
 
     var canSubmitAttempt = function() {
-        return Object.keys(state.items).length > 0 && attemptsRemain() && !submittingLocation();
+        return Object.keys(state.items).length > 0 && problemNotDue() && attemptsRemain() && !submittingLocation();
```
bradenmacdonald: Shouldn't this be enforced on the server as well? Otherwise students can just change their local clock and then submit, or run `const pastTime = (new Date).getTime() - 1e12; Date.prototype.getTime = () => pastTime;` in the browser console and then submit.

DawoudSheraz: Nice suggestion. I am thinking of moving all this logic to the server and just provide with a flag on the JS side.

drag_and_drop_v2/drag_and_drop_v2.py
DawoudSheraz: I will have to consult other members of my team for this decision. Even though it makes sense to have this common function in the InheritanceMixin, we need to be sure about the consequences.

drag_and_drop_v2/public/js/drag_and_drop.js
DawoudSheraz: @bradenmacdonald The problem with user changing JS param is solely for the due dates check, but wherever we are making checks, user can do the manipulation to enable/disable certain buttons. For example, I was able to manipulate the max attempts and enable the `Show Answer` button. So, even if I do all the checks on the server side, any user can change that flag using browser consoles. For other problems, we are doing such checks in the python and rendering HTML based on the context. Based on what I have tested at the moment, it is very easy for a user to manipulate and change the behavior of the DnD component. 

dnd_attempts_bug

tests/unit/test_basics.py
```diff
@@ -269,3 +279,23 @@ def test_image_url(self):
...
+        """
+        Verifies the dealine passed boolean value w.r.t pacing and due date.
+
+        Given the pacing information, due date and graceperiod,
+        confirm if the submission deadline has passed or not.
+        """
+        self.add_submission_deadline_information(due_date, graceperiod, self_paced)
+        self.assertEqual(is_submission_due, self.block.student_view_data()['has_deadline_passed'])
+
+
```
DawoudSheraz: nit: to remove extra line

tests/unit/test_basics.py
```diff
@@ -269,3 +279,21 @@ def test_image_url(self):
...
+    @ddt.data(
+        (False, now(), None, True),
+        (True, now(), None, False),
+        (False, now(), timedelta(days=1), False),
+        (True, now(), timedelta(days=1), False),
+        (False, now() - timedelta(hours=1), None, True),
+    )
+    @ddt.unpack
+    def test_submission_deadline(self, self_paced, due_date, graceperiod, is_submission_due):
```
DawoudSheraz: nit: to refactor `is_submission_due` to some other name as it seems a bit confusing at first glance.

DawoudSheraz:
Closing this PR in favor of https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/219

PR #225, opened 2019-09-19
BOM-470

awais786:
Updating xblock external libs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,20 @@ language: python
 sudo: false
 python:
     - "2.7"
-before_install:
-    - "export DISPLAY=:99"
-    - "sh -e /etc/init.d/xvfb start"
+services:
+    - xvfb
 install:
     - "sh install_test_deps.sh"
     # Remove the next two lines after our test suite is compatible with selenium 3.
     - "pip uninstall -y selenium"
-    - "pip install selenium==2.53.0"
+    - "pip install selenium==2.53.6"
     - "pip uninstall -y xblock-drag-and-drop-v2"
     - "python setup.py sdist"
     - "pip install dist/xblock-drag-and-drop-v2-*.tar.gz"
+    - "pip uninstall -y pylint"
+    - "pip install pylint==1.7.6"
 script:
-    - pep8 drag_and_drop_v2 tests --max-line-length=120
+    - pycodestyle drag_and_drop_v2 tests --max-line-length=120
     - pylint drag_and_drop_v2
     - pylint tests --rcfile=tests/pylintrc
     - python run_tests.py
--- a/install_test_deps.sh
+++ b/install_test_deps.sh
@@ -1,6 +1,6 @@
 # Installs xblock-sdk and dependencies needed to run the tests suite.
 # Run this script inside a fresh virtual environment.
-pip install -e git://github.com/edx/xblock-sdk.git@v0.1.3#egg=xblock-sdk==v0.1.3
+pip install -e git://github.com/edx/xblock-sdk.git@3b64219024d111f247a14e267cd5eaee72aefcd8#egg=xblock-sdk
 cd $VIRTUAL_ENV/src/xblock-sdk/ && pip install -r requirements/base.txt \
                                 && pip install -r requirements/test.txt && cd -
 pip install -r requirements.txt
--- a/pylintrc
+++ b/pylintrc
@@ -12,7 +12,9 @@ disable=
     too-many-instance-attributes,
     too-few-public-methods,
     too-many-public-methods,
-    unused-argument
+    unused-argument,
+    unsubscriptable-object,
+    no-else-return
 
 [SIMILARITIES]
 min-similarity-lines=4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
 mako==1.0.7
 edx-i18n-tools==0.4.7
+pycodestyle==2.5.0
 django-statici18n==1.8.0
 transifex-client==0.13.4
-git+https://github.com/edx/xblock-utils.git@v1.1.1#egg=xblock-utils==1.1.1
-git+https://github.com/edx/XBlock.git@xblock-0.5.0#egg=XBlock==0.5.0
+git+https://github.com/edx/xblock-utils.git@v1.2.2#egg=xblock-utils==1.2.2
+XBlock==1.2.6
 -e .
--- a/tests/integration/test_studio.py
+++ b/tests/integration/test_studio.py
@@ -114,9 +114,6 @@ class TestStudio(StudioEditableBaseTest):
         Verify user can provide a custom background image URL.
         """
         default_bg_img_src = 'http://localhost:8081/resource/drag-and-drop-v2/public/img/triangle.png'
-        # In order to use a working image and avoid load errors, we use the default image with a custom
-        # query string
-        custom_bg_img_src = '{}?my-custom-image=true'.format(default_bg_img_src)
 
         self.load_scenario()
         # Go to zones tab.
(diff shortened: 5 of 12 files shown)


install_test_deps.sh
```diff
@@ -1,6 +1,6 @@
 # Installs xblock-sdk and dependencies needed to run the tests suite.
 # Run this script inside a fresh virtual environment.
-pip install -e git://github.com/edx/xblock-sdk.git@v0.1.3#egg=xblock-sdk==v0.1.3
+pip install -e git://github.com/edx/xblock-sdk.git@master#egg=xblock-sdk==vmaster
```
awais786: 0.1.3 has iteritems and its failing on python3.
[xblock-sdk==v0.1.3](https://github.com/edx/xblock-sdk/blob/0c987356cba5c5a7d47c98654e030ee4c3cb47c2/workbench/runtime.py#L258)

requirements.txt
```diff
@@ -1,7 +1,8 @@
 mako==1.0.7
 edx-i18n-tools==0.4.7
 django-statici18n==1.8.0
+six==1.12.0
 transifex-client==0.13.4
-git+https://github.com/edx/xblock-utils.git@v1.1.1#egg=xblock-utils==1.1.1
-git+https://github.com/edx/XBlock.git@xblock-0.5.0#egg=XBlock==0.5.0
+git+https://github.com/edx/xblock-utils.git@master#egg=master
+git+https://github.com/edx/XBlock.git@xblock-1.2.4#egg=XBlock==1.2.4
```
awais786: 0.5.0 is very old release and latest release is python3 compatible.

install_test_deps.sh
feanil: Use a hash instead of master I think.  Also you don't need the `==vmaster`

tests/unit/test_basics.py
```diff
@@ -16,6 +18,8 @@ class BasicTests(TestCaseMixin, unittest.TestCase):
 
     def setUp(self):
         self.block = make_block()
+        self.block.unmixed_class = mock.Mock()
```
awais786:
```
ves/xblock-drag-and-drop-v2_master_py2/src/xblock-sdk/workbench/services.py", line 28, in get_settings_bucket
    xblock_settings_bucket = getattr(block, self.xblock_settings_bucket_selector, block.unmixed_class.__name__)
AttributeError: 'DragAndDropBlock' object has no attribute 'unmixed_class'
```
Fixed this test now by mocking.

.travis.yml
```diff
@@ -4,7 +4,8 @@ python:
     - "2.7"
 before_install:
     - "export DISPLAY=:99"
```
feanil: You should be able to remove this `before_install` clause as well because the xvfb service will set this.

awais786:
@feanil Previous PR became too big and hundreds of tests failing in python3. This repo has old external libs link which needs to update. After this merge I will do the python3 up-gradation.

pylintrc
```diff
@@ -12,7 +12,9 @@ disable=
     too-many-instance-attributes,
     too-few-public-methods,
     too-many-public-methods,
-    unused-argument
+    unused-argument,
```
awais786: We will fix some of pylint issues in next PR.

tests/pylintrc
```diff
@@ -15,7 +15,18 @@ disable=
     too-many-public-methods,
     invalid-name,
     no-member,
-    star-args
+    star-args,
```
awais786: We will fix some of pylint issues in next PR.

.travis.yml
```diff
@@ -13,8 +12,10 @@ install:
     - "pip uninstall -y xblock-drag-and-drop-v2"
     - "python setup.py sdist"
     - "pip install dist/xblock-drag-and-drop-v2-*.tar.gz"
+    - "pip uninstall -y pylint"
+    - "pip install pylint==1.7.6"
 script:
-    - pep8 drag_and_drop_v2 tests --max-line-length=120
+    - pycodestyle drag_and_drop_v2 tests --max-line-length=120
```
awais786: pep8 is not working any more.

tests/integration/test_studio.py
```diff
@@ -127,7 +124,12 @@ def test_custom_image(self):
         self.assertFalse(radio_buttons['auto'].is_selected())
         url_field = self.background_image_url_field
         self.assertEqual(url_field.get_attribute('value'), '')
-        self.assertEqual(self.target_preview_img.get_attribute('src'), default_bg_img_src)
+        self.assertIn(
+            default_bg_img_src.split('http://localhost:8081/')[1], self.target_preview_img.get_attribute('src')
```
awais786: Here both links contains different ports. Equal was not working here.

tests/integration/test_studio.py
```diff
@@ -127,7 +124,12 @@ def test_custom_image(self):
...
+        )
+
+        custom_bg_img_src = '{}?my-custom-image=true'.format(self.target_preview_img.get_attribute('src'))
```
awais786: Due to port issue link was broken here.

awais786:
@feanil You can review this.

tests/integration/test_studio.py
awais786: Following was the issue.
```
(Pdb) self.target_preview_img.get_attribute('src')
u'http://localhost:59274/resource/drag-and-drop-v2/public/img/triangle.png'
(Pdb) default_bg_img_src
'http://localhost:8081/resource/drag-and-drop-v2/public/img/triangle.png'
```

requirements.txt
```diff
@@ -1,7 +1,8 @@
 mako==1.0.7
 edx-i18n-tools==0.4.7
+pycodestyle==2.5.0
 django-statici18n==1.8.0
 transifex-client==0.13.4
-git+https://github.com/edx/xblock-utils.git@v1.1.1#egg=xblock-utils==1.1.1
-git+https://github.com/edx/XBlock.git@xblock-0.5.0#egg=XBlock==0.5.0
+git+https://github.com/edx/xblock-utils.git@master#egg=master
```
ihtram: @awais786 Can we create a release tag here instead of master?
cc: @feanil

tests/integration/test_studio.py
ihtram: @feanil I am not sure the reason behind the change of port. The above piece of code will fix the broken test as we are not comparing ports anymore. But if you want us to find out the root cause of this failure, please create a YONK ticket. 
cc: @awais786

awais786: @ihtram thanks for review. I will wait for @feanil's answer before taking further action.

requirements.txt
feanil: +1, we should release a new version of xblock sdk and refer to the has of that tagged version.

requirements.txt
feanil: You can pullthis from PyPI now instead of github. https://pypi.org/project/XBlock/

tests/integration/test_studio.py
feanil: @ihtram I think the issue is probably with how the test system is started up and it no longer having a default port.  I think this change is good and we don't need to dig much deeper.  I'm not sure the value of fixing it is worth the time.

awais786:
@feanil @ihtram I have made all changes. But I am getting some warning.
```
the future in order to support XBlockAsides
  warnings.warn("IdGenerator will be required in the future in order to support XBlockAsides", FutureWarning)
................................................................................................WARNING:root:XBlock <class 'xblock.internal.DragAndDropBlockWithMixins'> does not contain field url_name
WARNING:django.request:Not Found: /favicon.ico
```

feanil:
I think that warning is fine for now, though it will have to be fixed before this xblock can support XBlockAsides.

PR #229, opened 2019-10-08
BOM-470

awais786:
Bump the version.
setup.py was using older version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ def package_data(pkg, root_list):
 
 setup(
     name='xblock-drag-and-drop-v2',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -23,7 +23,7 @@ def package_data(pkg, root_list):
 
 setup(
     name='xblock-drag-and-drop-v2',
-    version='2.2.4',
+    version='2.2.6',
```
Ayub-Khan: why skip one ?

awais786: I have released 2.2.5 in last commit but missed to update this file.

PR #236, opened 2020-01-09
Render zone title as HTML

AhtishamShahid:
### TLDR

This PR resolves the following issue.

The preview generated within the component editor accepts HTML tags and entities (like, for example, the `<sup>` tag) and displays them properly, but the display within Studio and the LMS shows them as raw text instead.

In the editor, the title is rendered properly as HTML  as in this image 
Screenshot 2020-01-09 at 11 32 39 AM

Whereas on studio page it is rendered as text as in this image 

Screenshot 2020-01-09 at 11 31 57 AM

### FIX
This is fixed by setting the title as  Innerhtml of `<p>` tag.

### TEST
- Add Drag and Drop block in studio.
- Edit the block by clicking the edit button.
- The modal will open, click continue button on this modal.
- on this page select ` Generate image automatically` checkbox.
- scroll down and in zone definitions fill-up the form and add title ` MW (g &#183; mol<sup>-1</sup>)` and click save .

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -262,9 +262,11 @@ function DragAndDropTemplates(configuration) {
                 [
                     h(
                         'p',
-                        { className: className },
+                        {
+                            className: className,
+                            innerHTML: gettext(zone.title)
+                        },
                         [
-                            gettext(zone.title),
                             h('span.sr', gettext(', dropzone'))
                         ]
                     ),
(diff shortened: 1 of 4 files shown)


AhtishamShahid:
> Changes look good. Before approving, please check the following two things:
> 
> 1. Run the tests a few times to ensure this change didn't introduce any flakiness
> 2. Check the output of the zone using a screenreader with & without your fix. Just want to make sure that moving the title value position doesn't affect the a11y.
 
The screen reader is not working in both cases, and all tests are passing in all attempts I have made.
But in one instance `test_scrolling_during_placement` failed on Jenkins due to unknown reasons but on restarting test job it eventually passed without any change in code. And I am unable to reproduce this again on local and on Jenkins.

DawoudSheraz:
@AhtishamShahid Good to go. After the merge, update the version in the platform.

PR #238, opened 2020-04-13
fix: Properly limit 'feedback.opened' events

davidjoy:
Only send ‘feedback.opened’ if there are elements in feedback array.

See https://openedx.atlassian.net/browse/TNL-7184 for complete details and more background.

Long story short, if a drag and drop item doesn't have a success or error feedback message, the backend sends an empty array of feedback, which the frontend wasn't consuming properly.  This PR allows the frontend to interpret an empty array as not worthy of sending a feedback.opened event.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,37 @@
 Drag and Drop XBlock changelog
 ==============================
 
+Version 2.2.10 (2020-04-14)
+---------------------------
+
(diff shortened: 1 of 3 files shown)


ormsbee:
Please bump the patch version in setup.py and add this to the Changelog.md file.

davidjoy:
@ormsbee Done - thanks for the heads up, didn't realize that needed to be done.  I also backfilled Changelog.md with a bunch of missing versions.

PR #244, opened 2020-06-16
[TNL-6993] [BD-29]: Make Drag and Drop indexable.

ahmedbilal:
Make Drag and Drop indexable

**JIRA tickets**: 
* https://openedx.atlassian.net/browse/TNL-6993

**Merge deadline**: "None"

**Testing instructions**:

1. Create a new Drag and Drop in demo course or any course (it doesn't really matter).
2. Give it some *title* and *problem text*
3. Enroll the student into the course for which you just created the Drag and Drop.
4. On course page, use the search facility to find about the added drag and drop turn-by-turn using both the *title* and some string in the *problem text*.
5. If it appears, the functionality works as expected, if not there might be some issue in it.

**Reviewers**
- [ ] @xitij2000
- [ ] @davestgermain

### DIFF
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -16,6 +16,7 @@ import six.moves.urllib.error  # pylint: disable=import-error
 import six.moves.urllib.parse  # pylint: disable=import-error
 import six.moves.urllib.request  # pylint: disable=import-error
 import webob
+
 from django.utils import translation
 from xblock.core import XBlock
 from xblock.exceptions import JsonHandlerError
@@ -28,7 +29,7 @@ from xblockutils.settings import ThemableXBlockMixin, XBlockWithSettingsMixin
 from .default_data import DEFAULT_DATA
 from .utils import (
     Constants, DummyTranslationService, FeedbackMessage,
-    FeedbackMessages, ItemStats, StateMigration, _
+    FeedbackMessages, ItemStats, StateMigration, _clean_data, _
 )
 
 # Globals ###########################################################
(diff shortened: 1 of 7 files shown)


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -1119,3 +1120,61 @@ def workbench_scenarios():
                 "<vertical_demo><drag-and-drop-v2 mode='assessment' max_attempts='3'/></vertical_demo>"
             ),
         ]
+
+    def _strip_html_tags_and_extra_whitespaces(self, raw_html):
+        cleanr = re.compile('<.*?>')
+        cleantext = str(re.sub(cleanr, '', raw_html))
+        cleantext = " ".join(re.split(r"\s+", cleantext, flags=re.UNICODE)).strip()
+        return cleantext
```
xitij2000, line 1128: Instead of this, please use `bleach`, then you can simply call `bleach.clean(raw_html)`. 
You will also need to add it to the requirements file and setup.py. 

It's already installed with edx-platform so it's OK to have this as a dependency.

ahmedbilal: Thanks :+1:

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -1119,3 +1120,61 @@ def workbench_scenarios():
...
+
+        index_body = {
+            "display_name": self.display_name,
+            "question_text": self._strip_html_tags_and_extra_whitespaces(self.question_text),
+            "background_image_description": self.data.get("targetImgDescription", ""),
+            **items_display_names,
+            **items_image_description,
+            **zones_display_names,
+            **zones_description
+        }
```
xitij2000, line 1171: Actually, could you go back to using the `.update` syntax? This might be more concise, but it breaks on Python 2.

ahmedbilal: ofcourse

setup.py
```diff
@@ -23,7 +23,7 @@ def package_data(pkg, root_list):
 
 setup(
     name='xblock-drag-and-drop-v2',
-    version='2.2.10',
+    version='2.2.11',
     description='XBlock - Drag-and-Drop v2',
     packages=['drag_and_drop_v2'],
     install_requires=[
```
davestgermain, line 29: Does `bleach` need to be added to `install_requires`?

xitij2000: > Does `bleach` need to be added to `install_requires`?

It is now imported by the block so makes sense to add here?

ahmedbilal: It is already in requirements.txt but I now also added it into setup.py

ahmedbilal: @davestgermain @xitij2000  It is already in requirements.txt but I now also added it into setup.py

ahmedbilal:
@davestgermain Ping. I implemented your suggestion, It is good to go.

ahmedbilal:
@davestgermain Ping. Is this also just need commit squashing?

ahmedbilal:
@davestgermain Done. Squashed.

ahmedbilal:
@natabene Can you please assign some reviewer to it?

natabene:
@marcotuts Can you assist with reviewing this? Not sure who is in charge of BD-29.

davestgermain:
@ahmedbilal The tests are failing under python 2.7. Could you remove 2.7 from [.travis.yml](https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/.travis.yml)? And perhaps change the version of this XBlock to 2.3, since dropping support for python 2.7 is a somewhat major revision.

ahmedbilal:
@davestgermain Ready. I drop Python 2 support and bumped the minor version.

ahmedbilal:
@davestgermain I think you forgot to merge it.

xitij2000:
@ahmedbilal I can merge it. That's OK.

PR #247, opened 2020-07-14
BB-2681 Fix YTP XBlocks Translations RTL fix

taniwha:
This PR bring is the right-to-left fix from the raccoongang repository

**JIRA tickets**: https://tasks.opencraft.com/browse/BB-2681

**Discussions**: N/A

**Dependencies**: None

**Screenshots**: N/A

**Sandbox URL**: N/A

**Merge deadline**: "None"

**Testing instructions**:

1. install as per usual.
2. in studio add compoenent* problem -> advanced -> drag and drop. It may be a good idea to mare all instructor-input text to for clarity as to what is and is not translated.
3. publish
4. in lms, go to unit, ensure drag and drop is working
5.switch to Arabic (ar)]
6.reload drag and drop problem page, check translations: in particular, submit button, feedback messages, keyboard help, attempts remaining

**Author notes and concerns**:
This is just an scss tweak from right-to-left. The testing instructions are probably superfluous.

**Reviewers**
- [ ] @kaizoku 
- [ ] edX reviewer[s] TBD

PR #248, opened 2020-07-14
BB 2681 fix ytp x blocks translations raccoondog translations

taniwha:
This PR brings in the Arabic translates from the gingko-rg and master branches of the raccoongang repository. Note that these are alternative translations to what is already there and not additional translation.

**JIRA tickets**: https://tasks.opencraft.com/browse/BB-2681

**Discussions**: N/A

**Dependencies**: maybe https://github.com/edx-solutions/xblock-drag-and-drop-v2/pull/247

**Screenshots**: N/A

**Sandbox URL**: N/A

**Merge deadline**: None

**Testing instructions**:

1. install as per usual.
2. in studio add compoenent* problem -> advanced -> drag and drop. It may be a good idea to mare all instructor-input text to for clarity as to what is and is not translated.
3. publish
4. in lms, go to unit, ensure drag and drop is working
5. switch to Arabic (ar)]
6. reload drag and drop problem page, check translations: in particular, submit button, feedback messages, keyboard help, attempts remaining

As noted above, these translations are needed in an ironwood installation thus the PR is branched from the v2.1.8 tag. 

**Reviewers**
- [ ] @kaizoku 
- [ ] edX reviewer[s] TBD


kaizoku:
@taniwha, I think we should leave the translations as-is in master, rather than spontaneously changing the strings for all users of drag-and-drop in Arabic.

PR #249, opened 2020-09-10
MCKIN-24630 Arabic Translation correction

musmanmalik:
Translation correction


drag_and_drop_v2/public/js/translations/ar/text.js
```diff
@@ -1,8 +1,4 @@
 
-            (function(global){
-                var DragAndDropI18N = {
```
msaqib52, line 3: any specific reason for removing this namespace?

msaqib52 (changes requested):

drag_and_drop_v2/public/js/translations/ar/text.js
musmanmalik: Added again. Thanks for highlighting.

msaqib52 (approved):

PR #250, opened 2020-09-14
MCKIN-24661: merge master to bring python 3 changes

msaqib52:
This is a merge PR to bring python3 changes from master to `mckinsey-master` branch


ihtram (approved):
@msaqib52 Looks good to me. Just one thing, could you change the target brach name and rename to apros or mock? Also make this change for the other branch created for PY2 support.

msaqib52:
> @msaqib52 Looks good to me. Just one thing, could you change the target brach name and rename to apros or mock? Also make this change for the other branch created for PY2 support.

@ihtram updated the branch names but don't delete the mckinsey-master yet because it is being used for installations

PR #256, opened 2020-10-15
MCKIN-26123 Platform Issues - Translations

wasifarbisoft:


msaqib52 (changes requested):
@wasifarbisoft i don't see any PO/MO files in the PR, how did js files got updated?

wasifarbisoft:
@msaqib52 , in D&D repo there are duplicate folders for languages js especially for the languages that have Japanese, Spanish, Korean, German, Portuguese.(Not sure why there are duplicates there) perhaps the languages that have two word codes like ja-jp have two folders i.e ja-jp and ja_jp. So all the js files were created in the folders that have "_" but the code is referring to  "-".   I have just moved the translations to the correct js files.

msaqib52 (approved):
LGTM 👍

PR #268, opened 2021-03-02
[feat]: allow problem score override

awaisdar001:


drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -209,8 +210,22 @@ class DragAndDropBlock(
         enforce_type=True,
     )
 
+    raw_possible = Float(
+        help=_("Keeps maximum score achieved by student as a raw value between 0 and 1."),
```
DawoudSheraz, line 214: help text is incorrect.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -845,8 +868,8 @@ def _mark_complete_and_publish_grade(self):
             current_raw_earned_is_greater = True
 
         if current_raw_earned is None or current_raw_earned_is_greater:
-            self.raw_earned = current_raw_earned
-            self._publish_grade(Score(self.raw_earned, self.max_score()))
+            self.set_score(Score(current_raw_earned, self.max_score()))
```
DawoudSheraz, line 871: Why self.max_score here?

awaisdar001: will always be 1, its a class method.

drag_and_drop_v2/drag_and_drop_v2.py
awaisdar001: thanks for raising this.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -209,8 +210,22 @@ class DragAndDropBlock(
         enforce_type=True,
     )
 
+    raw_possible = Float(
```
asadazam93, line 213: weight attribute already existed but was not being used? And how was score being recorded before raw_possible attribute?

asadazam93 (approved):

AhtishamShahid (approved):

PR #290, opened 2022-07-21
feat: use platform translations as a fallback when XBlock translations are missing

tecoholic:
This adds a wrapper for `gettext` and `ngettext`, which uses global (platform-wide) translations when a translation is missing in the XBlock.

As a side effect, this allows overriding English text with comprehensive themes (by patching the default translation catalog).

Internal ticket: BB-6465

### DIFF
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -6,22 +6,22 @@ function DragAndDropTemplates(configuration) {
     if ('DragAndDropI18N' in window) {
         // Use DnDv2's local translations
         gettext = function(string) {
-            let translated = window.DragAndDropI18N.gettext(string);
+            var translated = window.DragAndDropI18N.gettext(string);
             // if DnDv2's translation is the same as the input, check if global has a different value
             // This is useful for overriding the XBlock's string by themes (only for English)
             if (string === translated && 'gettext' in window) {
                 translated = window.gettext(string);
             }
             return translated;
-        }
+        };
         ngettext = function(strA, strB, n) {
-            let translated = window.DragAndDropI18N.ngettext(strA, strB, n);
-            let string = n == 1 ? strA : strB;
+            var translated = window.DragAndDropI18N.ngettext(strA, strB, n);
+            var string = n == 1 ? strA : strB;
             if (string === translated && 'gettext' in window) {
                 translated = window.gettext(strA, strB, n);
             }
             return translated;
-        }
+        };
     } else if ('gettext' in window) {
         // Use edxapp's global translations
         gettext = window.gettext;
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -5,8 +5,23 @@ function DragAndDropTemplates(configuration) {
     var ngettext;
     if ('DragAndDropI18N' in window) {
         // Use DnDv2's local translations
-        gettext = window.DragAndDropI18N.gettext;
-        ngettext = window.DragAndDropI18N.ngettext;
+        gettext = function(string) {
+            let translated = window.DragAndDropI18N.gettext(string);
+            // if DnDv2's translation is the same as the input, check if global has a different value
+            // This is useful for overriding the XBlock's string by themes (only for English)
+            if (string === translated && 'gettext' in window) {
+                translated = window.gettext(string);
+            }
+            return translated;
+        }
+        ngettext = function(strA, strB, n) {
+            let translated = window.DragAndDropI18N.ngettext(strA, strB, n);
+            let string = n == 1 ? strA : strB;
+            if (string === translated && 'gettext' in window) {
+                translated = window.gettext(strA, strB, n);
+            }
+            return translated;
+        }
     } else if ('gettext' in window) {
         // Use edxapp's global translations
         gettext = window.gettext;
--- a/drag_and_drop_v2/public/js/drag_and_drop.js
+++ b/drag_and_drop_v2/public/js/drag_and_drop.js
@@ -1,27 +1,22 @@
 function DragAndDropTemplates(configuration) {
     "use strict";
 
-    // Fetch the platform translations if available, or fallback to the
-    // local translations. This allows overriding of translations in themes
-    // by patching the global translations.
-    var gettext = function(string) {
-        return 'gettext' in window ?
-            window.gettext(string) :
-            (
-                'DragAndDropI18N' in window ?
-                window.DragAndDropI18N.gettext(string) :
-                string
-            );
+    var gettext;
+    var ngettext;
+    if ('DragAndDropI18N' in window) {
+        // Use DnDv2's local translations
+        gettext = window.DragAndDropI18N.gettext;
+        ngettext = window.DragAndDropI18N.ngettext;
+    } else if ('gettext' in window) {
+        // Use edxapp's global translations
+        gettext = window.gettext;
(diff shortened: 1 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @tecoholic!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

openedx-webhooks:
Thanks for the pull request, @tecoholic!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

Agrendalath:
@tecoholic, thank you for investigating this! However, I don't think that platform-wide translations should have a higher priority than XBlock-specific ones. What if we made these platform-wide translations a fallback instead?

I have two ideas for implementing this without rebuilding the whole translation system:
1. How are JS translation files (like `drag_and_drop_v2/public/js/translations/en/text.js`) generated? Could we add fallbacks there, e.g. change this:
   ```js
    django.gettext = function(msgid) {
      var value = django.catalog[msgid];
      if (typeof(value) == 'undefined') {
        return msgid;
      } else {
        return (typeof(value) == 'string') ? value : value[0];
      }
    }
   ```
   to this:
   ```js
    django.gettext = function(msgid) {
      var value = django.catalog[msgid];
      if (typeof(value) != 'undefined') {
        return (typeof(value) == 'string') ? value : value[0];
      } else {
        return window.gettext(msgid);
      }
    };
   ```
   ?
1. If the previous point is too complex, we could add wrapper functions to [these two lines](https://github.com/openedx/xblock-drag-and-drop-v2/blob/7d8f41b7499cd9c39b18b6f2907db2ce3e7dc229/drag_and_drop_v2/public/js/drag_and_drop_edit.js#L7-L8) at the beginning of `drag_and_drop.js` and `drag_and_drop_edit.js`. It would work similarly to the solution from above (i.e. fall back to `window.gettext` if the XBlock's translation method's output would be equal to its input).

Now, this obviously limits these overrides to strings that are not translated into the specified language. However, it works correctly for all English text (because `input === output`), so a simple and non-invasive solution would be sufficient here.

tecoholic:
@Agrendalath The `text.js` files are created with the [django-static18n] plugin's `compilejsi18n` management command with the namespace `DragAndDropI18N`. Changing the content of it wouldn't be a straightforward solution as the main JS block seems to come from [within Django's JSONCatalog](https://github.com/zyegfryed/django-statici18n/blob/main/src/statici18n/management/commands/compilejsi18n.py#L96-L103). 

I have used the suggestion of wrapping the 2 lines in [f4b68ce](https://github.com/openedx/xblock-drag-and-drop-v2/pull/290/commits/f4b68ceb23469abc202356ee5a86616d3b733a35) and it works as expected.

Agrendalath:
@natabene, a quick check before we move forward with merging this. The user-facing impact here is that more text will appear as translated because we are retrieving missing translations directly from the edx-platform.

natabene:
@tecoholic Thank you for your contribution. Let me run this by the team.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -5,8 +5,23 @@ function DragAndDropTemplates(configuration) {
...
+            // if DnDv2's translation is the same as the input, check if global has a different value
+            // This is useful for overriding the XBlock's string by themes (only for English)
+            if (string === translated && 'gettext' in window) {
+                translated = window.gettext(string);
+            }
+            return translated;
+        }
+        ngettext = function(strA, strB, n) {
+            let translated = window.DragAndDropI18N.ngettext(strA, strB, n);
+            let string = n == 1 ? strA : strB;
```
jmbowman, line 19: What about languages with pluralization rules different than English?  It seems like there's a chance that ngettext could return the other input in that case, and we'd miss the fact that the text wasn't translated because we checked for equality with the wrong input.

jmbowman (commented):
I have one question about a potential corner case bug, but otherwise this seems reasonable to me.

natabene:
@tecoholic Thank you for your contribution. Is this ready?

drag_and_drop_v2/public/js/drag_and_drop.js
tecoholic: @jmbowman Sorry for the delay here. I was off for a week and couldn't respond immediately.

For other languages the `translated` value would be different than the `string` value so `if (string === translated && 'gettext' in window) ` would fail and it would return the `translated` value, as we would be comparing some other language with one of the input English strings. So, I think it should work fine.

robrap: FYI: @jmbowman will be out for a couple of weeks, so there could be a delay.

jmbowman: Fair, the logic bug I was thinking of would only prevent use of the fallback when you want to use it, but if we only want it when the language is English then that's a moot point.

jmbowman (approved):
Sorry for the unreasonably long delay in replying; this looks good to go.

jmbowman:
Do you want to do any manual squashing of commits, or should I squash and merge with the original title and description?

tecoholic:
@jmbowman @Agrendalath seems to have recently rebased the PR branch, updated the tests and ticket details. So everything is updated. Kindly squash and merge.

openedx-webhooks:
@tecoholic 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/xblock-drag-and-drop-v2&entry.867055334=https://github.com/openedx/xblock-drag-and-drop-v2/pull/290&entry.1484655318=https://github.com/tecoholic&entry.752974735=2022-07-21+06:19&entry.1917517419=2022-09-23+08:25&entry.2133058324=Yes) so we can improve your experience in the future.

openedx-webhooks:
@tecoholic 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/xblock-drag-and-drop-v2&entry.867055334=https://github.com/openedx/xblock-drag-and-drop-v2/pull/290&entry.1484655318=https://github.com/tecoholic&entry.752974735=2022-07-21+06:19&entry.1917517419=2022-09-23+08:25&entry.2133058324=Yes) so we can improve your experience in the future.

Agrendalath:
@jmbowman, the existing commit messages were no longer accurate. As you already gave the +1, I allowed myself to squash&merge this with the title from the PR and the body from its description. I also created a new release.

Thank you for reviewing this!

PR #294, opened 2022-09-23
fix: Support course-wide "Show Answer" setting

abodacs:
## Description

This PR fixes the global setting for the Show Answer button

The global setting for the Show Answer button is not working for the d&dv2 block; we have it set as "Finished" in the advanced settings (so the button should be clickable after they have used all attempts on the problem or they have correctly answered the problem.), but in d&d v2 blocks, it's greyed out until all attempts have been used.

## Testing instructions

- Login to Studio shell, switch to edxapp virtual env
- Go to /edx/src/xblock-drag-and-drop-v2 and run pip install -e .
- Go to "Settings" -> "Advanced Settings" in studio and the search for "Show Answer" and set it to finished.
- Go to Studio UI and create a new unit
- Select Problem > Advanced > Drag and Drop, to create a v2 drag and drop XBblock
- Click Edit and change mode to Assessment
- Click Continue twice and Save
- Click on View Live Version
- Show Answer button should be clickable after used all attempts on the problem or they have correctly answered the problem

## Supporting information

- [x] Fix the global setting for the Show Answer button
- [x] Fix ```showanswer``` setting failing tests
- [x] Add tests with different "Show answer" statues


openedx-webhooks:
Thanks for the pull request, @abodacs!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.




:warning: We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](https://openedx.org/cla)     or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.     If you've signed an agreement in the past, you may need to re-sign.  See     [The New Home of the Open edX Codebase](https://open.edx.org/blog/the-new-home-of-the-open-edx-codebase/)     for details.

openedx-webhooks:
Thanks for the pull request, @abodacs!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.




:warning: We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](https://openedx.org/cla)     or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.     If you've signed an agreement in the past, you may need to re-sign.  See     [The New Home of the Open edX Codebase](https://open.edx.org/blog/the-new-home-of-the-open-edx-codebase/)     for details.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -133,6 +132,28 @@ class DragAndDropBlock(
         enforce_type=True,
     )
 
+    showanswer = String(
+        display_name=_("Show Answer"),
+        help=_("Defines when to show the answer to the problem. "
+               "A default value can be set in Advanced Settings."),
+        scope=Scope.settings,
+        default=SHOWANSWER.FINISHED,
+        values=[
```
abodacs, line 141: Should I keep "showanswer" values list (ie, always, finish, etc)?
If the answer is yes,  How do I  add translations for display_name?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -968,7 +1069,8 @@ def _get_user_state(self):
         if self.mode == Constants.STANDARD_MODE:
             is_finished = self._is_answer_correct()
         else:
-            is_finished = not self.attempts_remain
+            is_finished = self.is_finished
```
abodacs, line 1072: Is this change acceptable?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -133,6 +132,28 @@ class DragAndDropBlock(
...
+            {"display_name": _("Finished"), "value": SHOWANSWER.FINISHED},
+            {"display_name": _("Correct or Past Due"), "value": SHOWANSWER.CORRECT_OR_PAST_DUE},
+            {"display_name": _("Past Due"), "value": SHOWANSWER.PAST_DUE},
+            {"display_name": _("Never"), "value": SHOWANSWER.NEVER},
+            {"display_name": _("After All Attempts"), "value": SHOWANSWER.AFTER_ALL_ATTEMPTS},
+            {"display_name": _("After All Attempts or Correct"), "value": SHOWANSWER.AFTER_ALL_ATTEMPTS_OR_CORRECT},
+            {"display_name": _("Attempted"), "value": SHOWANSWER.ATTEMPTED_NO_PAST_DUE},
+        ],
+        enforce_type=True,
+    )
```
tecoholic, line 155: We don't need to introduce a new settings for this XBlock. The value set in the "Advanced Settings" of the course is already available in the XBlock's `showanswer` attribute and can be accessed using `self.xblock`. So this can be dropped.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +703,83 @@ def has_submission_deadline_passed(self):
...
+
+    @property
+    def is_correct(self):
+        """
+        Returns True if answer is correct.
+        """
+        return self._is_answer_correct()
+
+    @property
+    def is_answer_available(self):
```
tecoholic, line 750: Even though we are removing the 'showanswer' setting, this property will still work since `showanswer` available as mentioned in the comment above. So, don't worry about removing the setting above.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1919,15 +1933,38 @@ function DragAndDropBlock(runtime, element, configuration) {
...
     };
 
     var attemptsRemain = function() {
         return !configuration.max_attempts || configuration.max_attempts > state.attempts;
     };
 
+
+    var canShowAnswer = function() {
+        if(configuration.mode !== DragAndDropBlock.ASSESSMENT_MODE) return false;
+        else if (configuration.show_answer_status === SHOW_ANSWER_STATUSES.NEVER) return false;
```
tecoholic, line 1954: I don't see the `show_answer_status` being passed as a configuration value [in the student_view](https://github.com/openedx/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/drag_and_drop_v2.py#L336), so this will always return false as the `show_answer_status` will be `undefined`.

I think the student_view's configuration should have something like
```
"show_answer_status": self.showanswer
```

tests/unit/test_basics.py
```diff
@@ -306,7 +306,6 @@ def test_studio_submit(self):
         body = self._make_submission()
         res = self.call_handler('studio_submit', body)
         self.assertEqual(res, {'result': 'success'})
-
```
tecoholic, line 309: Nit: rollback empty line removal.

tests/unit/data/html/config_out.json
```diff
@@ -9,6 +9,7 @@
     "graded": false,
     "weighted_max_score": 1,
     "show_title": false,
+    "showanswer": "attempteds",
```
tecoholic, line 12: Nit:
```suggestion
    "showanswer": "attempted",
```

tecoholic (changes requested):
@abodacs I have reviewed the changes as they are now, tested them locally and left some comments. Kindly let me know once the PR is ready for review and is non-draft.

drag_and_drop_v2/drag_and_drop_v2.py
abodacs: If I drop the above setting for  ```showanswer```,  tests breaks and I got an error

```python
  File "/home/abdullah/workspace/opencraft/src/xblock-drag-and-drop-v2/drag_and_drop_v2/drag_and_drop_v2.py", line 413, in student_view_data
    "show_answer_status": self.showanswer,
AttributeError: 'DragAndDropBlockWithMixins' object has no attribute 'showanswer'
```
Is there another way to access ```showanswer`` XBlock setting?

Thanks

tecoholic:
@natabene Hi, @abodacs has singed the CLA 2 days back and it doesn't seem to reflect in the PR. Can you kindly see if there is anything that needs to be done?

drag_and_drop_v2/drag_and_drop_v2.py
tecoholic: @abodacs Interestingly the attribute seems to load the right values from the advanced settings during runtime. I tried to see if I can find when/where this attribute is being added to the XBlock, but couldn't. I see a couple ways to get around this:
1. Mock `showanswer` in the tests
2. Define `showanswer` as a property in the XBlock and make it return the value by fetching the course settings from the runtime. But I think we might be unnecessarily re-implementing an existing functionality for the sake of tests.
3. Define a wrapper around the auto-injected `showanswer` like below and using that
```
@property
def show_answer(self):
    return getattr(self, 'showanswer', 'answered')
```

But, I am not sure which is the right way to go.

@Agrendalath Can you help here?

Agrendalath: @tecoholic, @abodacs,

This attribute comes from the [InheritanceMixin](https://github.com/openedx/edx-platform/blob/fcb594d6352b93e53aa89fcc758e18ae709752fc/xmodule/modulestore/inheritance.py#L86-L98). This mixin, [along with some other mixins](https://github.com/openedx/edx-platform/blob/fcb594d6352b93e53aa89fcc758e18ae709752fc/cms/envs/common.py#L932-L941), is usually injected [while constructing an XBlock](https://github.com/openedx/edx-platform/blob/fcb594d6352b93e53aa89fcc758e18ae709752fc/xmodule/modulestore/split_mongo/caching_descriptor_system.py#L232-L242).

That said, I believe it's a good idea to introduce the `showanswer` field directly in the XBlock - even though we define default behavior for the course, we may still want to override it for particular XBlocks.

Agrendalath:
@abodacs, to fix the CI, you may want to restore the `xblock-sdk` you've deleted. It exists for a reason I've described in `requirements.txt`.

abodacs:
@Agrendalath 
Thanks, I restored the ```xblock-sdk```

Agrendalath:
@abodacs, in 7c750c31d8c63ea4ef8df9363c181c0b2cd2ca25, you have added a different version of `xblock-sdk`…

drag_and_drop_v2/drag_and_drop_v2.py
tecoholic: @Agrendalath I see. Thank you for the references. I have needed them multiple times before as well, and have simply settled to using a debugger to inspect the final xblock. I am going to bookmark these for future uses.

@abodacs I am sorry for misleading you on this. You have been on the right path :+1:

abodacs:
@Agrendalath 
 to add submodule ```xblock-sdk``` and  Fix the submodule to a particular tag

1.  Should the submodule tag be ```v0.1.9```?
2-  Do I miss anything if I run those commands?

```bash
git submodule add git://github.com/asynchrony/xblock-sdk.git xblock-sdk
cd xblock-sdk
git checkout v0.1.9
````

Thanks

File: tests/unit/test_advanced.py [hunk 8]
```diff
@@ -943,3 +939,72 @@ def test_do_attempt_correct_takes_decoy_into_account(self):
...
+        res = self.call_handler(self.SHOW_ANSWER_HANDLER, data={}, expect_json=False)
+        self.assertEqual(res.status_code, 400)
+
+        self._submit_complete_solution()
+        res = self.call_handler(self.SHOW_ANSWER_HANDLER, data={}, expect_json=False)
+        self.assertEqual(res.status_code, 200)
+
+        self.assertTrue(self.block.attempts_remain)
+
+        self.assertFalse(self.block.closed)
```
tecoholic, line 1010: @abodacs The quality tests are failing here because the line count is exceeding 1000. Can you kindly see if you can remove some of the empty lines at appropriate placed to get it to stay under 1000?

drag_and_drop_v2/public/js/drag_and_drop.js
tecoholic: @abodacs I was testing the changes now, and I don't see the above changes (specifically updates to `drag_and_drop.js`) in the PR anymore. Did some commits go missing while rebasing?

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1919,15 +1933,37 @@ function DragAndDropBlock(runtime, element, configuration) {
...
     };
 
     var attemptsRemain = function() {
         return !configuration.max_attempts || configuration.max_attempts > state.attempts;
     };
 
+    var canShowAnswer = function() {
+        if(configuration.mode !== DragAndDropBlock.ASSESSMENT_MODE) return false;
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.NEVER) return false;
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.ATTEMPTED) return isAttempted() || isPastDue();
```
tecoholic, line 1954: @abodacs Setting the "Show Answer" value to "attempted" is causing the the button to be enabled even before I have made the first attempt. I think the issue is with the use of `isPastDue` function here. Since `isPastDue` is actually returning the *inverse* of `configuration.has_deadline_passed`, the logic of the Python equivalent `self.is_attempted or self.has_submission_deadline_passed` is being altered, resulting in wrong results.

abodacs: @tecoholic 
Thanks for detailed answered

I removed the inverse from ```isPastDue``` and edited ```canSubmitAttempt``` function based on the inverse

``` javascript
var canSubmitAttempt = function() {
        return Object.keys(state.items).length > 0 && !isPastDue() && attemptsRemain() && !submittingLocation();
    };

```

tecoholic: @abodacs Great. I tested the various settings again. They seem to work as expected :)

tecoholic (approved):
@abodacs :+1:

- [x] I tested this: I set the "Show Answer" settings in course's advanced settings to different values and tested that XBlock behaves as expected.
- [x] I read through the code
- [ ] Includes documentation

@Agrendalath Can you kindly approve CI workflow and review the PR if everything passes?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +700,82 @@ def has_submission_deadline_passed(self):
...
+        if self.max_attempts is None or self.max_attempts == 0:
+            return False
+        return self.attempts >= self.max_attempts
+
+    @property
+    def is_correct(self):
+        """
+        Returns True if answer is correct.
+        """
+        return self._is_answer_correct()
```
Agrendalath, line 743: We already have `self._is_answer_correct()`. Why don't we use it directly in `is_answer_available` instead of introducing a new layer of methods? If we want to to call it `is_correct`, then we should rename `_is_answer_correct` instead of introducing a chain of one-line methods.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +700,82 @@ def has_submission_deadline_passed(self):
...
 
+    @property
+    def closed(self):
+        """
+        Is the student still allowed to submit answers?
+        """
+        if self.used_all_attempts:
+            return True
+        if self.has_submission_deadline_passed:
+            return True
```
Agrendalath, line 711: This can be used directly in `is_answer_available` instead of adding a new method. Same for `is_attempted` and `is_finished`. Unless we are using them in any other places in this XBlock, we don't gain much by introducing new properties.

drag_and_drop_v2/drag_and_drop_v2.py
Agrendalath: What if the results should be hidden (e.g. during an exam)? Wouldn't this leak the answer correctness to users?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -612,10 +633,10 @@ def show_answer(self, data, suffix=''):
                 400,
                 self.i18n_service.gettext("show_answer handler should only be called for assessment mode")
             )
-        if self.attempts_remain:
+        if not self.is_answer_available:
             raise JsonHandlerError(
-                409,
-                self.i18n_service.gettext("There are attempts remaining")
+                400,
```
Agrendalath, line 638: What was wrong with HTTP 409?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -612,10 +633,10 @@ def show_answer(self, data, suffix=''):
...
+                self.i18n_service.gettext("answer is not available")
```
Agrendalath, line 639: Please use the title case, e.g. `The answer is unavailable`.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +700,82 @@ def has_submission_deadline_passed(self):
...
+            _answer_available = self.has_submission_deadline_passed
+        elif self.showanswer == SHOWANSWER.ALWAYS:
+            _answer_available = True
+        elif self.showanswer == SHOWANSWER.AFTER_ALL_ATTEMPTS:
+            _answer_available = self.used_all_attempts
+        elif self.showanswer == SHOWANSWER.AFTER_ALL_ATTEMPTS_OR_CORRECT:
+            _answer_available = self.used_all_attempts or self.is_correct
+        elif self.showanswer == SHOWANSWER.ATTEMPTED_NO_PAST_DUE:
+            _answer_available = self.is_attempted
+        return _answer_available
```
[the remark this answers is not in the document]
abodacs: @Agrendalath 
Thanks for your review.

I  did the copy and past but I got a [linter error #R0911](https://vald-phoenix.github.io/pylint-errors/plerr/errors/design/R0911.html) so I edited the code block to fix the linter error.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +700,82 @@ def has_submission_deadline_passed(self):
...
+        if self.mode != Constants.ASSESSMENT_MODE:
+            return False
+        _answer_available = False
+        if self.showanswer == SHOWANSWER.NEVER:
+            _answer_available = False
+        elif self.showanswer == SHOWANSWER.ATTEMPTED:
+            _answer_available = self.is_attempted or self.has_submission_deadline_passed
+        elif self.showanswer == SHOWANSWER.ANSWERED:
+            # NOTE: this is slightly different from 'attempted' -- resetting the problems
+            # makes lcp.done False, but leaves attempts unchanged.
```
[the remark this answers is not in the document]
abodacs: Thanks
Done

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +700,82 @@ def has_submission_deadline_passed(self):
...
+    def is_answer_available(self):
+        """
+        Is student allowed to see an answer?
+        """
+        if self.mode != Constants.ASSESSMENT_MODE:
+            return False
+        _answer_available = False
+        if self.showanswer == SHOWANSWER.NEVER:
+            _answer_available = False
+        elif self.showanswer == SHOWANSWER.ATTEMPTED:
```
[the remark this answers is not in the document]
abodacs: because the service does not exist I got ```NoSuchServiceError```

```python
raise NoSuchServiceError(f"Service {service_name!r} was not requested.")
xblock.exceptions.NoSuchServiceError: Service 'user' was not requested.
```
can you mention if there is another way to call ```user``` service?

Thanks

drag_and_drop_v2/drag_and_drop_v2.py
abodacs: @Agrendalath 
I tried to follow the code pattern mentioned [here](https://github.com/openedx/edx-platform/blob/master/xmodule/capa_module.py#L1383-L1430) which I believe can increase the readability of ```is_answer_available```

Thanks

drag_and_drop_v2/drag_and_drop_v2.py
abodacs: @Agrendalath

It will not leak the answer correctness to users as this code only applies in the assessment mode.


```is_finished ``` return true if a) the answer is correct or b) the student has used all allowed attempts or c)  the submission is past its deadline.

if this still leaks the answer correctness to users please let me know so I can change

Thanks @Agrendalath

File: drag_and_drop_v2/drag_and_drop_v2.py [hunk 17]
```diff
@@ -679,6 +700,82 @@ def has_submission_deadline_passed(self):
...
+        return self.closed or self.is_correct
+
+    @property
+    def used_all_attempts(self):
+        """
+        Returns True if student has used all allowed attempts.
+        """
+        if self.max_attempts is None or self.max_attempts == 0:
+            return False
+        return self.attempts >= self.max_attempts
```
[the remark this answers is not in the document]
abodacs: ```used_all_attempts``` is the inverse of ```attempts_remain```.

Thanks, I'll use ```not attempts_remain``` instead

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1917,17 +1931,40 @@ function DragAndDropBlock(runtime, element, configuration) {
...
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.FINISHED) return isClosed() || isCorrect();
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.CORRECT_OR_PAST_DUE) return isCorrect() || isPastDue();
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.PAST_DUE) return isPastDue();
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.ALWAYS) return true;
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.AFTER_ALL_ATTEMPTS) return !attemptsRemain();
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.AFTER_ALL_ATTEMPTS_OR_CORRECT) return !attemptsRemain() || isCorrect();
+        else if (configuration.showanswer === SHOW_ANSWER_STATUSES.ATTEMPTED_NO_PAST_DUE) return isAttempted();
+  
+        return false;
+    };
```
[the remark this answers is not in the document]
abodacs: - ```"showanswer": self.is_answer_available``` will be used only at question load when calling ```student_view_data```, right?

My understanding is that the Show answer button will not work after ```do_attempt```

- Also ```showanswer``` is a [course-setting](https://github.com/abodacs/xblock-drag-and-drop-v2/blob/ca61d3902bb3af60b2970bdf8f0c4a6c36ba773a/drag_and_drop_v2/drag_and_drop_v2.py#L135)  and passing it as a boolean I find confusing.

But I agree with you unifying the logic should be the goal

drag_and_drop_v2/drag_and_drop_v2.py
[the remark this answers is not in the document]
abodacs: I refactored the code to solve  the linter error and ```staff``` permissions check

drag_and_drop_v2/public/js/drag_and_drop.js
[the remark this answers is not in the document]
tecoholic: @Agrendalath I requested this addition as the client state for the "Show Answer" button doesn't get updated after user interaction. For example, when set in the "answered" mode, submitting the answer doesn't enable it without reload.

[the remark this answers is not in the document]
Agrendalath: @abodacs, @tecoholic, `do_attempt` is returning a dict with updated XBlock state. What would prevent us from returning `"showanswer": self.is_answer_available` [here](https://github.com/openedx/xblock-drag-and-drop-v2/blob/3fddeb57ce28c98d2ad40db6ecb074a69b9e2a0f/drag_and_drop_v2/drag_and_drop_v2.py#L571-L578) and updating it [here](https://github.com/openedx/xblock-drag-and-drop-v2/blob/3fddeb57ce28c98d2ad40db6ecb074a69b9e2a0f/drag_and_drop_v2/public/js/drag_and_drop.js#L1868-L1897)?

We definitely do not want to implement this logic in two places (Python and JS) because it's hard to maintain it and keep it in sync.

drag_and_drop_v2/drag_and_drop_v2.py
Agrendalath: @abodacs, the assessment mode is what is used during exams. It could be hard to use the standard mode there.

What would happen in the following scenario?
1. Users are taking an exam. Users should not be able to see if the answer is correct until the exam is finished (is past the due date).
2. Users have multiple attempts available, so even after they submit an answer, they can still change it as long as the exam is active.
3. A user submits their answer and it's incorrect. Therefore, `is_finished` is `False`, and they can still change their answer.
4. A user submits their answer and it's correct. Therefore, `is_finished` is `True`, and they cannot change their answer anymore. They don't know the answer, but they can see it's correct because the XBlock has the `finished` state.
5. They could potentially even use the browser's Network tab to check the `finished` param in the response.

I didn't test the described scenario, but this is what I'm understanding from a quick look at the implementation. Please correct me if I'm wrong.

drag_and_drop_v2/public/js/drag_and_drop.js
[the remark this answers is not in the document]
tecoholic: @Agrendalath Ah. That would be a more appropriate solution. :100: I agree about not duplicating the logic in 2 places. 
cc: @abodacs

drag_and_drop_v2/drag_and_drop_v2.py
abodacs: @Agrendalath 

Aha, I think I got confused with the variable ```is_finished``` and the new added ```self.is_finished``` I thought there should be equal

My current understanding

- ```is_finished``` means should the user allowed to submit an answer
- ```self.is_finished``` means should the user allowed to show the answer button
- submits an answer is equal to do attempt

What would happen in the following scenario?
 _my answer when ```is_finished = self.is_finished```_

1. Users are taking an exam. Users should not be able to see if the answer is correct until the exam is finished (is past the due date).

 `is_finished` is `True` as `self.has_submission_deadline_passed` is also `True` which `self.is_finished` depends on.

2. Users have multiple attempts available, so even after they submit an answer, they can still change it as long as the exam is active.
My understanding for point 2 is that  `is_finished` is `False` as long as there are still attempts remaining and the exam is active
but this confuses me because _**they can still change it as long as the exam is active.**_  doesn't match with 3 and 4 or points 3 and 4 should be 2.a and 2.b
 
3. A user submits their answer and it's incorrect. Therefore, `is_finished` is `False`, and they can still change their answer.

`is_finished` is `False`as exam is still active and attempts remaining are available `not self.attempts_remain`

4. A user submits their answer and it's correct. Therefore, `is_finished` is `True`, and they cannot change their answer anymore. They don't know the answer, but they can see it's correct because the XBlock has the finished state.

`is_finished` is `True` as the exam is still active and the answer is correct `self.is_correct`

abodacs: solutions to that

- revert `is_finished` back to equal to  `not self.attempts_remain` and open an issue if it is wrong.
```python
   is_finished = not self.attempts_remain
```
or

- edit `is_finished` code like 
```python
   is_finished = not self.attempts_remain or self.is_correct or self.has_submission_deadline_passed
```
which is the same as `self.is_finished` but `self.is_finished ` is used for another purpose (ie, Show answer button) 

but this solution conflicts with point 2 above (in my understanding)

Of course, any other suggestions are welcomed

drag_and_drop_v2/public/js/drag_and_drop.js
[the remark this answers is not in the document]
abodacs: @Agrendalath

I added `"answer_available": self.is_answer_available` becuase `showanswer` is a [course-setting](https://github.com/abodacs/xblock-drag-and-drop-v2/blob/ca61d3902bb3af60b2970bdf8f0c4a6c36ba773a/drag_and_drop_v2/drag_and_drop_v2.py#L135) and passing it as a boolean causes tests to fail.

Thanks

cc: @tecoholic

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1991,6 +1991,7 @@ function DragAndDropBlock(runtime, element, configuration) {
         // gain focus while keyboard placement is in progress.
         var item_bank_focusable = (state.keyboard_placement_mode || state.showing_answer) &&
             configuration.mode === DragAndDropBlock.ASSESSMENT_MODE;
+
```
Agrendalath, line 1994: Nit: please remove unnecessary blank line changes from the PR - they seem to be redundant.

drag_and_drop_v2/public/js/drag_and_drop.js
```diff
@@ -1917,16 +1917,16 @@ function DragAndDropBlock(runtime, element, configuration) {
...
 
-    var canShowAnswer = function() {
-        return configuration.mode === DragAndDropBlock.ASSESSMENT_MODE && !attemptsRemain();
-    };
 
     var attemptsRemain = function() {
         return !configuration.max_attempts || configuration.max_attempts > state.attempts;
     };
+    var canShowAnswer = function() {
+        return state.answer_available || configuration.answer_available;
```
Agrendalath, line 1928: Nit: we should use `configuration.answer_available` only when `state.answer_available` is undefined. Otherwise, we should rely solely on `state.answer_available`.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +703,63 @@ def has_submission_deadline_passed(self):
...
+            SHOWANSWER.ATTEMPTED: lambda _self: _self.is_attempted or _self.has_submission_deadline_passed,
+            SHOWANSWER.ANSWERED: lambda _self: _self.is_correct,
+            SHOWANSWER.CLOSED: lambda _self: _self.closed,
+            SHOWANSWER.FINISHED: lambda _self: _self.is_finished,
+            SHOWANSWER.CORRECT_OR_PAST_DUE: lambda _self: _self.is_correct or _self.has_submission_deadline_passed,
+            SHOWANSWER.PAST_DUE: lambda _self: _self.has_submission_deadline_passed,
+            SHOWANSWER.ALWAYS: lambda _self: True,
+            SHOWANSWER.AFTER_ALL_ATTEMPTS: lambda _self: not _self.attempts_remain,
+            SHOWANSWER.AFTER_ALL_ATTEMPTS_OR_CORRECT: lambda _self: not _self.attempts_remain or _self.is_correct,
+            SHOWANSWER.ATTEMPTED_NO_PAST_DUE: lambda _self: _self.is_attempted,
```
Agrendalath, line 750: Why are we using this weird syntax? `SHOWANSWER.ATTEMPTED_NO_PAST_DUE: lambda: self.is_attempted,` should work too, no?

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +703,63 @@ def has_submission_deadline_passed(self):
...
+        """
+        return self.closed or self.is_correct
+
+    @property
+    def is_answer_available(self):
+        """
+        Is student allowed to see an answer?
+        """
+        def _return_false(_x):
+            return False
```
Agrendalath, line 738: Please just use `lambda: False` instead of defining something like this…

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +703,63 @@ def has_submission_deadline_passed(self):
...
+        }
+
+        if self.mode != Constants.ASSESSMENT_MODE:
+            return False
+        _current_user = self.runtime.service(self, 'user').get_current_user()
+        user_is_staff = _current_user.opt_attrs.get(Constants.ATTR_KEY_USER_IS_STAFF)
+        if self.showanswer not in [SHOWANSWER.NEVER, ''] and user_is_staff:
+            # admins can see the answer unless the problem explicitly prevents it.
+            return True
+        check_available_fun = _look_up.get(self.showanswer, _return_false)
```
Agrendalath, line 760:
```suggestion
        check_available_function = _look_up.get(self.showanswer, lambda: True)
```

There is still some `"fun" in "function"`. We don't need this abbreviation.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +703,63 @@ def has_submission_deadline_passed(self):
...
+            SHOWANSWER.ATTEMPTED_NO_PAST_DUE: lambda _self: _self.is_attempted,
+        }
+
+        if self.mode != Constants.ASSESSMENT_MODE:
+            return False
+        _current_user = self.runtime.service(self, 'user').get_current_user()
+        user_is_staff = _current_user.opt_attrs.get(Constants.ATTR_KEY_USER_IS_STAFF)
+        if self.showanswer not in [SHOWANSWER.NEVER, ''] and user_is_staff:
+            # admins can see the answer unless the problem explicitly prevents it.
+            return True
```
Agrendalath, line 759: Nit: please add an extra line below there `return` statements. This code is pretty crowdy without them.

Agrendalath (changes requested):
> Show Answer button should be clickable after used all attempts on the problem or they have correctly answered the problem

This behavior is not matching the one from the Problem XBlock. In the Problem XBlock, the button [is not rendered](https://github.com/open-craft/edx-platform/blob/a5f687bf70a6d395a56ab360ce1d802735615eb6/lms/templates/problem.html#L48-L52) when the answer is unavailable.

drag_and_drop_v2/drag_and_drop_v2.py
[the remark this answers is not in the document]
Agrendalath: @abodacs, please add this explanation, though - it's useful to know the difference.

Agrendalath:
@abodacs, also, please document this feature in the `README`. It's nice to have a feature, but to be useful to the community, users should be able to read about it somewhere. Check [this page](https://edx.readthedocs.io/projects/open-edx-building-and-running-a-course/en/latest/course_components/create_problem.html#show-answer) to see how it's documented for the Problem XBlock.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +703,63 @@ def has_submission_deadline_passed(self):
...
+            SHOWANSWER.PAST_DUE: lambda _self: _self.has_submission_deadline_passed,
+            SHOWANSWER.ALWAYS: lambda _self: True,
+            SHOWANSWER.AFTER_ALL_ATTEMPTS: lambda _self: not _self.attempts_remain,
+            SHOWANSWER.AFTER_ALL_ATTEMPTS_OR_CORRECT: lambda _self: not _self.attempts_remain or _self.is_correct,
+            SHOWANSWER.ATTEMPTED_NO_PAST_DUE: lambda _self: _self.is_attempted,
+        }
+
+        if self.mode != Constants.ASSESSMENT_MODE:
+            return False
+        _current_user = self.runtime.service(self, 'user').get_current_user()
```
Agrendalath, line 755: Nit: we're not prefixing variables with an underscore anywhere in this repository. Please be consistent with the existing convention.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -132,6 +133,27 @@ class DragAndDropBlock(
         default=True,
         enforce_type=True,
     )
+    showanswer = String(
```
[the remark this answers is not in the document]
abodacs: `showanswer` is a global course setting, you can go to "Settings" -> "Advanced Settings" in the studio and search for "Show Answer".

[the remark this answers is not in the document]
Agrendalath: @abodacs, we are defining it in the XBlock, so it's an XBlock-specific option which **can be** overridden in the "Advanced Settings", as I've mentioned [in this comment](https://github.com/openedx/xblock-drag-and-drop-v2/pull/294#discussion_r983343560). You can find an explanation of how the Scope.settings works [in this document](https://edx.readthedocs.io/projects/xblock/en/latest/fields.html#xblock.fields.Scope).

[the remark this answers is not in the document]
Agrendalath: @abodacs, I'm answering your Jira comment here, to preserve the context - yes, your understanding is correct. When you add the `showanswer` field to an XBlock, it determines the value in the following order:

1. If the value is set in the DnDv2 editor - use it. This is an XBlock-specific setting.
2. If the XBlock-specific setting is not set - use the one from the course "Advanced Settings".

This mechanism is already implemented, so we don't need to worry about it. The only thing we want is to allow defining it in the Studio editor. Therefore, we want to add a dropdown with possible values (like [this one](https://github.com/openedx/xblock-drag-and-drop-v2/blob/cfd2b2dcc7e77e3fbb8949f8f1d8a548b29aa078/drag_and_drop_v2/templates/html/drag_and_drop_edit.html#L25-L34)).
There is a small catch - this XBlock doesn't have a native "reset" value option, but we need to allow course authors to reset this value in case they want to revert to using a course-specific option. I suggest adding a ["Clear" button](https://github.com/open-craft/edx-platform/blob/740d6ed59c7bb4821c23db936b0b4402f22d22d7/cms/templates/js/metadata-option-entry.underscore#L12-L14) which sets an empty value for this setting. Then, we can handle it [here](https://github.com/openedx/xblock-drag-and-drop-v2/blob/cfd2b2dcc7e77e3fbb8949f8f1d8a548b29aa078/drag_and_drop_v2/drag_and_drop_v2.py#L474) - when the `self.showanswer` is empty, we run `del self.showanswer`, which simply removes the XBlock-specific setting.
If you're curious about how it works, you can find it [here](https://github.com/openedx/XBlock/blob/a89dcae5b97e1822ceedbf5f4e299bfad5f3aa1a/xblock/fields.py#L550-L576).

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -679,6 +702,62 @@ def has_submission_deadline_passed(self):
...
+            SHOWANSWER.ALWAYS: lambda: True,
+            SHOWANSWER.AFTER_ALL_ATTEMPTS: lambda: not self.attempts_remain,
+            SHOWANSWER.AFTER_ALL_ATTEMPTS_OR_CORRECT: lambda: not self.attempts_remain or self.is_correct,
+            SHOWANSWER.ATTEMPTED_NO_PAST_DUE: lambda: self.is_attempted,
+        }
+
+        if self.mode != Constants.ASSESSMENT_MODE:
+            return False
+        current_user = self.runtime.service(self, 'user').get_current_user()
+        user_is_staff = current_user.opt_attrs.get(Constants.ATTR_KEY_USER_IS_STAFF)
```
Agrendalath, line 753: We should check if the service exists. There are some rare edge cases where it can be absent.
For example, it can happen when a user tries to access a Course Outline page of a course that doesn't have its block structures generated yet.

openedx-webhooks:
@abodacs 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/xblock-drag-and-drop-v2&entry.867055334=https://github.com/openedx/xblock-drag-and-drop-v2/pull/294&entry.1484655318=https://github.com/abodacs&entry.752974735=2022-09-23+15:34&entry.1917517419=2022-10-18+11:28&entry.2133058324=Yes) so we can improve your experience in the future.

Agrendalath (commented):
:+1:

* [x] I tested this: tested the functionality on Maple, Nutmeg, and Master
* [x] I read through the code
* [x] Includes documentation
* [x] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository: n/a

bradenmacdonald:
Thanks for this contribution @abodacs, and thanks for reviewing it @Agrendalath !

PR #308, opened 2023-01-31
feat!: Update to the latest version of bleach.

feanil:
Update the requirements and matching code. We also cleanup a bunch of
code around the CSS sanitizer since edx-platform runs with bleach 5.0.0
now.

BREAKING CHANGE: Updating to the latest version of bleach will mean we
can't run with bleach < 6.0.0

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,11 +1,18 @@
 Drag and Drop XBlock changelog
 ==============================
 
+Version 3.1.0 (2023-01-31)
+--------------------------
+
+* Upgrade to be compatible with `bleach==6.0.0` and `bleach<6.0.0`
+* Make the dependency on the `css` extras explicit.
+
 Version 3.0.0 (2022-11-18)
 ---------------------------
 
-* Sanitize HTML tags to prevent XSS vulnerabilities.  
-  BREAKING CHANGE: Disallowed HTML tags (e.g. `<script>`) will no longer be rendered in LMS and Studio. 
+* Sanitize HTML tags to prevent XSS vulnerabilities.
+
+  BREAKING CHANGE: Disallowed HTML tags (e.g. `<script>`) will no longer be rendered in LMS and Studio.
 
 Version 2.7.0 (2022-11-15)
 ---------------------------
@@ -20,7 +27,7 @@ Version 2.6.0 (2022-10-24)
 Version 2.5.0 (2022-10-13)
 ---------------------------
 
-* Make the "Show Answer" condition customizable (like in the Problem XBlock). 
+* Make the "Show Answer" condition customizable (like in the Problem XBlock).
 
 Version 2.4.2 (2022-10-13)
 ---------------------------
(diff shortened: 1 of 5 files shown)


Changelog.md
```diff
@@ -1,11 +1,20 @@
...
+
+  BREAKING CHANGE: We can't run with bleach < 6.0.0
+
 Version 3.0.0 (2022-11-18)
 ---------------------------
 
-* Sanitize HTML tags to prevent XSS vulnerabilities.  
-  BREAKING CHANGE: Disallowed HTML tags (e.g. `<script>`) will no longer be rendered in LMS and Studio. 
+* Sanitize HTML tags to prevent XSS vulnerabilities.
+  BREAKING CHANGE: Disallowed HTML tags (e.g. `<script>`) will no longer be rendered in LMS and Studio.
```
Agrendalath, line 17: Nit: I suppose it was an automatic change from your editor, but double space means a line break in Markdown, so this changes the rendered result. If you'd like to get rid of it, please either change `  ` to ` \` (the result will be the same) or add a new line above `BREAKING CHANGE` (this will increase the space between these lines, but it's okay).

GitHub's rich diff does not handle it well, but you can check out how it is currently rendered [here](https://github.com/openedx/xblock-drag-and-drop-v2/blob/784e6ad5e4b09f2a0fd87d80d187bfe0ac36da7c/Changelog.md#version-300-2022-11-18).

drag_and_drop_v2/utils.py
```diff
@@ -37,7 +31,7 @@ def _clean_data(data):
     return cleaned_text
 
 
-ALLOWED_TAGS = bleach.ALLOWED_TAGS + [
+ALLOWED_TAGS = bleach.ALLOWED_TAGS | {
```
Agrendalath, line 34: Idea: this bleach `5.0.1` -> `6.0.0` upgrade does not need to break anything. I suggested an approach in which we could preserve compatibility with Nutmeg and Olive with only five characters of code overhead. We could also unpin the bleach constraint. If we went ahead with this approach, we could deploy it to the `edx-platform` right away as the `3.1.0` version. What do you think?

The Maple support removal looks excellent, of course. Thanks for doing it!

```suggestion
# Convert `bleach.ALLOWED_TAGS` to a set because it is a list in `bleach<6.0.0`.
ALLOWED_TAGS = set(bleach.ALLOWED_TAGS) | {
```

Agrendalath (changes requested):

Changelog.md
```diff
@@ -1,11 +1,18 @@
 Drag and Drop XBlock changelog
 ==============================
 
+Version 4.0.0 (2023-01-31)
```
Agrendalath, line 4: Nit: please update the version here too. Also, we could use `3.1.0` instead of `3.0.1`. After all, we are removing Maple compatibility.

Agrendalath (approved):
:+1:

* [x] I tested this: checked that the XBlock is still working correctly in Nutmeg
* [x] I read through the code
* [x] Includes documentation: n/a
* [x] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository: n/a

PR #540, opened 2026-07-14
feat: include item metadata in drag-and-drop report state

hiteshkava:
## Description

Enhance the Drag-and-Drop XBlock report data by including additional item
metadata within `item_state`.

Currently, reports only expose zone mappings for each item, which limits the
ability to understand learner interactions and outcomes when analyzing report
data.

This change extends `item_state` to include relevant item metadata such as:

- `displayName`
- `feedback`
- `correct`
- `imageURL`

Example:

```json{
  "item_state": {
    "0": {
      "zone": "top",
      "displayName": "Goes to the top",
      "feedback": "Correct placement",
      "correct": true,
      "imageURL": "/static/example.png"
    }
  }
}'''

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,10 +2,12 @@ Drag and Drop XBlock changelog
 ==============================
 
 Unreleased
-Version 5.0.7 (2026-07-23)
 --------------------------
 
-* Added `displayName`, `feedback`, `correct` and `imageURL` to the learner item state returned by `student_view_user_state` and stored in `item_state`.
+Version 5.1.0 (2026-07-23)
+--------------------------
+
+* Added `display_name`, `feedback` and `image_url` to the learner item state returned by `student_view_user_state` and stored in `item_state`.
 
 Version 5.0.6 (2026-04-29)
 --------------------------
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,10 @@ Drag and Drop XBlock changelog
 ==============================
 
 Unreleased
----------------------------
+Version 5.0.7 (2026-07-23)
+--------------------------
+
+* Added `displayName`, `feedback`, `correct` and `imageURL` to the learner item state returned by `student_view_user_state` and stored in `item_state`.
 
 Version 5.0.6 (2026-04-29)
 --------------------------
--- a/drag_and_drop_v2/__init__.py
+++ b/drag_and_drop_v2/__init__.py
@@ -1,4 +1,4 @@
 """ Drag and Drop v2 XBlock """
 from .drag_and_drop_v2 import DragAndDropBlock
 
-__version__ = "5.0.7"
+__version__ = "5.1.0"
--- a/drag_and_drop_v2/__init__.py
+++ b/drag_and_drop_v2/__init__.py
@@ -1,4 +1,4 @@
 """ Drag and Drop v2 XBlock """
 from .drag_and_drop_v2 import DragAndDropBlock
 
-__version__ = "5.0.6"
+__version__ = "5.0.7"
--- a/drag_and_drop_v2/drag_and_drop_v2.py
+++ b/drag_and_drop_v2/drag_and_drop_v2.py
@@ -977,10 +977,11 @@ class DragAndDropBlock(
                 raise JsonHandlerError(400, "Item zone data invalid.")
 
     @staticmethod
-    def _make_state_from_attempt(attempt, correct, item={}):
+    def _make_state_from_attempt(attempt, correct, item=None):
         """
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by Agrendalath** [comment]
@hiteshkava, could you please provide more context about this change? Why would we want to duplicate data that's already available in the XBlock? What report are you referring to? How to test this?

**Comment by hiteshkava** [comment]
> @hiteshkava, could you please provide more context about this change? Why would we want to duplicate data that's already available in the XBlock? What report are you referring to? How to test this?

Thanks for the questions, @Agrendalath.

The change is intended to enhance the learner state data returned by the Drag-and-Drop XBlock reports. Currently, the  report only contains the zone mapping for each item, for example:

```
{
  "item_state": {
    "0": {
      "zone": "top",
      "correct": : true
    }
  }
}
```


While the item configuration exists within the XBlock definition, consumers of the report only receive the learner state and do not have direct access to the corresponding item metadata. As a result, it is difficult to interpret report data without separately querying and correlating the XBlock configuration.

The goal of this change is to enrich the report payload with useful metadata such as:

- displayName
- feedback
- correct
- imageURL

This makes the exported report self-contained and easier to consume for reporting, analytics, and downstream integrations.

The report I'm referring to is the learner **Problem Response** report data generated from the Drag-and-Drop XBlock, where item_state is currently returned with only zone information.

How to test
Create a Drag-and-Drop XBlock with items that include display names, feedback, and images.
Complete the exercise as a learner.
Retrieve the learner state/report data before and after the change.
Verify that the item_state payload now includes the additional metadata alongside the existing zone and correct values.

Example:

```
{
  "item_state": {
    "0": {
      "zone": "top",
      "displayName": "Goes to the top",
      "correct": true,
      "feedback": "Correct placement"
    }
  }
}
```

**Comment by jristau1984** [comment]
@Agrendalath just wanted to check on when you think this PR could get reviewed and merged?  I want to set expectations internally at 2U, and a timeline would be most helpful.  Thanks!

**Comment by Agrendalath** [comment]
HI @jristau1984, I'll check this one tomorrow. 

@hiteshkava, in the meantime, would you mind fixing the tests, adding an entry to the changelog, and bumping the minor version?

**Comment by hiteshkava** [comment]
> @hiteshkava, in the meantime, would you mind fixing the tests, adding an entry to the changelog, and bumping the minor version?

@Agrendalath  Done! I've fixed the tests, added the changelog entry, bumped the minor version, and updated the changes on the repository.

File: Changelog.md [hunk 1]
```diff
@@ -2,7 +2,10 @@ Drag and Drop XBlock changelog
 ==============================
 
 Unreleased
----------------------------
```
Agrendalath, line 5: This line + newline should be preserved. See the rendered document.

drag_and_drop_v2/drag_and_drop_v2.py
```diff
@@ -977,13 +977,17 @@ def _validate_drop_item(self, item):
...
         """
         Converts "attempt" data coming from browser into "state" entry stored in item_state
         """
+        item = item or {}
         return {
             'zone': attempt['zone'],
-            'correct': correct
+            'correct': correct,
+            'imageURL': item.get('imageURL'),
+            'displayName': item.get('displayName'),
```
Agrendalath: Let's use snake_case to keep the naming consistent. Examples: `item_state`, `raw_earned`, `raw_possible`.

Changelog.md
```diff
@@ -2,7 +2,10 @@ Drag and Drop XBlock changelog
...
+Version 5.0.7 (2026-07-23)
+--------------------------
+
+* Added `displayName`, `feedback`, `correct` and `imageURL` to the learner item state returned by `student_view_user_state` and stored in `item_state`.
```
Agrendalath: Nit: `correct` was already there.

drag_and_drop_v2/__init__.py
```diff
@@ -1,4 +1,4 @@
 """ Drag and Drop v2 XBlock """
 from .drag_and_drop_v2 import DragAndDropBlock
 
-__version__ = "5.0.6"
+__version__ = "5.0.7"
```
Agrendalath: This is a patch version bump, not the minor one.

Agrendalath (commented):
@hiteshkava, everything works well. Just a few small requests.

hiteshkava:
> @hiteshkava, everything works well. Just a few small requests.

Done ! @Agrendalath

Agrendalath (approved):
:+1:

* [x] I tested this: checked that the new fields are correctly included in the "report of problem responses"
* [x] I read through the code
* [x] Includes documentation: n/a

